PR #15, opened 2020-11-19
DCNG-891 move Ingress definitions into a new functest Helm chart

kennymacleod:
This slightly speculative, but it bothers me that the Ingress definitions that we load for testing use an entirely different templating mechanism (`envsubst`) than the Helm cherts do, and it means installating a new tool to use it.

Instead, I've added a new `functest` Helm chart to `src/test`, which contains the EKS and KITT ingress definitions. There's only one functest chart shared between all products, with product-specific conditionals in the template.

This has the advantage of removing the per-product Ingress resources, and using `helm install` for all deployed components, so it's all slightly simpler.

It'll also work better for @kcichy-atlassian if he needs to add per-cluster-node backdoor ingress routes.


src/test/charts/functest/templates/ingress-EKS.yaml
```diff
@@ -1,7 +1,12 @@
+{{- if eq "EKS" .Values.clusterType }}
```
kennymacleod, line 1: even though the functest Helm chart has both EKS and KITT ingress definitions, only one wil be rendered, based on this conditional

pbruski (approved):

kcichy-atlassian:
> It'll also work better for @kcichy-atlassian if he needs to add per-cluster-node backdoor ingress routes.

Haha, a bit of disservice, because he already put together the `envsubst` code. Will refactor on Monday 😃.

PR #16, opened 2020-11-19
DCNG-881: Create services and ingress routes to access individual nodes

kcichy-atlassian:
In the light of @kennymacleod 's #15 PR, this might need to change. Let's discuss as it is now and I'll refactor later if necessary.

### DIFF
--- a/src/test/charts/functest/templates/ingress-EKS.yaml
+++ b/src/test/charts/functest/templates/ingress-EKS.yaml
@@ -16,17 +16,17 @@ metadata:
     "nginx.ingress.kubernetes.io/proxy-send-timeout": "60"
 spec:
   rules:
-    {{- range $ingressName := .Values.ingressNames }}
-    - host: {{ printf "%s.%s" . $.Values.ingressDomain }}
+    {{- range $serviceName := .Values.ingressNames }}
+    - host: {{ printf "%s.%s" $serviceName $.Values.ingressDomain }}
       http:
         paths:
           - backend:
-              serviceName: {{ $ingressName }}
+              serviceName: {{ $serviceName }}
               servicePort: 80
             path: /
           {{- if eq "confluence" $.Values.product }}
           - backend:
-              serviceName: {{ $ingressName }}-synchrony
+              serviceName: {{ $serviceName }}-synchrony
               servicePort: 80
             path: /synchrony
           {{ end }}
--- a/src/test/charts/functest/templates/ingress-EKS.yaml
+++ b/src/test/charts/functest/templates/ingress-EKS.yaml
@@ -16,17 +16,19 @@ metadata:
     "nginx.ingress.kubernetes.io/proxy-send-timeout": "60"
 spec:
   rules:
-    - host: {{ printf "%s.%s" .Values.productReleaseName .Values.ingressDomain }}
+    {{- range $ingressName := .Values.ingressNames }}
+    - host: {{ printf "%s.%s" . $.Values.ingressDomain }}
       http:
         paths:
           - backend:
-              serviceName: {{ .Values.productReleaseName }}
+              serviceName: {{ $ingressName }}
               servicePort: 80
             path: /
-          {{- if eq "confluence" .Values.product }}
+          {{- if eq "confluence" $.Values.product }}
           - backend:
-              serviceName: {{ .Values.productReleaseName }}-synchrony
+              serviceName: {{ $ingressName }}-synchrony
               servicePort: 80
             path: /synchrony
           {{ end }}
+    {{ end }}
 {{ end }}
--- a/src/test/config/jira/helm_parameters
+++ b/src/test/config/jira/helm_parameters
@@ -12,3 +12,5 @@ DOCKER_IMAGE_REGISTRY=${dockerImage.registry}
 DOCKER_IMAGE_VERSION=${dockerImage.version}
 INGRESS_DOMAIN_KITT=${kitt.ingress.domain}
 INGRESS_DOMAIN_EKS=${eks.ingress.domain}
+# .Values.replicaCount=1, but in here we specify how many nodes we are planning to get to prepare ingress rules for each
+TARGET_REPLICA_COUNT=2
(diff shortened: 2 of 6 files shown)


src/test/config/jira/service-expose-node.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: ${SERVICE_NAME}
+spec:
+  type: ClusterIP
+  ports:
+    - port: ${PORT_PRODUCT_SERVICE}
+      protocol: TCP
+      targetPort: ${PORT_PRODUCT_CONTAINER}
+  selector:
+    statefulset.kubernetes.io/pod-name: ${SERVICE_NAME}
```
kennymacleod, line 12: this is a little misleading.... the name of this backdoor service is the same as the name of the pod it's backdoor-ing, so maybe this var should be `POD_NAME`

src/test/config/jira/helm_parameters
```diff
@@ -2,6 +2,7 @@ RELEASE_PREFIX=${helm.release.prefix}
 CHART_SRC_PATH=${helm.src.path}/jira
 PRODUCT_CHART_VALUES_FILES=${project.build.directory}/config/jira/values
 PRODUCT_INGRESS_TEMPLATE=${project.build.directory}/config/jira/ingress
+PRODUCT_SERVICE_EXPOSE_NODE_TEMPLATE=${project.build.directory}/config/jira/service-expose-node
```
kennymacleod, line 5: We'll need these changes made to all 3 products' `helm_parameters` files

src/test/config/jira/helm_parameters
```diff
@@ -11,4 +12,8 @@ TARGET_NAMESPACE=${kubernetes.target.namespace}
 DOCKER_IMAGE_REGISTRY=${dockerImage.registry}
 DOCKER_IMAGE_VERSION=${dockerImage.version}
 INGRESS_DOMAIN_KITT=${kitt.ingress.domain}
-INGRESS_DOMAIN_EKS=${eks.ingress.domain}
\ No newline at end of file
+INGRESS_DOMAIN_EKS=${eks.ingress.domain}
+# .Values.replicaCount=1, but in here we specify how many nodes we are planning to get to prepare ingress rules for each
+TARGET_REPLICA_COUNT=2
```
kennymacleod, line 17: Can we just hard-code this for now? We're unlikely to go beyond 2 nodes in testing.

src/test/config/jira/helm_parameters
```diff
@@ -11,4 +12,8 @@ TARGET_NAMESPACE=${kubernetes.target.namespace}
...
+PORT_PRODUCT_SERVICE=80
```
kennymacleod, line 18: again, let's just hard-code this, it's not going to vary

src/test/config/jira/service-expose-node.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: ${SERVICE_NAME}
+spec:
+  type: ClusterIP
+  ports:
+    - port: ${PORT_PRODUCT_SERVICE}
+      protocol: TCP
+      targetPort: ${PORT_PRODUCT_CONTAINER}
```
kennymacleod, line 10: The `targetPort` can instead be the name of a container port, so this can just be `http` (since that's the name we use for that container port in all 3 products, regardless of the numerical value)

src/test/scripts/helm_install.sh
```diff
@@ -81,10 +81,41 @@ fi
...
+      envsubst \
+      < "${PRODUCT_INGRESS_TEMPLATE}-${clusterType}.yaml" \
+      > ${LOG_DOWNLOAD_DIR}/ingress-${HELM_RELEASE_NAME}.yaml
+  kubectl apply -n "${TARGET_NAMESPACE}" --filename ${LOG_DOWNLOAD_DIR}/ingress-${HELM_RELEASE_NAME}.yaml
+}
+
+# params:
+# NODE - numeric id of the node for which service should be installed
+# SERVICE_NAME - the requested service name
+function install_service_expose() {
```
kennymacleod, line 98: let's rename this, or add a comment, to be clear that we're installing a backdoor service for each pod, purely for func testing purposes

src/test/config/jira/helm_parameters
kcichy-atlassian: Do we? Other products won't use it unless they specify `TARGET_REPLICA_COUNT`. If `TARGET_REPLICA_COUNT` is undef, this won't be used.

kcichy-atlassian:
@kennymacleod 
I added the helm-way of generating backdoor services and ingresses.

src/test/config/jira/service-expose-node.yaml
kcichy-atlassian: Renamed to `$podName`.

src/test/config/jira/helm_parameters
kcichy-atlassian: Hmm, I'm allergic to hardcoding. Why do you think it's a problem in this case? It would allow us to change a for loop with something flat, but I'm not sure if that's much of a gain.

src/test/charts/functest/templates/ingress-EKS.yaml
```diff
@@ -16,17 +16,19 @@ metadata:
     "nginx.ingress.kubernetes.io/proxy-send-timeout": "60"
 spec:
   rules:
-    - host: {{ printf "%s.%s" .Values.productReleaseName .Values.ingressDomain }}
+    {{- range $ingressName := .Values.ingressNames }}
+    - host: {{ printf "%s.%s" . $.Values.ingressDomain }}
```
kennymacleod, line 20: should be using `$ingressName` instead of `.` here, for consstency

kcichy-atlassian: Absolutely. I spotted this in `ingress-KITT.yaml` last night, but I missed this one.

kennymacleod (approved):

PR #17, opened 2020-11-23
Add ServiceAccount and ClusterRole to Confluence chart

kennymacleod:
Currently, our Helm charts requires the customer to pre-configure a `ServiceAccount`, `ClusterRole` and `ClusterRoleBinding`. In the case of Confluence and Bitbucket, this requires very specific role permissions in order for peer discovery to work, and this isn't really something we want customers to have to do.

So I've added those 3 resources to the Confluence Helm chart, with corresponding entries in `values.yaml` to control how they get expressed. By default, all 3 will be created automatically, with generated names, but the customer can pick their own names, or disable their creation entirely.

I've also changed the `values.yaml` that we use for EKS so that we *do* use the generated `ServiceAccount`, `ClusterRole` and `ClusterRoleBinding`, so that we exercise this part of the chart in our tests.

For KITT, we need to keep using the current approach of using the pre-defined service account, since we have no permissions to create a new one.

See #18 doing the same thing for Jira

### DIFF
--- a/src/main/charts/confluence/templates/_helpers.tpl
+++ b/src/main/charts/confluence/templates/_helpers.tpl
@@ -48,6 +48,50 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
+The name of the service account to be used.
+If the name is defined in the chart values, then use that,
+else if we're creating a new service account then use the name of the Helm release,
+else just use the "default" service account.
+*/}}
+{{- define "confluence.serviceAccountName" -}}
+{{- if .Values.serviceAccount.name }}
+{{- .Values.serviceAccount.name }}
+{{- else }}
+{{- if .Values.serviceAccount.create }}
+{{- include "confluence.fullname" . -}}
+{{ else }}
+default
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+The name of the ClusterRole that will be created.
+If the name is defined in the chart values, then use that,
+else use the name of the Helm release.
+*/}}
+{{- define "confluence.clusterRoleName" -}}
+{{- if .Values.serviceAccount.clusterRole.name }}
+{{- .Values.serviceAccount.clusterRole.name }}
+{{- else }}
+{{- include "confluence.fullname" . -}}
+{{- end }}
+{{- end }}
+
+{{/*
+The name of the ClusterRoleBinding that will be created.
+If the name is defined in the chart values, then use that,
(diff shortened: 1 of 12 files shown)


src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -47,6 +47,50 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
 {{- end }}
```
kennymacleod, line 49: Here I've added 3 new template fragments for rendering the names of th new resources, with sensible defaults.

src/main/charts/confluence/templates/clusterrole.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - apiGroups:
+      - ""
+    resources:
+      - endpoints
+      - pods
+      - nodes
+    verbs:
+      - get
+      - list
+{{- end -}}
```
kennymacleod, line 18: This files defines the permissions that the hazelcast-kubernetes adapter needs to perform peer discovery

src/main/charts/confluence/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,16 @@
...
+  {{- include "confluence.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "confluence.clusterRoleName" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "confluence.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+{{ end }}
```
kennymacleod, line 16: This file associates the service account with the cluster role.

src/test/java/test/HelmOutputComparisonTest.java
```diff
@@ -47,6 +47,7 @@ private void captureHelmTemplateOutput(String product, Path outputFile) throws E
                         getHelmReleaseName(product),
                         getHelmChartPath(product).toString(),
                         "--debug",
+                        "-n mynamespace",
```
kennymacleod, line 50: I was getting different test results depending on which environment I was running in, so I added this to make it predictable

pbruski (approved):

PR #20, opened 2020-11-23
DCNG-897 Add hook-delete-policy to chart tests and nfs-fixer job

kennymacleod:
Adds the following annotation to each Chart hook we have in our charts (i.e. the chart tests, and the nfs-permission-fixers):

    "helm.sh/hook-delete-policy": hook-succeeded

This will ensure that the hook pods are cleaned up when they're finished, rather than cuttering up the namespace as they currently do. Note that Helm hook pods are *not* cleaned up by `helm uninstall`, unlike every other resource in the Chart.

No, I don't know why either.

### DIFF
--- a/src/test/scripts/helm_install.sh
+++ b/src/test/scripts/helm_install.sh
@@ -81,7 +81,7 @@ then
 fi
 
 # Run the chart's tests
-helm test "$PRODUCT_RELEASE_NAME" --logs -n "${TARGET_NAMESPACE}"
+helm test "$PRODUCT_RELEASE_NAME" -n "${TARGET_NAMESPACE}"
 
 # Package and install the functest helm chart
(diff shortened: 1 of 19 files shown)


src/test/scripts/helm_install.sh
```diff
@@ -81,7 +81,7 @@ then
 fi
 
 # Run the chart's tests
-helm test "$PRODUCT_RELEASE_NAME" --logs -n "${TARGET_NAMESPACE}"
```
kennymacleod, line 84: The `--logs` will now no longer work, since the pod will be deleted as soon as the test passes.

Note that failed tests should still have their logs captured by the "log download" task.

pbruski (changes requested):

pbruski (approved):

PR #22, opened 2020-11-24
DCNG-893 change Jira chart to disable use of PVs by default

kennymacleod:
As requested by Max Bern, I've changed the Jira chart so that the use of `PersistentVolumes` is disabled by default. This makes it a lot easier to get up and running with the chart, since you don't have to provision any storage.

If the flag is false (i.e. the default) then the `local-home` and `shared-home` volumes in the statefulset will of type `emptyDir`, which means the volume contents will be lst when the pod is shut down. 

If the flag is true (e.g. via a customer `values.yaml`) file then the full Persistent Volumes will be used as before.

Note that the CI loop will enable persistence, so that will behave as before.


src/test/resources/expected_helm_output/jira/output.yaml
```diff
@@ -168,20 +168,13 @@ spec:
               mountPath: "/opt/atlassian/jira/atlassian-jira/WEB-INF/atlassian-bundled-plugins/myplugin.jar"
               subPath: "plugins/myplugin.jar"
       volumes:
+        - name: local-home
+          emptyDir: {}
         - name: shared-home
-          persistentVolumeClaim:
-            claimName: jira-shared-home
```
kcichy-atlassian, line 173: Could we have another "unit test" that would assert what happens when `persistence.enabled=true`?

kcichy-atlassian (approved):

src/test/resources/expected_helm_output/jira/output.yaml
kennymacleod: Yes, we really should. I'm chewing on an idea for a better way of unit testing these helm templates, but for now we're relying on the deployment integration tests to test that.

bkwiatek-atlassian (approved):

src/test/resources/expected_helm_output/jira/output.yaml
kennymacleod: @kcichy-atlassian please see #23 for the "better way". I haven't added tests for this specifically yet, but I will.

pbruski (changes requested):

PR #23, opened 2020-11-24
DCNG-898 better unit tests for chart rendering

kennymacleod:
Our current unit tests for the Helm chart rendering are too crude. They simply compare the YAML output of the rendered chart against an expected YAML file, failing on any differences.

What I've added in this PR is a finer-grained set of tests that only assert on the specific part of the output file that we're interested in. This allows us to write many more unit tests, more focussed on one thing at a time. 

This is becoming more important as we start adding chart features that customrs will find useful, but which we don't (or can't) use ourselves in our CI loop.

I've already found one bug in the ServiceAccount rendering.


src/main/charts/bitbucket/templates/_helpers.tpl
```diff
@@ -41,15 +41,15 @@ else if we're creating a new service account then use the name of the Helm relea
 else just use the "default" service account.
 */}}
 {{- define "bitbucket.serviceAccountName" -}}
```
kennymacleod, line 43: The new unit tests found a bug in this template. Fixed it with the indiscriminate use of whitespace chomping indicators.

src/test/java/test/HelmOutputComparisonTest.java
```diff
@@ -10,78 +14,43 @@
 
 class HelmOutputComparisonTest {
```
kennymacleod, line 15: This is the original unit test that just compared the entire Helm YAML output with an expectation file. It's useful as a baseline, so I'm keeping it for now. I've just refactored it to use the extracted `Helm` utlity class.

src/test/java/test/ServiceAccountTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        assertThat(resources.getAll(ServiceAccount))
+                .describedAs("No ServiceAccount resources should be created")
+                .isEmpty();
+
+        resources.getStatefulSets()
+                .forEach(statefulSet ->
+                        assertThat(statefulSet.getPodSpec().path("serviceAccountName"))
+                                .describedAs("StatefulSet %s should have the default ServiceAccount name", statefulSet.getName())
+                                .hasTextEqualTo("default"));
```
kennymacleod, line 56: this test caught a bug in the serviceAccountName template. Hooray.

src/test/java/test/helm/Helm.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * A wrapper around the helm CLI tool, which we use for rendering a chart's templates and capturing the output.
+ */
+public final class Helm {
```
kennymacleod, line 20: The contents of this class were mostly extracted from the original unit test, but now it's a bit more general and is used by all the new tests.

src/test/java/test/jackson/JsonNodeAssert.java
```diff
@@ -0,0 +1,36 @@
...
+import org.assertj.core.api.AbstractAssert;
+
+import java.util.Objects;
+
+import static com.fasterxml.jackson.databind.node.JsonNodeType.STRING;
+
+/**
+ * A simple custom AssertJ assertion class for handling basic {@link JsonNode} assertions.
+ */
+public class JsonNodeAssert extends AbstractAssert<JsonNodeAssert, JsonNode> {
```
kennymacleod, line 13: Not strictly necessary, but gives us cleaner assertions in the tests. Quite surprised such a thing doesn't already exist.

src/test/java/test/model/Product.java
```diff
@@ -0,0 +1,24 @@
+package test.model;
+
+public enum Product {
+    jira {
```
pbruski, line 4: hmm, lowercase naming convention?

pbruski (approved):

src/test/java/test/helm/Helm.java
```diff
@@ -0,0 +1,86 @@
...
+                .redirectOutput(outputFile.toFile())
+                .redirectError(outputFile.toFile())
+                .start();
+        final var exitCode = process.waitFor();
+        assertThat(exitCode).isEqualTo(0);
+    }
+
+    public KubeResources captureKubeResourcesFromHelmChart(Product product, Map<String, String> values) throws Exception {
+        final var outputFile = getHelmTemplateOutputFile(product);
+        captureHelmTemplateOutput(product, outputFile, values);
```
kcichy-atlassian, line 51: I just started reviewing this, so I might be missing something.
This line is
```
captureHelmTemplateOutput(product, outputFile, values);
```
while line 29 is
```
captureHelmTemplateOutput(product, valuesFile, outputFile);
```
Is that on purpose, or are the 2nd and third parameters swapped?

PR #34, opened 2020-12-07
DCNG-927 Make Tomcat https/secure config consistent between products

kennymacleod:
The Bitbucket Helm chart allows the Tomcat "scheme" and "secure" config properties to be configured via chart vaues, but the other products' charts do not.

This PR makes them consistent by using the same value structure for all 3. In the process I've renamed the `bitbucket.proxy` structure to `bitbucket.ingress` (and also for confluence+jira), since the ingress is the kubernetes terminoogy for what the Bitbucket docs call a "proxy".

### DIFF
--- a/src/main/charts/bitbucket/values.yaml
+++ b/src/main/charts/bitbucket/values.yaml
@@ -86,7 +86,7 @@ bitbucket:
     fqdn:
     # -- The port number of the ingress
     port: 443
-    # -- The protocol scheme used by the ingress that sends traffic to the application.
+    # -- The protocol scheme used by the browser to access the application.
     # This is necessary so that the application generates the correct URLs.
     # Note that, if present, the value of x-forwarded-proto header will override this setting.
     scheme: https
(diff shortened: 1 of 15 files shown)


pbruski (approved):

src/main/charts/bitbucket/values.yaml
```diff
@@ -81,13 +81,16 @@ bitbucket:
...
     emailAddressSecretKey: emailAddress
 
-  proxy:
+  ingress:
     # -- The fully-qualified domain name of the ingress
     fqdn:
     # -- The port number of the ingress
     port: 443
-    # -- note that, if present, the value of x-forwarded-proto header will trump this setting
+    # -- The protocol scheme used by the ingress that sends traffic to the application.
```
pbruski, line 89: Shouldn't that be "the protocol scheme used by the browser to access the application"?

pbruski: To clarify: our ingresses in EKS use HTTP, not HTTPS. HTTPS happens outside of Kubernetes.

pbruski (changes requested):

PR #35, opened 2020-12-08
DCNG-906 Add support for declaring ElasticSearch properties

kennymacleod:
Adds a new optional structure to the Bitbucket chart values, allowing for the declaration of the ElasticSearch base URL and credentials.

It's pretty basic at this point, but means the customer won't have to manually specify the additional environment varables themselves.

It mandates that the credentials are stored in a Kubernetes secret, consistent with the other optional credentials in the charts.

### DIFF
--- a/src/main/charts/bitbucket/README.md
+++ b/src/main/charts/bitbucket/README.md
@@ -28,7 +28,7 @@ Kubernetes: `>=1.17.x-0`
 | bitbucket.additionalLibraries | list | `[]` | Specifies a list of additional Java libraries that should be added to the Bitbucket container. Each item in the list should specify the name of the volume which contain the library, as well as the name of the library file within that volume's root directory. Optionally, a subDirectory field can be included to specify which directory in the volume contains the library file. |
 | bitbucket.additionalVolumeMounts | list | `[]` | Defines any additional volumes mounts for the Bitbucket container. These can refer to existing volumes, or new volumes can be defined in volumes.additional. |
 | bitbucket.clustering.enabled | bool | `false` | Set to true if Data Center clustering should be enabled This will automatically configure cluster peer discovery between cluster nodes. |
-| bitbucket.elasticSearch.baseUrl | string | `nil` | The base URL of the ElasticSearch instance to be used. |
+| bitbucket.elasticSearch.baseUrl | string | `nil` | The base URL of the external ElasticSearch instance to be used. If this is defined, then Bitbucket will disable its internal ElasticSearch instance. |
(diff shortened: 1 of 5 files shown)


pbruski (approved):

bianchi2:
@kennymacleod maybe I'm missing smth but shouldn't we also let a user set ELASTICSEARCH_ENABLED to false if an external ES instance is used? As far as I understand the main use case is to let a user define an external ES (which means we don't need to start one in Bitbucket container).

Also, what if no ES URL is set, we won't set the env var to container spec or set an empty one? You have ifs for secrets but not for the URL itself. I guess it won't be set then - just want to confirm :)

kennymacleod:
> @kennymacleod maybe I'm missing smth but shouldn't we also let a user set ELASTICSEARCH_ENABLED to false if an external ES instance is used? As far as I understand the main use case is to let a user define an external ES (which means we don't need to start one in Bitbucket container).

I don't see the ELASTICSEARCH_ENABLED env var documented anywhere, where are you seeing that?

My own understanding is that if you dont define a baseUrl for an external ES, then BBS will bootstrap its own internal one, is that not correct?

> Also, what if no ES URL is set, we won't set the env var to container spec or set an empty one? You have ifs for secrets but not for the URL itself. I guess it won't be set then - just want to confirm :)

Is the URL value is not provided, then the BASEURL env var will not be set. There's no `{{ if }}` there, but there is a `{{ with }}`, which has the same effect.

bianchi2:
@kennymacleod I followed https://hub.docker.com/r/atlassian/bitbucket-server and ELASTICSEARCH_ENABLED is mentioned there. There's also an example of docker run with an external ES where this env var is used along with url and creds. I can't confirm though that ES won't start in a container if only the URL is defined. Maybe DockerHub description is outdated.

bianchi2:
It's in the image entrypoint:

```
docker run -ti --entrypoint=bash atlassian/bitbucket-server
root@72beb2f19f07:/# cat entrypoint.py 
#!/usr/bin/python3 -B
from entrypoint_helpers import env, str2bool, start_app
RUN_USER = env['run_user']
RUN_GROUP = env['run_group']
BITBUCKET_INSTALL_DIR = env['bitbucket_install_dir']
BITBUCKET_HOME = env['bitbucket_home']

start_cmd = f"{BITBUCKET_INSTALL_DIR}/bin/start-bitbucket.sh -fg"
if str2bool(env['elasticsearch_enabled']) is False or env['application_mode'] == 'mirror':
    start_cmd += ' --no-search'

start_app(start_cmd, BITBUCKET_HOME, name='Bitbucket Server')
```

bianchi2 (approved):
LGTM

PR #36, opened 2020-12-08
DCNG-892 Add Ingress templates

kennymacleod:
After much head-scratching about how Ingresses are defined across different Kubernetes API versions, I've added an Ingress template to each of our Helm charts.

By default, the Ingress is not created, the customer will have to set the `create` value to `true`. The customer also has to specify the `host` value, plus the annotations corresponding to the specific Ingress controller that the customer hapens to be using, e.g. ingress-nginx.

Also, because we use ingress-nginx in our EKS cluster, we can use (and test) these new Ingress templates in our EKS bamboo jobs. The KITT jobs still need to create the custom Contour Ingresses in the functest chart.

### DIFF
--- a/src/main/charts/bitbucket/templates/_helpers.tpl
+++ b/src/main/charts/bitbucket/templates/_helpers.tpl
@@ -102,10 +102,10 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 {{- define "bitbucket.baseUrl" -}}
-{{ $httpPortMismatch := (and (eq .Values.bitbucket.ingress.scheme "http") (ne (int .Values.bitbucket.ingress.port) 80) ) -}}
-{{ $httpsPortMismatch := (and (eq .Values.bitbucket.ingress.scheme "https") (ne (int .Values.bitbucket.ingress.port) 443) ) -}}
-{{ .Values.bitbucket.ingress.scheme }}://{{ .Values.bitbucket.ingress.host -}}
-{{ if or $httpPortMismatch $httpsPortMismatch }}:{{ .Values.bitbucket.ingress.port }}{{ end }}
+{{ $httpPortMismatch := (and (eq .Values.ingress.scheme "http") (ne (int .Values.ingress.port) 80) ) -}}
+{{ $httpsPortMismatch := (and (eq .Values.ingress.scheme "https") (ne (int .Values.ingress.port) 443) ) -}}
+{{ .Values.ingress.scheme }}://{{ .Values.ingress.host -}}
+{{ if or $httpPortMismatch $httpsPortMismatch }}:{{ .Values.ingress.port }}{{ end }}
 {{- end }}
 
 {{/*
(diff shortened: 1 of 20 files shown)


src/main/charts/bitbucket/templates/_helpers.tpl
```diff
@@ -104,7 +104,7 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- define "bitbucket.baseUrl" -}}
 {{ $httpPortMismatch := (and (eq .Values.bitbucket.ingress.scheme "http") (ne (int .Values.bitbucket.ingress.port) 80) ) -}}
 {{ $httpsPortMismatch := (and (eq .Values.bitbucket.ingress.scheme "https") (ne (int .Values.bitbucket.ingress.port) 443) ) -}}
-{{ .Values.bitbucket.ingress.scheme }}://{{ .Values.bitbucket.ingress.fqdn -}}
+{{ .Values.bitbucket.ingress.scheme }}://{{ .Values.bitbucket.ingress.host -}}
```
kennymacleod, line 107: renamed `fqdn` to `host`. It's less cryptic, and matches the terminology used in the Ingress rules.

src/main/charts/bitbucket/templates/ingress.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.bitbucket.ingress.create }}
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
```
kennymacleod, line 3: The content of this template describes a bog-standard Kubernetes Ingress. On its own, this isn't enough to work, the customer will need to provide the annotations that give the Ingress its flavour, e.g. for the nginx ingress controller.

src/main/charts/confluence/templates/ingress.yaml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  rules:
+    - host: {{ .Values.confluence.ingress.host }}
+      http:
+        paths:
+          - path: "/"
+            backend:
+              serviceName: {{ include "confluence.fullname" . }}
+              servicePort: {{ $.Values.confluence.service.port }}
+          {{ if .Values.synchrony.enabled }}
```
kennymacleod, line 21: The Confluence chart's ingress has an extra rule rulef ro Synchrony (if it's enabled)

src/test/charts/functest/templates/ingress-EKS.yaml
```diff
@@ -16,7 +16,7 @@ metadata:
     "nginx.ingress.kubernetes.io/proxy-send-timeout": "60"
 spec:
   rules:
-    {{- range $serviceName := .Values.ingressNames }}
+    {{- range $serviceName := .Values.backdoorServiceNames }}
```
kennymacleod, line 19: Now that we're using the Ingress from the main chart when on EKS, we only need to create an ingress for the backdoor services in the functest chart

src/test/config/bitbucket/values-EKS.yaml
```diff
@@ -5,7 +5,15 @@ serviceAccount:
...
-    fqdn: ${helm.release.prefix}-bitbucket.${eks.ingress.domain}
+    create: true
+    host: ${helm.release.prefix}-bitbucket.${eks.ingress.domain}
+    annotations:
+      "kubernetes.io/ingress.class": "nginx"
+      "nginx.ingress.kubernetes.io/affinity": "cookie"
+      "nginx.ingress.kubernetes.io/affinity-mode": "persistent"
+      "nginx.ingress.kubernetes.io/proxy-connect-timeout": "60"
+      "nginx.ingress.kubernetes.io/proxy-read-timeout": "60"
+      "nginx.ingress.kubernetes.io/proxy-send-timeout": "60"
```
kennymacleod, line 16: These annotations are going to be common if the customer uses the de-facto standard kube-nginx ingress controller, so in a later PR I'll move these into the main charts.

pbruski (approved):

bianchi2:
@kennymacleod there's one thing that we may want to (optionally) add here:

```
  tls:
  - hosts:
    - $hostName
      secretName: $secretName
```
It's not always that an ingress controller is set up with one secret, and often users provide secrets for different ingresses.

PR #37, opened 2020-12-09
DCNG-892 simplify config of https/http

kennymacleod:
Currently, the charts have config values for the `SERVER_SCHEME` and `SERVER_SECURE` env vars. The values for these must be consistent, and are difficult to describe at the level of abstraction that the Helm charts operate at. There's no reason this can't be represented in the charts as a single boolean value, `https`, from which the 2 env vars can be derived.

### DIFF
--- a/src/main/charts/bitbucket/templates/_helpers.tpl
+++ b/src/main/charts/bitbucket/templates/_helpers.tpl
@@ -102,10 +102,14 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 {{- define "bitbucket.baseUrl" -}}
-{{ $httpPortMismatch := (and (eq .Values.ingress.scheme "http") (ne (int .Values.ingress.port) 80) ) -}}
-{{ $httpsPortMismatch := (and (eq .Values.ingress.scheme "https") (ne (int .Values.ingress.port) 443) ) -}}
-{{ .Values.ingress.scheme }}://{{ .Values.ingress.host -}}
-{{ if or $httpPortMismatch $httpsPortMismatch }}:{{ .Values.ingress.port }}{{ end }}
+{{ ternary "https" "http" .Values.ingress.https -}}
+://
+{{- .Values.ingress.host -}}
+{{ with .Values.ingress.port }}:{{ . }}{{ end }}
+{{- end }}
+
+{{- define "bitbucket.ingressPort" -}}
+{{ default (ternary "443" "80" .Values.ingress.https) .Values.ingress.port -}}
 {{- end }}
 
 {{/*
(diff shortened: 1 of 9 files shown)


src/main/charts/bitbucket/templates/_helpers.tpl
```diff
@@ -102,10 +102,18 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
 {{- define "bitbucket.baseUrl" -}}
-{{ $httpPortMismatch := (and (eq .Values.ingress.scheme "http") (ne (int .Values.ingress.port) 80) ) -}}
-{{ $httpsPortMismatch := (and (eq .Values.ingress.scheme "https") (ne (int .Values.ingress.port) 443) ) -}}
-{{ .Values.ingress.scheme }}://{{ .Values.ingress.host -}}
-{{ if or $httpPortMismatch $httpsPortMismatch }}:{{ .Values.ingress.port }}{{ end }}
+{{ ternary "https" "http" .Values.ingress.https -}}
+://
+{{- .Values.ingress.host -}}
+{{ with .Values.ingress.port }}:{{ . }}{{ end }}
+{{- end }}
```
kennymacleod, line 109: I've rewritten the above function to be a little less terse. It'll include the `port` in the baseUrl if (and only if) it's specified in values.

src/main/charts/bitbucket/templates/statefulset.yaml
```diff
@@ -76,15 +76,17 @@ spec:
             {{ if .Values.ingress.host }}
             - name: SERVER_PROXY_NAME
               value: {{ .Values.ingress.host | quote }}
+            - name: SERVER_PROXY_PORT
+              value: {{ include "bitbucket.ingressPort" . | quote }}
             - name: SETUP_BASEURL
               value: {{ include "bitbucket.baseUrl" . | quote }}
             {{ end }}
-            - name: SERVER_PROXY_PORT
-              value: {{ .Values.ingress.port | quote }}
```
kennymacleod, line 83: move `SERVER_PROXY_PORT` into the same `if` block as `SERVER_PROXY_NAME`, they need to go together

src/main/charts/bitbucket/templates/statefulset.yaml
```diff
@@ -76,15 +76,17 @@ spec:
...
+            {{ if .Values.ingress.https }}
             - name: SERVER_SCHEME
-              value: {{ .Values.ingress.scheme | quote }}
+              value: "https"
             - name: SERVER_SECURE
-              value: {{ .Values.ingress.secure | quote }}
+              value: "true"
+            {{ end }}
```
kennymacleod, line 89: Now we set the `SERVER_SCHEME` and `SERVER_SECURE` env vars on the basis of a singe chart value, rather than one chart value for each.

bianchi2 (approved):
:+1:

PR #42, opened 2020-12-16
DCNG-942: Add a comment about updating Bash

kcichy-atlassian:
Very low priority. This is something I encountered when using the script the first time. As at that point everything was new to me, I had to scratch my head a few times before I understood the problem. Maybe we can save someone some minutes of their life.
I know this guide only talks about OS X, but I suppose the update is easier for Linux users, so no guide is necessary.

### DIFF
--- a/src/test/scripts/helm_install.sh
+++ b/src/test/scripts/helm_install.sh
@@ -18,6 +18,8 @@ currentContext=$(kubectl config current-context)
 
 echo Current context: $currentContext
 
+# If this fails, you might be using an old version of Bash.
+# Update following https://itnext.io/upgrading-bash-on-macos-7138bd1066ba.
 clusterType=$(case "${currentContext}" in
   *eks*) echo EKS;;
   *aks*) echo AKS;;


kennymacleod (approved):

src/test/scripts/helm_install.sh
```diff
@@ -18,6 +18,8 @@ currentContext=$(kubectl config current-context)
 
 echo Current context: $currentContext
 
+# If this fails, you might be using an old version of Bash.
+# Update following https://itnext.io/upgrading-bash-on-macos-7138bd1066ba.
```
kennymacleod, line 22: Maybe we could make the script fail if the bash version is too old?

pbruski:
> I know this guide only talks about OS X, but I suppose the update is easier for Linux users, so no guide is necessary.

Hahaha, I doubt that any of our customers is on a version of Linux with Bash as ancient as the one on Macos. It should say "Kept for historical purposes only" in the man page.

src/test/scripts/helm_install.sh
kcichy-atlassian: Good idea, I made Bash 5 a requirement for this script.

kcichy-atlassian:
Wait, you already merged that? 😲

PR #44, opened 2020-12-21
DCNG-925 support for BYO NFS server

pbruski:


bkwiatek-atlassian (approved):

src/main/charts/bitbucket/templates/shared-home-pv.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.volumes.sharedHome.persistentVolume.create }}
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: {{ include "bitbucket.fullname" . }}-shared-home-pv
+  labels:
+  {{- include "bitbucket.labels" . | nindent 4 }}
+spec:
+  capacity:
+    storage: 1Gi
```
kennymacleod, line 10: this should be an entry in `values.yaml`

src/main/charts/bitbucket/values.yaml
```diff
@@ -186,12 +186,24 @@ volumes:
...
+        # -- Specifies the path exported by the NFS server, used in the mount command
+        path: ""
     persistentVolumeClaim:
       # -- If true, then a PersistentVolumeClaim will be created for the shared-home volume.
       create: false
       # -- Specifies the name of the storage class that should be used for the shared-home volume claim.
       storageClassName:
       # -- Specifies the standard Kubernetes resource requests and/or limits for the shared-home volume claims.
+      volumeName:
+      # -- Specifies the name of the persistent volume to claim
```
kennymacleod, line 206: These lines appear to be in the wrong order (comment goes before value) and in the wrong place (should be before the previous comment)

src/test/config/bitbucket/values-EKS.yaml
```diff
@@ -12,13 +12,23 @@ volumes:
...
+      create: true
+      mountOptions:
+        - rw
+        #    - nfsvers=3 #todo DCNG-946
+        - lookupcache=pos
+        - noatime
+        - intr
+        - rsize=32768
+        - wsize=32768
+        - _netdev
```
kennymacleod, line 25: would it make sense to include any of these values in the PV template?

src/main/charts/bitbucket/templates/shared-home-pvc.yaml
```diff
@@ -9,7 +9,14 @@ spec:
   accessModes:
     - ReadWriteMany
   {{- if .Values.volumes.sharedHome.persistentVolumeClaim.storageClassName }}
+    {{- if eq "nil" .Values.volumes.sharedHome.persistentVolumeClaim.storageClassName  }}
```
kennymacleod, line 12: Using "nil" as a string it a bit wird. Why not just use 

    {{- with .Values.volumes.sharedHome.persistentVolumeClaim.storageClassName  }} 
      storageClassName: {{ . | quote }}
    {{- end }}

src/test/scripts/helm_install.sh
```diff
@@ -3,6 +3,18 @@
...
+
+  case "${currentContext}" in
+    *eks*) echo EKS;;
+    *aks*) echo AKS;;
+    *gke*) echo GKE;;
+    *shared-dev*|*default-context*) echo KITT;;
+    *) echo CUSTOM;;
+  esac
+}
+
```
kennymacleod, line 17: Why move this? Isn't the bash version check logic here because of the use of this syntax?

src/test/scripts/helm_uninstall.sh
```diff
@@ -15,12 +15,20 @@ helm uninstall -n "${TARGET_NAMESPACE}" "${FUNCTEST_RELEASE_NAME}"
...
+  # we are using a "shared shared home", remove the subdirectory of the shared-home
+  "$THISDIR"/shared_home_browser_install.sh "${TARGET_NAMESPACE}"
+  kubectl exec -n "${TARGET_NAMESPACE}" shared-home-browser -- rm -rf "/shared-home/$PRODUCT_RELEASE_NAME"
+fi
+
+# delete any and all persistent volumes and claims by label
+# todo DCNG-947
+echo Deleting PVCs, if the uninstall script gets stuck deleting the shared pvc here, run:
+sharedPvcName=$(kubectl get -n dcng pvc -l app.kubernetes.io/instance=pbruski-bitbucket | grep shared | awk '{print $1;}')
+echo kubectl patch pvc -n "${TARGET_NAMESPACE}" "$sharedPvcName" -p "'{\"metadata\":{\"finalizers\":null}}'"
```
kennymacleod, line 28: Why is the below line not enough to delete the PVC?

kennymacleod (commented):
I'm not convinced that including a PV template template in a Helm chart is a good idea. PVs are almost always going to be provisioned and managed independently of the Helm chart lifecycle (e.g. you don't want to PV to be deleted and the data nuked just bcause you uninstalled the chart).

Note that the Helm charts as they currently are already allow an NFS server to be used, without going to the bother of adding  PV for it, e.g.

```
volumes:
  sharedHome:
    customVolume:
      nfs: 
```
    
I would suggest that we shelve the PV template for now, and just use the inline syntax above. It's good enough for NFS testing.

pbruski:
> 
> Note that the Helm charts as they currently are already allow an NFS server to be used, without going to the bother of adding PV for it, e.g.
> 
> ```
> volumes:
>   sharedHome:
>     customVolume:
>       nfs: 
> ```
> 
> I would suggest that we shelve the PV template for now, and just use the inline syntax above. It's good enough for NFS testing.

It's not. For BB, you need to add custom mount options and that can only happen if you define a PV separately. The documentation we currently have in Config.md is misleading and I suggest we remove it - or at least indicate that it's not supposed to be used for BB.

For supported BB installations, 100% of them is going to be used with NFS. I think it makes sense to limit the Kubernetes-side configuration to IP/path (and potentially mount options, but perhaps these should default to the recommended ones). If we don't support that, we'd be forcing the customers to set up PV/PVC on their own without a good reason for that (well, maybe except for consistency with other products - but maybe we should discuss inclusion of BYO NFS templates there too).

In case of NFS, PV deletion does not have any effects on the exported file system.

src/main/charts/bitbucket/templates/shared-home-pv.yaml
pbruski: ATM, it's not used and exists only to satisfy Kubernetes validation rules, adding it to values would be misleading.

src/main/charts/bitbucket/values.yaml
pbruski: Right!

src/test/config/bitbucket/values-EKS.yaml
pbruski: Hmm, I think it does.

src/test/scripts/helm_install.sh
pbruski: The version check will be executed before the function is parsed anyway, regardless of location.

src/test/scripts/helm_uninstall.sh
pbruski: What I saw several times was that for no good reason, the pods/PVs/PVCs were stuck in terminating state. This was probably caused by incorrect order of deletion. When that happens, the finalizers need to be removed from the object. I want to see if that happens in CI too, if it does then we'll need to come up with a solution.

src/main/charts/bitbucket/templates/shared-home-pvc.yaml
pbruski: Will it work? What I need to have here is a literal empty string if the storage class is defined as an empty string. I think `with` will skip the stanza completely. I gave it a shot at https://play.golang.org/ and it seems like `with` works like an `if` in this case.

src/test/scripts/helm_uninstall.sh
pbruski: BTW, this is not specific to NFS, earlier I saw that with EFS too.

src/main/charts/bitbucket/templates/shared-home-pv.yaml
bianchi2: I agree with @kennymacleod about not including a PV. A couple of reasons here:
* like @kennymacleod said, flushing data when the chart is uninstalled isn't good
* it's common to setup up automatic PV provisioners in K8s clusters, so, we just create PVC and a provisioner will create a PV
* it's common that people who install and manage apps installed from Helm charts do not have permissions to create PVs which are requested from infra team on a per-case basis
* PV is this sort of an object that is not updated like stateful set or some configmap on helm upgrade. It's more like part of K8s infra to me even though it's an API object.It is also not related to the app itself.
* if we do want to include a PV template we need to make it all configurable then, not just limit to NFS. I mean let a user provide annotations (very important!) and PV spec in values.yaml so that a user can create a PV in AKS, EKS, GKE etc. If we do that, it'd be great not to track it in Helm after creation, i.e. not update/delete (maybe pre-install hook, or smth like that)

src/test/scripts/helm_uninstall.sh
bianchi2: Deleting a PV first should solve the problem, though making a PVC unprotected via the above patch work too. Yes, it's not related to storage backend.

src/main/charts/bitbucket/templates/shared-home-pvc.yaml
kcichy-atlassian: Can you use `with` with `else`? This seems to work:

```
package main

import (
	"os"
	"text/template"
)

func main() {

	instanceWithValue := DataStruct{"bar"}
	instanceWithoutValue := DataStruct{}

	tmpl, err := template.New("templ").Parse(`
{{ with .Foo}}
Foo is {{ $.Foo }}
{{ else }}
Foo is empty
{{ end }}`)
	
	if err != nil {
		panic(err)
	} else {
		tmpl.Execute(os.Stdout, instanceWithValue)
		tmpl.Execute(os.Stdout, instanceWithoutValue)
	}
}

type DataStruct struct {
	Foo string
}
```

src/main/charts/bitbucket/templates/shared-home-pv.yaml
pbruski: Eugene, thank you for providing a real world perspective on the how charts should look like. I understand that with your experience with proper Kubernetes deployments, you're looking at it from a different angle than I do. I think we're in a slightly different scenario, since we're packaging an app that (at least for some cases) has a hard requirement on NFS, specifically version 3 with very specific mount options, otherwise it won't work correctly.

To address your points:
> like @kennymacleod said, flushing data when the chart is uninstalled isn't good
It's not flushed in case of NFS.

> it's common to setup up automatic PV provisioners in K8s clusters, so, we just create PVC and a provisioner will create a PV
That's fine, we support that now. It's not a reason to not make it easier to support an NFS server (potentially) not tied to Kubernetes.

> it's common that people who install and manage apps installed from Helm charts do not have permissions to create PVs which are requested from infra team on a per-case basis

I don't think it's a reason to not support PV creation. In the scenario you've outlined, the PV definition can be used by the infra team to create a PV with precisely the parameters that we require in BBS. The app team can then go the PVC route, which we also support.

> PV is this sort of an object that is not updated like stateful set or some configmap on helm upgrade. It's more like part of K8s infra to me even though it's an API object.It is also not related to the app itself.

I understand that this is true in general, but I don't think that's correct in this case, where we're more or less providing an representation of an external infra as a Kubernetes object specifically for the purposes of our deployment. For example, we need the PV to contain specific mount options for NFS (otherwise especially BBS will not function correctly). If we start supporting newer NFS versions or require different options, we need to update the PV definition with the chart. I think this makes perfect sense (caveat: for NFS only).

> we do want to include a PV template we need to make it all configurable then, not just limit to NFS.

For the time being, we support 'everything' through the PVCs. We also have a specific case that makes it easier to use NFS, which is the only supported option for BBS and simply has to be used there to have a supported setup. Personally, I think supporting PVCs and a special case for NFS PV is good enough until the feature requests start trickling in. And that's where the product managers will step in.

src/main/charts/bitbucket/templates/shared-home-pvc.yaml
pbruski: But it's a tristate value. Empty string has a different meaning than no value and a different meaning than some value.

pbruski: I ended up doing:

`  {{- if not (kindIs "invalid" .Values.volumes.sharedHome.persistentVolumeClaim.storageClassName) }}
  storageClassName: {{ .Values.volumes.sharedHome.persistentVolumeClaim.storageClassName | quote }}
  {{- end }}`

src/test/config/bitbucket/values-EKS.yaml
pbruski: I've added all of these to the PV template and changed the NFS to v3 in the other PR.

PR #45, opened 2020-12-21
DCNG-945 set up an nfs server with the bitbucket server test

pbruski:


bkwiatek-atlassian (approved):

src/test/config/bitbucket/values-EKS.yaml
```diff
@@ -16,14 +16,16 @@ volumes:
       create: true
       mountOptions:
         - rw
-        #    - nfsvers=3 #todo DCNG-946
+        - nfsvers=4.1
+#        - nfsvers=3
```
kennymacleod, line 20: I thought BBS was NFSv3 only?

pbruski: If 4.1 turns out to be good enough for functional tests, let's stick to it. The problem with v3 is that we'll have to handle locks separately - in v4, they are a part of the protocol, in v3, they need a separate port and a daemon. Normally, this isn't a problem, but in Kubernetes this means that we need to add more configuration to the node itself to port-lock the deamon and I'd like to avoid it. For customers, this is unlikely going to be a problem because I'd expect they will be provisioning NFS server outside of Kubernetes.

PR #52, opened 2021-01-12
DCNG-890 - Fix Bitbucket test failure for 2 nodes

pbruski:
Surprise, surprise, we need direct node access for BB tests too (otherwise user deletion is problematic because it has -  an intended - cluster propagation delay).

### DIFF
--- a/src/test/scripts/helm_install.sh
+++ b/src/test/scripts/helm_install.sh
@@ -59,11 +59,21 @@ if grep -q nfs: ${chartValueFiles} /dev/null; then
     echo This configuration requires a private NFS server, starting...
     nfsServerPodName="${PRODUCT_RELEASE_NAME}-nfs-server"
     startNfsServer "${PRODUCT_RELEASE_NAME}" "${nfsServerPodName}"
-    nfsServerIp=$(kubectl get pods -n $TARGET_NAMESPACE "$nfsServerPodName" -o json | jq -r .status.podIP)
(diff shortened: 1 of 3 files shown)


kennymacleod (approved):

src/test/scripts/helm_install.sh
```diff
@@ -59,11 +59,21 @@ if grep -q nfs: ${chartValueFiles} /dev/null; then
...
     startNfsServer "${PRODUCT_RELEASE_NAME}" "${nfsServerPodName}"
-    nfsServerIp=$(kubectl get pods -n $TARGET_NAMESPACE "$nfsServerPodName" -o json | jq -r .status.podIP)
-    if [ -z "$nfsServerIp" ]; then
-      echo NFS server not found.
-      exit 1
-    fi
+
+    for ((try = 0; try < 60; try++)) ; do
+      echo Detecting NFS server IP...
+      nfsServerIp=$(kubectl get pods -n $TARGET_NAMESPACE "$nfsServerPodName" -o json | jq -r .status.podIP)
```
kcichy-atlassian, line 65: Do we have `jq` available on Bamboo?

kcichy-atlassian (approved):

src/test/scripts/helm_install.sh
pbruski: we do, I already have a green branch of that running.

PR #55, opened 2021-01-18
DCKUBE-44: Generify parameters

nanux:
Enable translation of the `values-KITT.yaml` file with different values to deploy in a different namespace.

Example:
`mvn clean install -Dshared.pvc.name=efs-pvc-dcd -Dkubernetes.target.namespace=dcd`

### DIFF
--- a/src/test/config/bitbucket/values-KITT.yaml
+++ b/src/test/config/bitbucket/values-KITT.yaml
@@ -21,7 +21,7 @@ volumes:
   sharedHome:
     customVolume:
       persistentVolumeClaim:
-        claimName: efs-claim # Pre-provisioned, and shared by all of our pods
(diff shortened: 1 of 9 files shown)


kennymacleod (approved):

pbruski (approved):

src/test/config/bitbucket/values-KITT.yaml
```diff
@@ -21,7 +21,7 @@ volumes:
   sharedHome:
     customVolume:
       persistentVolumeClaim:
-        claimName: efs-claim # Pre-provisioned, and shared by all of our pods
+        claimName: ${shared.pvc.name} # Pre-provisioned, and shared by all of our pods
```
pbruski, line 24: Hmm, maybe we should prefix it with a namespace instead? One less property to worry about. Something like
`${namespace}-atlassian-dc-shared-home`

PR #57, opened 2021-01-18
DCNG-976 add support for EFK (Elasticsearch/Fluentd/Kibana) stack

kennymacleod:
Customers using our Helm charts are almost certainly going to want a sensible log aggregation solution. For innovation week, I've added one to the Confluence chart.

The solution consists of 3 components:

1. A `fluentd` sidecar attached to each Confluence pod. This mounts the `local-home` volume, tails the various Confluence log files, tags them, and sends them on to Elasticsearch
2. An Elasticsearch deployment running in the same Kubernetes namespace
3. A Kibana deployment running in the same Kubernetes namespace, that queries and visualises the logs stored in ES

By default this is all turned off. Customers who want to use it can enable the fluentd sidecar, and provide their own deployments of ES and Kibana.

The shared Kibana in the DCNG namespace is available at https://kibana-dcng.internal.shared-dev.us-west-2.kitt-inf.net/


src/main/charts/confluence/templates/_fluentd_templates.tpl
```diff
@@ -0,0 +1,37 @@
+{{- define "fluentd.container" -}}
```
kennymacleod, line 1: Rather than drop the YAML for the fluentd sidecar container and config volume into `statefulset.yaml`, I've defined them as templates in this new file, which are then included as fragments in the statefulset.

src/main/charts/confluence/templates/_fluentd_templates.tpl
```diff
@@ -0,0 +1,37 @@
...
+{{ if .Values.fluentd.enabled }}
+- name: fluentd
+  image: {{ .Values.fluentd.imageName }}
+  command: ["fluentd", "-c", "/fluentd/etc/fluent.conf", "-v"]
+  volumeMounts:
+    - name: local-home
+      mountPath: /application-data
+      readOnly: true
```
kennymacleod, line 9: Mount the pod's `local-home` volume so that fluentd has access to the application's log files.

src/main/charts/confluence/templates/_fluentd_templates.tpl
```diff
@@ -0,0 +1,37 @@
...
+    - name: fluentd-config
+      mountPath: /fluentd/etc
+      readOnly: true
```
kennymacleod, line 12: Inject the fluentd config as a volume

src/main/charts/confluence/templates/_fluentd_templates.tpl
```diff
@@ -0,0 +1,37 @@
...
+    - name: POD_NAMESPACE
+      valueFrom:
+        fieldRef:
+          fieldPath: metadata.namespace
+    - name: POD_IP
+      valueFrom:
+        fieldRef:
+          fieldPath: status.podIP
+    - name: HELM_RELEASE_NAME
+      value: {{ include "confluence.fullname" . }}
```
kennymacleod, line 27: These env vars are injected into the sidecar so that fluentd can add them as log record fields

src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -153,10 +153,6 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end -}}
 {{- end }}
 
-{{- define "confluence.sysprop.disableHomeLogAppender" -}}
--DConfluenceHomeLogAppender.disabled=true
-{{- end }}
```
kennymacleod, line 158: now that fluentd is reading the confluence log files directly, we don't need this sysprop any more.

src/main/charts/confluence/templates/configmap-fluentd.yaml
```diff
@@ -0,0 +1,40 @@
+{{ if .Values.fluentd.enabled }}
+apiVersion: v1
+kind: ConfigMap
```
kennymacleod, line 3: Here we define a `ConfigMap` which contains the `fluentd.conf` file. This in turn is injected into the pods as a mounted volume.

src/main/charts/confluence/templates/configmap-fluentd.yaml
```diff
@@ -0,0 +1,40 @@
...
+metadata:
+  name: {{ include "confluence.fullname" . }}-fluentd-config
+  labels:
+  {{- include "confluence.labels" . | nindent 4 }}
+data:
+  fluent.conf: |
```
kennymacleod, line 9: The below config should be taken very much as a first attempt. It's the simplest thing that'll work, and probably needs tweaking for prod usage.

src/main/charts/confluence/templates/configmap-fluentd.yaml
```diff
@@ -0,0 +1,40 @@
...
+    <source>
+      @type tail
+      path /application-data/logs/*.log
```
kennymacleod, line 12: assumes that this is where the local-home volume is mounted

src/main/charts/confluence/templates/configmap-fluentd.yaml
```diff
@@ -0,0 +1,40 @@
...
+      read_from_head true
+      tag applogs
+      <parse>
+        @type regexp
+        expression /^(?<time>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3}) (?<level>[A-Z]+) \[(?<thread>\S+)\] \[(?<logger>\S+)\] (?<method>[\w|\$]+) (?<message>.*)$/
```
kennymacleod, line 17: This regex works for the log4j pattern used by Confluence

src/main/charts/confluence/templates/statefulset.yaml
```diff
@@ -71,6 +71,10 @@ spec:
             - name: ATL_TOMCAT_SECURE
               value: "true"
             {{ end }}
+            - name: UMASK
+              value: {{ .Values.confluence.umask | quote }}
```
kennymacleod, line 75: By default, the Confluence docker image creates its log files with file permissions so that only the Confluence user can read them. Change the umask so that they can be read by anyone.

We're assuming here that random punters can't just poke around on the filesystem, so we don't really need restrictive file permissions, but it can always be changed using the values file.

src/main/charts/confluence/templates/statefulset.yaml
```diff
@@ -71,6 +71,10 @@ spec:
...
+            - name: SET_PERMISSIONS
+              value: "false"
```
kennymacleod, line 77: Set this to false to prevent the Docker image from locking down the local-home permissions. Without this, fluentd can't read the files.

src/main/charts/confluence/values.yaml
```diff
@@ -179,6 +182,15 @@ ingress:
   # Optional if Ingress Controller is configured to use one secret for all ingresses
   tlsSecretName:
 
+fluentd:
+  # -- True if the fluentd sidecar should be added to each pod
+  enabled: false
+  # -- The name of the image containing the fluentd sidecar
+  imageName: atlassian/fluentd-elasticsearch:1.0
```
kennymacleod, line 189: Note - this image doesn't actually exist yet. Will need to publish it before we release this.

src/test/config/confluence/values-KITT.yaml
```diff
@@ -26,4 +26,10 @@ volumes:
         claimName: efs-claim # Pre-provisioned, and shared by all of our pods
     subPath: ${helm.release.prefix}-confluence # Since all of our pods share the same EFS PV, we use subpath mounts to prevent interference
     nfsPermissionFixer:
-      enabled: true
\ No newline at end of file
+      enabled: true
+
+fluentd:
+  enabled: true
+  imageName: docker.atl-paas.net/kmacleod/fluentd:1
```
kennymacleod, line 33: temporary fluentd image for testing

src/test/config/confluence/values-KITT.yaml
```diff
@@ -26,4 +26,10 @@ volumes:
...
+  elasticsearch:
+    hostname: elasticsearch-master
```
kennymacleod, line 35: assumes that elasticsearch has already been deployed under this name

src/main/charts/confluence/templates/_fluentd_templates.tpl
bianchi2: Do we need to mount an entire confluence home? Does `/application-data/logs` make more sense? Also, our default is https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/main/charts/confluence/values.yaml#L200. Perhaps it makes sense to use `{{ .Values.volumes.localHome.persistentVolumeClaim.mountPath }}` rather than hardcode the path since we leave it configurable anyway?

src/main/charts/confluence/values.yaml
bianchi2: Can't we use https://hub.docker.com/r/fluent/fluentd-kubernetes-daemonset/tags?page=1&ordering=last_updated?

src/main/charts/confluence/templates/_fluentd_templates.tpl
kennymacleod: > Do we need to mount an entire confluence home? Does /application-data/logs make more sense?

Good idea. I've added a `subPath` to the `volumeMount` so that only the logs subdir is mounted.

As for the mount path itself, we *could* use the same path as the confluence container, but I don't think it adds anything, and it suggests that they need to be the same, when in fact they don't.

src/main/charts/confluence/values.yaml
kennymacleod: Maybe? I'm not really sure what's in it.

bianchi2: Worth giving it a go? 
```
docker run -ti --entrypoint=bash fluent/fluentd-kubernetes-daemonset:v1.11.5-debian-elasticsearch7-1.2
root@6479773cb876:/home/fluent#
root@6479773cb876:/home/fluent# gem list | grep "fluent-plugin-elasticsearch"
fluent-plugin-elasticsearch (4.3.3)
```
That's an official repo, and this tag seems to have what we need. Makes things simpler for us.

kennymacleod: Seems to work! Thanks!

bianchi2 (approved):

pbruski (approved):

PR #58, opened 2021-01-18
DCNG-977 Capture Jira+Confluence access logs into local-home volume

kennymacleod:
Due to the way that the Jira+Confluence applications are constructed, with the application deployed as a WAR into Tomcat, the Tomcat logs (including the access logs) are stored in a different place in the filesystem to the application logs. 

Currently, the application logs are a sub-directory of the local-home volume, but the tomcat+access logs are just being stored on the container's internal filesystem.

This PR captures those tomcat+access logs into the local-home volume. It does this by double-mounting the local-home volume on to two separate paths in the container. The first mount point is the existing one (which captures the application logs due to them being in the `logs` sub-directory), but now I've added another mount point for the local-home volume, so that the Tomcat logs directory is also mounted on the logs subdirectory of local-home.

It's a bit peculiar to double-mount the same volume twice, but it's perfectly valid. The end result is that the same logs subdirectory of the local-home volume now contains both sets of logs, and the customer no longer has to worry about where to look for them.

### DIFF
--- a/src/main/charts/confluence/templates/_helpers.tpl
+++ b/src/main/charts/confluence/templates/_helpers.tpl
@@ -176,6 +176,19 @@ The command that should be run by the nfs-fixer init container to correct the pe
 {{- end }}
 {{- end }}
 
+{{ define "confluence.volumeMounts" }}
+- name: local-home
+  mountPath: {{ .Values.volumes.localHome.mountPath | quote }}
+- name: local-home
+  mountPath: {{ .Values.confluence.accessLog.mountPath | quote }}
+  subPath: {{ .Values.confluence.accessLog.localHomeSubPath | quote }}
+- name: shared-home
+  mountPath: {{ .Values.volumes.sharedHome.mountPath | quote }}
+  {{- if .Values.volumes.sharedHome.subPath }}
+  subPath: {{ .Values.volumes.sharedHome.subPath | quote }}
+  {{- end }}
+{{ end }}
+
 {{/*
 For each additional library declared, generate a volume mount that injects that library into the Confluence lib directory
 */}}
--- a/src/main/charts/confluence/templates/_helpers.tpl
+++ b/src/main/charts/confluence/templates/_helpers.tpl
@@ -176,6 +176,11 @@ The command that should be run by the nfs-fixer init container to correct the pe
 {{- end }}
 {{- end }}
 
+{{/*
+Defines the volume mounts used by the Confluence container.
+Note that the local-home volume is mounted twice, once for the local-home directory itself, and again
+on Tomcat's logs directory. THis ensures that Tomcat+Confluence logs get captured in the same volume.
+*/}}
 {{ define "confluence.volumeMounts" }}
 - name: local-home
   mountPath: {{ .Values.volumes.localHome.mountPath | quote }}
--- a/src/main/charts/confluence/templates/statefulset.yaml
+++ b/src/main/charts/confluence/templates/statefulset.yaml
@@ -52,13 +52,7 @@ spec:
           {{- toYaml . | nindent 12 }}
           {{- end }}
           volumeMounts:
-            - name: local-home
-              mountPath: {{ .Values.volumes.localHome.mountPath | quote }}
-            - name: shared-home
-              mountPath: {{ .Values.volumes.sharedHome.mountPath | quote }}
-              {{- if .Values.volumes.sharedHome.subPath }}
-              subPath: {{ .Values.volumes.sharedHome.subPath | quote }}
-              {{- end }}
+            {{- include "confluence.volumeMounts" . | nindent 12 }}
             {{- with .Values.confluence.additionalVolumeMounts }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
@@ -71,6 +65,8 @@ spec:
             - name: ATL_TOMCAT_SECURE
               value: "true"
             {{ end }}
+            - name: ATL_TOMCAT_ACCESS_LOG
+              value: {{ .Values.confluence.accessLog.enabled | quote }}
             - name: ATL_PRODUCT_HOME_SHARED
               value: {{ .Values.volumes.sharedHome.mountPath | quote }}
             - name: JVM_SUPPORT_RECOMMENDED_ARGS
--- a/src/main/charts/confluence/values.yaml
+++ b/src/main/charts/confluence/values.yaml
@@ -79,7 +79,7 @@ confluence:
 
   accessLog:
     # -- True if access logging should be enabled.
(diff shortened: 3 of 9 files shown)


src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -176,6 +176,19 @@ The command that should be run by the nfs-fixer init container to correct the pe
...
+  mountPath: {{ .Values.volumes.localHome.mountPath | quote }}
+- name: local-home
+  mountPath: {{ .Values.confluence.accessLog.mountPath | quote }}
+  subPath: {{ .Values.confluence.accessLog.localHomeSubPath | quote }}
+- name: shared-home
+  mountPath: {{ .Values.volumes.sharedHome.mountPath | quote }}
+  {{- if .Values.volumes.sharedHome.subPath }}
+  subPath: {{ .Values.volumes.sharedHome.subPath | quote }}
+  {{- end }}
+{{ end }}
```
kennymacleod, line 190: The `volumeMounts` section of `statefulset.yaml` was getting a bit verbose, so I've extracted it into its own template fragment. It's the same as before, but now with the 2nd `local-home` mount added in.

src/main/charts/confluence/templates/statefulset.yaml
```diff
@@ -71,6 +65,8 @@ spec:
             - name: ATL_TOMCAT_SECURE
               value: "true"
             {{ end }}
+            - name: ATL_TOMCAT_ACCESS_LOG
+              value: {{ .Values.confluence.accessLog.enabled | quote }}
```
kennymacleod, line 69: The Confluence Docker image allows the access log to be turned on/off, so I've added that knob.

src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -176,6 +176,19 @@ The command that should be run by the nfs-fixer init container to correct the pe
 {{- end }}
 {{- end }}
 
+{{ define "confluence.volumeMounts" }}
```
nanux, line 179: Could we add some documentation why we are mounting into two places under the same name (application and access logs availability)?

src/main/charts/confluence/values.yaml
```diff
@@ -77,6 +77,14 @@ confluence:
     # -- The number of consecutive failures of the Confluence container readiness probe before the pod fails readiness checks
     failureThreshold: 30
 
+  accessLog:
+    # -- True if access logging should be enabled.
+    enabled: false
```
nanux, line 82: For reference, the access logs are disabled in the docker image by default.
`ATL_TOMCAT_ACCESS_LOG (default: false)`
https://hub.docker.com/r/atlassian/confluence-server/

src/main/charts/confluence/values.yaml
```diff
@@ -77,6 +77,14 @@ confluence:
...
+    # -- The path within the Confluence container where the local-home volume should be mounted in order to capture access logs.
+    mountPath: "/opt/atlassian/confluence/logs"
+    # -- The subdirectory within the local-home volume where access logs should be stored.
+    localHomeSubPath: "logs"
```
nanux, line 86: Does it mean, the access logs will be in `/opt/atlassian/confluence/logs/logs` directory?

src/main/charts/jira/values.yaml
```diff
@@ -60,6 +60,12 @@ jira:
     # -- The number of consecutive failures of the Jira container readiness probe before the pod fails readiness checks
     failureThreshold: 30
 
+  accessLog:
+    # -- The path within the Jira container where the local-home volume should be mounted in order to capture access logs.
+    mountPath: "/opt/atlassian/jira/logs"
+    # -- The subdirectory within the local-home volume where access logs should be stored.
+    localHomeSubPath: "log"
```
nanux, line 67: The same as with confluence, is the access log in `/opt/atlassian/confluence/logs/logs`?

nanux (approved):
I was trying to find some documentation regarding the double-mounting of the volume in the container `spec`. Unfortunately without success. Do you have something I can read about it? I haven't seen it before.
Does it have any implication on the container or pod scheduling?

src/main/charts/jira/values.yaml
bianchi2:
```
root@jira-0:/var/atlassian/application-data/jira/log# ls
access_log.2021-01-19  atlassian-jira-gc-2021-01-19_05-40-50.log  atlassian-jira.log  catalina.2021-01-19.log  host-manager.2021-01-19.log  localhost.2021-01-19.log  manager.2021-01-19.log
root@jira-0:/var/atlassian/application-data/jira/log# pwd
/var/atlassian/application-data/jira/log
```
It's going to be log directory in a local-home volume.

bianchi2 (approved):

src/main/charts/jira/templates/_helpers.tpl
```diff
@@ -93,6 +93,19 @@ The command that should be run by the nfs-fixer init container to correct the pe
 {{- end }}
 {{- end }}
 
+{{ define "jira.volumeMounts" }}
+- name: local-home
+  mountPath: {{ .Values.volumes.localHome.mountPath | quote }}
+- name: local-home
```
pbruski, line 99: I wonder if we really need two mounts. I know it's cleaner from the perspective of the sidecar, but does that 'perspective' matter? It does make the charts more complicated.

pbruski (approved):

src/main/charts/confluence/values.yaml
kennymacleod: No. The `subPath` property of `volumeMount` refers to the subdirectory of the volume that should be mounted on `mountPath`. So the `logs` directory inside the `local-home` volume gets mounted on to the `/opt/atlassian/confluence/logs` directory in the container.

src/main/charts/jira/templates/_helpers.tpl
kennymacleod: The alternative is to have an extra volume just to mount the tomcat logs, and for consistency that extra volume would need to be a PV. That's a lot of extra baggage when we can just double-mount the same volume instead.

src/main/charts/confluence/templates/_helpers.tpl
kennymacleod: Done

src/main/charts/confluence/values.yaml
kennymacleod: I've just changed the `values.yaml` file to default this to `true`, since access logs are always enabled for Jira and Bitbucket. Te Docker images may be inconsistent, but we can paper over the cracks here.

PR #65, opened 2021-01-21
DCNG-976 use http source for Confluence fluentd events

kennymacleod:
When Confluence is configured to use the fluentd log appender, it will post log events directly to the fluentd sidecar, avoiding the need to use a complicated regex to parse the human-readable log file.

### DIFF
--- a/src/main/charts/confluence/templates/_fluentd_templates.tpl
+++ b/src/main/charts/confluence/templates/_fluentd_templates.tpl
@@ -7,10 +7,6 @@
     - containerPort: {{ .Values.fluentd.httpPort }}
       protocol: TCP
   volumeMounts:
-    - name: local-home
-      mountPath: /application-data/logs
-      subPath: logs
-      readOnly: true
     - name: fluentd-config
       mountPath: /fluentd/etc
       readOnly: true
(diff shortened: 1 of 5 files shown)


src/main/charts/confluence/templates/_fluentd_templates.tpl
```diff
@@ -3,6 +3,9 @@
 - name: fluentd
   image: {{ .Values.fluentd.imageName }}
   command: ["fluentd", "-c", "/fluentd/etc/fluent.conf", "-v"]
+  ports:
+    - containerPort: {{ .Values.fluentd.httpPort }}
+      protocol: TCP
```
kennymacleod, line 8: Note that pod containers don't *need* to declare their ports, but it's good practice to do so.

src/main/charts/confluence/templates/configmap-fluentd.yaml
```diff
@@ -8,14 +8,9 @@ metadata:
...
   fluent.conf: |
     <source>
-      @type tail
-      path /application-data/logs/atlassian-confluence.log*
-      read_from_head true
-      tag confluence.log
-      <parse>
-        @type regexp
-        expression /^(?<time>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3}) (?<level>[A-Z]+) \[(?<thread>\S+)\] \[(?<logger>\S+)\] (?<method>[\w|\$]+) (?<message>.*)$/
-      </parse>
```
kennymacleod, line 18: We can now remove this insider knowledge of the Confluence log format, now we just accept the JSON event whatever it may be

pbruski (approved):

PR #66, opened 2021-01-22
DCNG-702 Allow Jira license to be supplied by Kubernetes secret

bkwiatek-atlassian:


src/main/charts/jira/templates/statefulset.yaml
```diff
@@ -42,6 +42,13 @@ spec:
               {{ end }}
             {{- include "jira.databaseEnvVars" . | nindent 12 }}
             {{- include "jira.clusteringEnvVars" . | nindent 12 }}
+            {{ with .Values.jira.license.secretName }}
+            - name: JIRA_SETUP_LICENSE
+              valueFrom:
+                secretKeyRef:
+                  name: { { . } }
+                  key: { { $.Values.jira.license.secretKey } }
```
kennymacleod, line 50: weird spacing of the `{{` and `}}` here, could you make it consistent with the rest of them?

kennymacleod (changes requested):
Formatting consistency

pbruski (approved):

src/main/charts/jira/templates/statefulset.yaml
bkwiatek-atlassian: Good catch, must have had auto-formatting during commit enabled.

PR #68, opened 2021-01-26
DCNG-784 Provision a GCP Kubernetes cluster

pbruski:


kennymacleod (approved):

src/test/scripts/helm_install.sh
```diff
@@ -6,11 +6,17 @@ set -x
 getCurrentClusterType() {
```
kcichy-atlassian, line 6: TIL Bash is even worse than I used to think. No return values and returning by echoing 🤦 .

src/test/scripts/helm_install.sh
```diff
@@ -6,11 +6,17 @@ set -x
...
   local serverAddress=$(kubectl config view --minify -o json | jq -r '.clusters[0].cluster.server')
 
-  case "${serverAddress}" in
+  local clusterType=$(case "${serverAddress}" in
     *.eks.amazonaws.com*) echo EKS;;
     *.azmk8s.io*) echo AKS;;
-    *gke*) echo GKE;;
     *.kitt-inf.net*) echo KITT;;
+  esac)
+  [ "$clusterType" ] && { echo $clusterType ; return ; }
```
kcichy-atlassian, line 14: How about replacing this trick with immediate returns?
```
  case "${serverAddress}" in
    *.eks.amazonaws.com*) echo EKS; return;;
    *.azmk8s.io*) echo AKS; return;;
    *gke*) echo GKE; return;;
    *.kitt-inf.net*) echo KITT; return;;
  esac

  case "$(kubectl config view --minify -o json | jq -r '.clusters[0].name')" in
    gke*) echo GKE; return;;
    *) echo CUSTOM; return;;
  esac
```
?

kcichy-atlassian (approved):

src/test/config/bitbucket/values-AKS.yaml
```diff
@@ -20,7 +20,7 @@ volumes:
     customVolume:
       persistentVolumeClaim:
         claimName: ${shared.pvc.name} # Pre-provisioned, and shared by all of our pods
-    subPath: ${helm.release.prefix}-bitbucket # Since all of our pods share the same EFS PV, we use subpath mounts to prevent interference
```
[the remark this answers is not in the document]
pbruski: Yes, we do - both are AWS-based.

src/test/scripts/helm_install.sh
pbruski: well, you can return a status. You can also return via a param outpu variable, I don't particularly like that syntax, though.

PR #69, opened 2021-01-26
Change the default pvc name to atlassian-dc-shared-home-pvc

pbruski:


pom.xml
```diff
@@ -82,7 +82,7 @@
 <!--        <eks.ingress.domain>eks.k8s.bbsperf.com</eks.ingress.domain>-->
         <eks.ingress.domain>dcngkube.bbsperf.com</eks.ingress.domain>
 
-        <shared.pvc.name>efs-claim</shared.pvc.name>
+        <shared.pvc.name>atlassian-dc-shared-home-pvc</shared.pvc.name>
```
kcichy-atlassian, line 85: I'm wondering, is this shared home shared between Jira and Confluence?

kcichy-atlassian (approved):

pom.xml
pbruski: It is. There's just one instance of the PV, shared by all products, all builds, all branches.

PR #73, opened 2021-02-02
delete the NFS service too, fixed a stray CR in fluentd include

pbruski:


src/main/charts/jira/templates/_helpers.tpl
```diff
@@ -236,7 +236,5 @@ volumeClaimTemplates:
 {{ end }}
 
 {{- define "jira.sysprop.fluentdAppender" -}}
-{{- if .Values.fluentd.enabled -}}
--Datlassian.logging.cloud.enabled=true
-{{- end -}}
+-Datlassian.logging.cloud.enabled={{.Values.fluentd.enabled}}
```
pbruski, line 239: "Alternatively, they can be written with the folded style (denoted by “>”) where each line break is folded to a space unless it ends an empty or a more-indented line."
So, if this ends up being an empty line, it will be preserved in the java command line.

kennymacleod (approved):

PR #76, opened 2021-02-15
Dump current status of all relevant kube resources for diagnostics

kennymacleod:
Add some extra diagnostics to `download_logs.sh`, so that we display the summary status of every resource in the Helm release.


src/test/scripts/download_logs.sh
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-set -x
```
kennymacleod, line 3: this is just too noisy

pbruski (changes requested):

kennymacleod:
Dropping in favour of #77

PR #77, opened 2021-02-16
DCNG-1045 Add java test to verify post-install state

kennymacleod:
Adds a new java test that gets run in the "final tasks" section of the bamboo plan, and which verifies the state of the pods and their containers in the product statefulset.

Hopefully this will be useful, but we can tweak it as necessary.

### DIFF
--- a/src/test/java/test/postinstall/Utils.java
+++ b/src/test/java/test/postinstall/Utils.java
@@ -1,0 +1,21 @@
+package test.postinstall;
+
+import io.vavr.collection.HashMap;
+import io.vavr.collection.Map;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


src/test/java/test/postinstall/Utils.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Properties;
+
+final class Utils {
+    static Map<String, String> readPropertiesFile(final Path fileLocation) throws IOException {
+        final var properties = new Properties();
+        try (var is = Files.newInputStream(fileLocation)) {
```
pbruski, line 14: we should probable use a reader instead

pbruski (approved):

PR #82, opened 2021-03-02
feat(confluence): managed synchrony option

Boojapho:
Feature for Confluence Helm chart that adds the ability to deploy Synchrony using a Confluence-managed configuration.  See https://confluence.atlassian.com/doc/possible-confluence-and-synchrony-configurations-958779064.html for details this configuration.

Changes included:
- Added `synchrony.managed` boolean to tell the Helm chart whether Synchrony should be deployed as a managed configuration or as a stand-alone cluster.  This is ignored if `synchrony.enabled` is `false`.
- If `synchrony.managed` is `true` and `synchrony.enabled` is `true`, the following are deployed:
   - JVM Option for `-Dsynchrony.btf.disabled=false`
   - Synchrony service (same as the stand-alone configuration).
   - Confluence statefulset with the following changes:
      - Synchrony `http` and `hazelcast` ports exposed.  The port names had to be renamed in order to avoid a conflict with Confluence's `http` and `hazelcast` ports
      - Additional Synchrony selector labels for Synchrony service
- Readme update using `helm-docs`
- Chart version bump

Testing notes: I manually tested Confluence in stand-alone mode only and verified `synchrony.enabled: false`, `synchrony.enabled: true` + `synchrony.managed: false`, and `synchrony.enabled: true` + `synchrony:managed: true`.  With Synchrony enabled, I tested with 2 users simultaneously editing a page.  I was using a PostgreSQL v10 database on the backend.

### DIFF
--- a/src/main/charts/confluence/Chart.yaml
+++ b/src/main/charts/confluence/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: confluence
 description: A chart for installing Confluence DC on Kubernetes
 type: application
-version: 0.1.0
+version: 0.1.1
 appVersion: 7.9.0-jdk11
 kubeVersion: ">=1.17.x-0"
 keywords:
--- a/src/main/charts/confluence/README.md
+++ b/src/main/charts/confluence/README.md
@@ -1,6 +1,6 @@
 # confluence
 
-![Version: 0.1.0](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 7.9.0-jdk11](https://img.shields.io/badge/AppVersion-7.9.0--jdk11-informational?style=flat-square)
+![Version: 0.1.1](https://img.shields.io/badge/Version-0.1.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 7.9.0-jdk11](https://img.shields.io/badge/AppVersion-7.9.0--jdk11-informational?style=flat-square)
 
 A chart for installing Confluence DC on Kubernetes
 
@@ -22,6 +22,9 @@ Kubernetes: `>=1.17.x-0`
 | additionalInitContainers | list | `[]` | Additional initContainer definitions that will be added to all Confluence pods |
 | additionalLabels | object | `{}` | Additional labels that should be applied to all resources |
 | affinity | object | `{}` | Standard Kubernetes affinities that will be applied to all Confluence and Synchrony pods |
+| confluence.accessLog.enabled | bool | `true` | True if access logging should be enabled. |
+| confluence.accessLog.localHomeSubPath | string | `"logs"` | The subdirectory within the local-home volume where access logs should be stored. |
+| confluence.accessLog.mountPath | string | `"/opt/atlassian/confluence/logs"` | The path within the Confluence container where the local-home volume should be mounted in order to capture access logs. |
 | confluence.additionalBundledPlugins | list | `[]` | Specifies a list of additional Confluence plugins that should be added to the Confluence container. These are specified in the same manner as the additionalLibraries field, but the files will be loaded as bundled plugins rather than as libraries. |
 | confluence.additionalEnvironmentVariables | list | `[]` | Defines any additional environment variables to be passed to the Confluence container. See https://hub.docker.com/r/atlassian/confluence-server for supported variables. |
 | confluence.additionalJvmArgs | list | `[]` | Specifies a list of additional arguments that can be passed to the Confluence JVM, e.g. system properties |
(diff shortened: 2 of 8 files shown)


badgersow:
Hi @Boojapho and thank you for the contribution.

Can you please clarify what problem are you trying to solve by adding this configuration option? I'm trying to understand if there is any value in having managed synchrony option and so far I don't see when it will provide any benefits. However, there are several downsides

* running both Confluence and Synchrony JVM in a single container is not the best industry practice
* we don't have the ability to scale synchrony independently
* supporting both deployment options have its cost

The reason why managed synchrony was originally implemented is to eliminate the complexity of managing separate synchrony cluster. The existence of helm charts make the initial problem go away (there is no complexity in synchrony cluster anymore), so we don't really need managed synchrony to solve this issue.

Hence is my question about your use case - is there any additional value which is worth the pain of supporting both options?

Boojapho:
@badgersow You stated the main reason "to eliminate the complexity of managing a separate synchrony cluster".   The customer I'm supporting does not want to manage Synchrony as a separate pod.  In addition, in the [documentation here](https://confluence.atlassian.com/doc/set-up-a-synchrony-cluster-for-confluence-data-center-958779073.html) it states that managed is recommended by Atlassian.

The customer that I'm supporting already has a managed Synchrony that was deployed using a different helm chart.  We are migrating them to this helm chart since it is coming from Atlassian.  I have to make the changes submitted in the MR to support their request.

To me, the helm chart is not a place to support or not support a feature.  If the application has a supported feature, like managed Synchrony, the helm chart should at least give the option to deploy the application that way.  If it were up to me and I had a new deployment, I would choose the stand-alone version, and think that is a better solution.  But unless Atlassian calls managed Synchrony deprecated, I think you need to support both options.

badgersow:
Thanks for the clarification. To be honest, I still don't see why managed synchrony might be better than having a separate `StatefulSet`. When you say that the customer "does not want to manage Synchrony as a separate pod", I wonder why this is the case - stateful set is supposed to take care of this and no managing is really required here. To the customer it shoudn't matter how the topology looks like. Unless I'm missing something.

@kennymacleod @nanux What do you guys think about that? If we merge this, we will need to make sure that our builds cover both options. I think it depends on our philosophy - do we want to be opinionated in helm charts or is our goal is to cover many different configurations? What do we do with helm charts of other products?

kennymacleod:
I recommend against this change. The only reason that Confluence currently supports self-managed Synchrony processes is for ease of deployment, i.e. it was onerous for customers to have to run a separate Synchrony process cluster. This rationale does not apply in a Kubernetes environment. As @badgersow says, there *is* no management overhead to running the additional StatefulSet for Synchrony.

jareddohrman:
As a customer of these helm charts, I would like to see this feature. Currently, we're facing issues with synchrony/hazelcast + istio-proxy (sidecar). Synchrony and Istio will not work together if Synchrony insists on binding the pod IP
https://istio.io/latest/docs/ops/deployment/requirements/#application-bind-address

Bonn93:
Adding some insights here. Istio is supposed to be transparent, and needing breaking changes and upstream changes to hazelcast as a whole doesn't seem to be on anyone's agenda, this is a pretty niche use case and can be navigated entirely with Istio until Hazelcast decide to bind to addresses differently, but seems like their architecture does not want this to avoid routing issues or binding to non-routable addresses. 

That being, there are some potential quirks and options you can try which have been successful in other environments with an Istio mesh in place. 

### wait until the proxy is running before trying to start + ignore hazelcast port
```
podAnnotations: {"traffic.sidecar.istio.io/excludeInboundPorts": "5701", "proxy.istio.io/config": '{ "holdApplicationUntilProxyStarts": "true" }'}
```

### Remove x-b3-sampled header due to CONFSERVER-59477
```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: {{ .Values.confluence.subdomain }}-{{ .Release.Name }}
spec:
  gateways:
  - {{ .Release.Name }}-gateway
  hosts:
  - {{ .Values.confluence.subdomain }}.{{ .Values.domain }}
  http: 
  - match:
    - port: {{ .Values.gateway.port.number }}
      uri:
        prefix: /synchrony
    headers:
      request:  #https://jira.atlassian.com/browse/CONFSERVER-59477
        remove:
        - x-b3-sampled
        add:
          x-b3-sampled: '0'
```

nanux:
@Boojapho and @jareddohrman have you had any success with the workaround described above? This PR is getting stale.

Thank you

jareddohrman:
@nanux, we've resorted to getting an exemption from gatekeeper seeing as Istio envoy doesn't like the fact synchrony uses PodIP - therefore Synchrony will not have a sidecar, and mTLS will not be enforced.

Boojapho:
I have resorted to making a fork of the helm chart with the modification needed to support managed synchrony.

pete-eams:
@Boojapho has this been tested in a clustered setup?

nanux:
Hi @Boojapho, as we have closed https://github.com/atlassian-labs/data-center-helm-charts/issues/81 and you have fixed for you, I will close this PR.

PR #83, opened 2021-03-05
DCKUBE-103: add a system to enable collab editing by default

badgersow:
This system property instructs Confluence to enable collab editing on fresh install if Synchrony is deployed in separate cluster. This property has been introduced in master branch of Confluence and so far it's used only by helm charts.

### DIFF
--- a/src/main/charts/confluence/templates/_helpers.tpl
+++ b/src/main/charts/confluence/templates/_helpers.tpl
@@ -149,6 +149,10 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 -Datlassian.logging.cloud.enabled={{.Values.fluentd.enabled}}
 {{- end }}
 
+{{- define "confluence.sysprop.enable.synchrony.by.default" -}}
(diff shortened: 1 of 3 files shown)


src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -149,6 +149,10 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 -Datlassian.logging.cloud.enabled={{.Values.fluentd.enabled}}
 {{- end }}
 
+{{- define "confluence.sysprop.enable.synchrony.by.default" -}}
+-Dsynchrony.by.default.enable.collab.editing.if.manually.managed=true
```
nanux, line 153: What would happen if this is used in Confluence that doesn't support the system property? Nothing?

nanux (approved):

src/main/charts/confluence/templates/_helpers.tpl
badgersow: Nothing will happen, as old Connie doesn't check that property

Issue #85, opened 2021-03-16
Additional variables do not seem to work properly


bordenit:
Here is an example configmap that does work....
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: jira-config
  labels:
    app: jira
data:
  JVM_MINIMUM_MEMORY: 2048m
  JVM_MAXIMUM_MEMORY: 4096m
  ATL_JDBC_USER: "{{ jiradbUsername }}"
  ATL_JDBC_PASSWORD: "{{ jiradbPassword }}"
  ATL_DB_DRIVER: org.postgresql.Driver
  ATL_JDBC_URL: jdbc:postgresql://postgres-jira:5432/jiradb
  CLUSTERED: "true"
  ATL_DB_TYPE: postgres72
  JIRA_SHARED_HOME: /var/atlassian/application-data/jira/shared
```

I tried adding ATL_TOMCAT_CONTEXTPATH variable using the helm chart and I think it maybe puts it in the wrong place. You could possibly group all the container variable options under data in configmap like the above.

bordenit:
Actually, this issue is that your readiness probe in the helm chart is not automatically updated if a contextpath is used. For example, if you have context path as /jira, then the readiness probe path should auto-update to /jira/status

nanux:
@bordenit I believe this was fixed by the above. Did it solve the issue for you?

bordenit:
Yes, it is resolved with the context path updates. Thanks.

PR #86, opened 2021-03-16
Set the context path and the readinessProbe Automatically

bordenit:
The helm chart currently does not allow for a context path to be used, as the readinessProbe will never become ready as it will not point to the correct context path. This change is intended to set the ATL_TOMCAT_CONTEXTPATH and the readinessProbe path dynamically based on the context path  provided in the values.yaml.

### DIFF
--- a/src/main/charts/confluence/templates/_helpers.tpl
+++ b/src/main/charts/confluence/templates/_helpers.tpl
@@ -149,6 +149,10 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 -Datlassian.logging.cloud.enabled={{.Values.fluentd.enabled}}
 {{- end }}
 
+{{- define "confluence.sysprop.enable.synchrony.by.default" -}}
+-Dsynchrony.by.default.enable.collab.editing.if.manually.managed=true
+{{- end -}}
+
 {{- define "confluence.sysprop.synchronyServiceUrl" -}}
 {{- if .Values.synchrony.enabled -}}
 -Dsynchrony.service.url={{ .Values.synchrony.ingressUrl }}/v1
--- a/src/main/charts/confluence/templates/_helpers.tpl
+++ b/src/main/charts/confluence/templates/_helpers.tpl
@@ -149,10 +149,6 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 -Datlassian.logging.cloud.enabled={{.Values.fluentd.enabled}}
 {{- end }}
 
-{{- define "confluence.sysprop.enable.synchrony.by.default" -}}
--Dsynchrony.by.default.enable.collab.editing.if.manually.managed=true
-{{- end -}}
-
 {{- define "confluence.sysprop.synchronyServiceUrl" -}}
 {{- if .Values.synchrony.enabled -}}
 -Dsynchrony.service.url={{ .Values.synchrony.ingressUrl }}/v1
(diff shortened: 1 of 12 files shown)


bordenit:
Relates to issue #85

src/main/charts/jira/templates/statefulset.yaml
```diff
@@ -73,6 +73,8 @@ spec:
                 configMapKeyRef:
                   key: reserved_code_cache
                   name: {{ include "jira.fullname" . }}-jvm-config
+            - name: ATL_TOMCAT_CONTEXTPATH
```
pbruski, line 76: Should we fix Confluence chart too?

pbruski (approved):

src/main/charts/jira/templates/statefulset.yaml
nanux: I have created https://github.com/atlassian-labs/data-center-helm-charts/pull/90 to provide support for Confluence. The Bitbucket is still missing it as it provides access via Spring boot properties but I believe the `healthinessProbe` will still need to be changed.

nanux:
@bordenit, thank you for your contribution. I believe if you run `mvn clean install` the build will fail as the helm output for Jira won't be as expected. I have created https://github.com/atlassian-labs/data-center-helm-charts/pull/90 that is passing and should give you an idea of what changes are required to make it passing. I have also created a test specifically for the context path value and it could be used by Jira as well. I can do that in my PR after we merge this.

Could you please fix the current tests before we are ready to merge?
Thank you

src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -149,6 +149,10 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 -Datlassian.logging.cloud.enabled={{.Values.fluentd.enabled}}
 {{- end }}
 
+{{- define "confluence.sysprop.enable.synchrony.by.default" -}}
```
nanux, line 152: Why is this as part of the PR?

nanux (changes requested):

kennymacleod (changes requested):
Remove irrelevant changes related to synchrony

bordenit:
@nanux  Thanks. I addressed the comments and it looks like the tests passed.

nanux:
> @nanux Thanks. I addressed the comments and it looks like the tests passed.

Thank you, could you please add it to the existing test? It is `@ParametrizedTest` so we can leverage set of products:

```
    @ParameterizedTest
    @EnumSource(value = Product.class, names = {"confluence", "jira"})
    void confluence_context_path(Product product) throws Exception {
        final var resources = helm.captureKubeResourcesFromHelmChart(product, Map.of(
                "confluence.service.contextPath", "/" + product.name()));

        resources.getStatefulSet(product.getHelmReleaseName())
                .getContainer()
                .getEnv()
                .assertHasValue("ATL_TOMCAT_CONTEXTPATH", "/" + product.name());


        assertEquals(resources.getStatefulSet(
                product.getHelmReleaseName()).getContainer().get("readinessProbe").get("httpGet").get("path").asText(),
                "/" + product.name() + "/status");
    }
```

bordenit:
@nanux I hope that worked, but if not, feel free to take it from here.

nanux (approved):
Thank you for the changes to the test and your contribution.

PR #94, opened 2021-03-31
DCKUBE-183 Make testing more flexible so parameters can be passed to …

wkritzinger-atlassian:
…setup.


src/test/scripts/nfs/etc/sv/nfs/run
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+modprobe -v nfs
```
nanux, line 3: What was the difference in NFS behaviour when we didn't load the `kmod` and nfs modules into kernel with `modprobe` before?

Are those modules loaded in the host system, i.e. the node running the pod?

src/test/scripts/nfs/nfs-server.yaml.template
```diff
@@ -51,12 +51,25 @@ spec:
...
-      args: ["-c", "apt -qq update && DEBIAN_FRONTEND=noninteractive apt install -qq -y nfs-kernel-server runit inotify-tools; mkdir -p /etc/sv/nfs ; runsvdir /etc/sv"]
+      args: ["-c", "apt -qq update && DEBIAN_FRONTEND=noninteractive apt install -qq -y nfs-kernel-server runit inotify-tools kmod; mkdir -p /etc/sv/nfs ; runsvdir /etc/sv"]
       volumeMounts:
         - mountPath: /srv/nfs
           name: nfs-export-dir
+        - mountPath: /lib/modules
+          name: lib-modules
+          readOnly: true
+      env:
+        - name: NFS_EXPORT_0
```
nanux, line 67: How is this used?

nanux (approved):

src/test/scripts/nfs/etc/sv/nfs/run
wkritzinger-atlassian: If another NFS container loaded these modules, or if they are loaded by default on the host system, this is not needed. In Docker for Mac, those modules are not loaded by default on the host VM, so we are loading them here. `modprobe` is a no-op if the module is already loaded, so adding this does not hurt in this case.

src/test/scripts/nfs/nfs-server.yaml.template
wkritzinger-atlassian: Oops, that was a leftover from another container I used, previously. I'll remove it.

PR #96, opened 2021-04-04
feat(range): Fix support for multiple plugins

sylus:
This should fix the error where if you have more then one plugin / library specified it will give the following error in helm:

```
Error: UPGRADE FAILED: YAML parse error on jira/templates/statefulset.yaml: error converting YAML to JSON: yaml: line 100:
```


sylus:
@pbruski would this be possible to merge? :D

pbruski (approved):
Thanks for fixing it!

sylus:
Woot thanks so much @pbruski  :D

PR #97, opened 2021-04-04
feat(perms): Paramterize setPermissions boolean flag

sylus:
I wanted to make the set_permissions a flag that can be set in values.yaml

Thanks :)


pbruski (approved):

src/main/charts/confluence/values.yaml
```diff
@@ -66,6 +66,8 @@ confluence:
   # -- The umask used by the Confluence process when it creates new files.
   # Default is 0022, which makes the new files read/writeable by the Confluence user, and readable by everyone else.
   umask: "0022"
+  # Boolean flag on whether to set permissions in the Confluence container.
```
nanux, line 69: `# -- Boolean to define whether to set home directory permissions on startup of Confluence container. Set to false to disable this behaviour.`

nanux (changes requested):

nanux (approved):

PR #98, opened 2021-04-06
DCKUBE-131: Alter NFS permissions via init container

### LINKED ISSUE #75: NFS-FIXER pod starts before PVC created
Hi,

I ran into something rather unusual when using `nfsPermissionFixer:`

```bash
helm install -n $n jira --values values.yaml data-center-helm-charts/jira --debug
```

I noticed that the `nfs-fixer` would time out waiting for the condition

```bash
install.go:172: [debug] Original chart version: ""
install.go:189: [debug] CHART PATH: /root/data-center-helm-charts/src/main/charts/jira
client.go:268: [debug] Starting delete for "jira-nfs-fixer" Job
client.go:297: [debug] jobs.batch "jira-nfs-fixer" not found
client.go:122: [debug] creating 1 resource(s)
client.go:477: [debug] Watching for changes to Job jira-nfs-fixer with timeout of 5m0s
client.go:505: [debug] Add/Modify event for jira-nfs-fixer: ADDED
client.go:544: [debug] jira-nfs-fixer: Jobs active: 0, jobs failed: 0, jobs succeeded: 0
client.go:505: [debug] Add/Modify event for jira-nfs-fixer: MODIFIED
client.go:544: [debug] jira-nfs-fixer: Jobs active: 1, jobs failed: 0, jobs succeeded: 0
Error: failed pre-install: timed out waiting for the condition
helm.go:81: [debug] failed pre-install: timed out waiting for the condition
```

After scratching my head until I was bald, I noticed I could alter the pre-install value in `nfs-permission-fixer.yaml`

Before:
```yaml
  annotations:
    "helm.sh/hook": pre-install
    "helm.sh/hook-delete-policy": "before-hook-creation,hook-succeeded"
```

After:
```yaml
  annotations:
    "helm.sh/hook": post-install
    "helm.sh/hook-delete-policy": "before-hook-creation,hook-succeeded"
```

This appears to have resolved one of my problems.

The second problem was the actual command run by `nfs-fixer`; if others are using an EFS backed PVC, then I found the following command in your `values.yaml` doesn't prevent Jira from starting:

```yaml
command: "chown 2001: /shared-home"
```

Hopefully, this may be of help to others encountering the same!

dylanrathbone:
Fixes #75


src/main/charts/bitbucket/templates/statefulset.yaml
```diff
@@ -32,6 +32,20 @@ spec:
       initContainers:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.volumes.sharedHome.nfsPermissionFixer.enabled }}
```
pbruski, line 35: While we're refactoring it, maybe we should call it sharedHome.initScript ?

src/main/charts/bitbucket/templates/statefulset.yaml
```diff
@@ -32,6 +32,20 @@ spec:
...
+        - name: nfs-permission-fixer
+          image: alpine
+          securityContext:
+            runAsUser: 0 # make sure we run as root so we get the ability to change the volume permissions
+          volumeMounts:
+            - name: shared-home
+              mountPath: {{ .Values.volumes.sharedHome.nfsPermissionFixer.mountPath | quote }}
+              {{- if .Values.volumes.sharedHome.subPath }}
+              subPath: {{ .Values.volumes.sharedHome.subPath | quote }}
+          {{- end }}
```
pbruski, line 46: misaligned if/end.

pbruski (approved):

src/main/charts/bitbucket/templates/statefulset.yaml
dylanrathbone: I'm not opposed to this, but I think the current naming is acceptable. It conveys intent but without exposing the details of how that intent is achieved which at this level I don't feel is important. Rewording things to include `initScript` is almost duplicating what we already know here. When looking at the surrounding YAML we see that this boolean is nested in an `initContainer` so I feel there is no need to double down on the name of this boolean.

bianchi2:
@dylanrathbone there can be scenarios when sts is invalid because of duplicated init containers like here https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/main/charts/bitbucket/templates/statefulset.yaml#L32, i.e. if I enable additional init containers (say, deliver plugins or anything else) and nfs fixer init container pod simultaneously.

Here's a simple test:

```
helm template bitbucket ./ --set volumes.sharedHome.nfsPermissionFixer.enabled=true
```
and in values.yaml I have:

```
additionalInitContainers:
  - name: add-prometheus-jar
    image: hello/hello
    imagePullPolicy: IfNotPresent
    command: [ "/bin/sh" ]
    args: [ "-c", "mkdir -p /opt/prometheus; cp /opt/jmx_prometheus_javaagent.jar /opt/jmx_prometheus_javaagent.jar" ]
    volumeMounts:
      - mountPath: /opt/prometheus
        name: prometheus
```
This results in:

```
      initContainers:
        - args:
          - -c
          - mkdir -p /home/jboss/prometheus; cp /opt/jmx_prometheus_javaagent.jar /home/jboss/prometheus/jmx_prometheus_javaagent.jar
          command:
          - /bin/sh
          image: hello/hello
          imagePullPolicy: IfNotPresent
          name: add-prometheus-jar
          volumeMounts:
          - mountPath: /home/jboss/prometheus
            name: prometheus
      initContainers:
        - name: nfs-permission-fixer
          image: alpine
          securityContext:
            runAsUser: 0 # make sure we run as root so we get the ability to change the volume permissions
          volumeMounts:
            - name: shared-home
              mountPath: "/shared-home"
          command: ["sh", "-c", "(chgrp 2003 /shared-home; chmod g+w /shared-home)"]
```

dylanrathbone:
😮  good point @bianchi2. This isnt going to work for scenarios like that. Let me look into this and see if there an approach to avoid creating duplicate `initContainers`

PR #104, opened 2021-04-13
DCKUBE-30: Reference NFS server example

nanux:
* Provides a reference example for customers to deploy NFS server
* Swaps the testing implementation


reference-infrastructure/storage/nfs/nfs-server.yaml
```diff
@@ -0,0 +1,102 @@
+---
```
pbruski, line 1: That looks neat, do we have stable (ish) green builds on Bitbucket with that setup?

reference-infrastructure/storage/nfs/nfs-server.yaml
```diff
@@ -0,0 +1,102 @@
...
+      role: nfs-server
+  template:
+    metadata:
+      labels:
+        role: nfs-server
+    spec:
+      containers:
+        - name: nfs-server
+          image: k8s.gcr.io/volume-nfs:0.8
+          imagePullPolicy: Always
```
dylanrathbone, line 67: should we change the pull policy to `IfNotPresent`?

src/main/charts/bitbucket/values.yaml
```diff
@@ -230,7 +230,7 @@ volumes:
     nfsPermissionFixer:
       # -- If enabled, this will alter the shared-home volume's root directory so that Bitbucket can write to it.
       # This is a workaround for a Kubernetes bug affecting NFS volumes: https://github.com/kubernetes/examples/issues/260
-      enabled: false
+      enabled: true
```
dylanrathbone, line 233: This should be `false`.

reference-infrastructure/storage/nfs/nfs-server.yaml
pbruski: Also, it looks like that image hasn't been touched for 3 years, which means it will be behind a lot of security updates (and fixes in general). The previous approach (although clearly more complicated) had the advantage of using a fresh, up to date image. I think we should either ship the previous solution, or, if we want something simpler, we should ship a remade image.

nanux: @pbruski you are correct. The `centos:latest` image contains 4 vulnerabilities (per `snyk scan`). It is doable to update them with
```
RUN yum update -y openssl-libs bind-export-libs gnutls
```
but I am not sure if we want to create a new Docker image. We were pondering this together with @dylanrathbone and we are leaning towards creating our own image based on a similar service approach you did in the `scripts/nfs` while using this new template. I don't think it will be still a production-ready approach but I am not sure if we can advise customers on a production-grade NFS server.

pbruski: That sounds good - yeah, I agree that producing an image as an official artifact would be problematic.

reference-infrastructure/README.md
```diff
@@ -0,0 +1,33 @@
...
+## Shared storage
+
+### Cloud managed shared storage
+
+#### Requirements
+
+#### Example
+
+
+### NFS server - Bitbucket DC requirement
```
dylanrathbone, line 19: Hi @tan-ro it would be great if you could give the wording a here a review.

tan-ro: @dylanrathbone Looks good to me.

nanux:
The builds are now passing and this is ready for review.
https://server-syd-bamboo.internal.atlassian.com/browse/DCNG-K8SBBS20

src/main/charts/bitbucket/values.yaml
```diff
@@ -235,7 +235,7 @@ volumes:
...
```
errcode1202, line 238: - [ ] this should be false.

errcode1202 (changes requested):

errcode1202 (approved):

nanux:
The long list of commits created an unnecessary noise in the commit list so I have squashed the commits.

- PR for reference

PR #107, opened 2021-04-20
DCKUBE-90: Additional details relating to the Ingress controller

dylanrathbone:
Changes rendered...

image


bianchi2 (approved):
LGTM. Though I find the diagram a bit confusing with 3 being above ingress controller. Maybe we can show that ing object is the provider of ing controller config. Not sure how to depict it :) 

Also, haven't you thought about bringing endpoints to the picture? svc > endpoint (which is pod host:port)?

dylanrathbone:
> Though I find the diagram a bit confusing with 3 being above ingress controller.
Now that you say it I agree. Let me tweak this.
> Maybe we can show that ing object is the provider of ing controller config. Not sure how to depict it :)
Yup, I wasn't too sure how best to depict this. the controller and objects themselves I do think are worth calling out as they are distinct objects. Let me play around with this some more and see what I can do :) 
> haven't you thought about bringing endpoints to the picture? svc > endpoint (which is pod host:port)?
It didn't occur to me to do this, I think partially because it might be getting a bit low level. Like the other comments above let me see how it fits when added to the diagram as a whole.

bianchi2:
@dylanrathbone I love the new diagram now :) As to endpoints, yeah, perhaps it's too many details, and it's not really necessary.

INSTALL.md
```diff
@@ -25,7 +25,22 @@ by the Helm charts will expect to be present.
    including a shared read-write volume. See the "Volumes" section below.
 * Ingress controller
    * Because different Kubernetes clusters use different Ingress configurations,
-   the Helm charts provide Nginx ingress templates only. By default, *ingress.create* is set to false, an ingress should be created manually after the chart is installed
+   the Helm charts provide Nginx ingress object templates only. By default, *ingress.create* is set to false, an ingress should be created manually after the chart is installed
```
nanux, line 28: It is not clear what will actually happen when people set `ingress.create` to `true`. We can provide some details for that workflow as it might be used in the testing scenario but I think emphasising that customers should create Ingress manually is a good approach.

nanux (approved):

docs/images/ingress.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
dylanrathbone: This is for draw.io. I could probably add a comment to indicate the fact

INSTALL.md
dylanrathbone: > @dylanrathbone I love the new diagram now :) As to endpoints, yeah, perhaps it's too many details, and it's not really necessary.

Yup, I want this to be more about the Ingress and not about the other interacting concepts too. I actually think this current diagram is a bit overboard too, but I'll leave as is for now.

docs/images/ingress.xml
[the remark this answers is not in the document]
dylanrathbone: Just added a comment on this to the top of the file ;)

INSTALL.md
dylanrathbone: > It is not clear what will actually happen when people set `ingress.create` to `true`. We can provide some details for that workflow as it might be used in the testing scenario but I think emphasising that customers should create Ingress manually is a good approach.

I agree, let me update the wording around this so that its more clear.

PR #108, opened 2021-04-21
Issue/dckube 204 add helm debug option to builds

badgersow:
This will provide `--debug` flag to `helm install`. If provisioning fails, the logs will indicate why exactly it has failed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,7 @@
 
         <extra.parameters />
 
-        <helm.debug>false</helm.debug>
+        <helm.debug>true</helm.debug>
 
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -122,6 +122,8 @@
 
         <extra.parameters />
 
+        <helm.debug>false</helm.debug>
```
nanux, line 125: I am leaning towards making this `true` as otherwise `helm` is very quiet and it will be potentially also useful output in case customers are going to reach out to Support.

badgersow: I agree, this makes sense. I've changed the default.

nanux (approved):

PR #110, opened 2021-04-22
DCKUBE-231: Max body size should be configurable

dylanrathbone:


wkritzinger-atlassian (approved):

nanux (approved):

src/main/charts/jira/values.yaml
```diff
@@ -140,11 +140,14 @@ ingress:
...
   # https://kubernetes.github.io/ingress-nginx/
-  
   # This will populate the Ingress Resource with annotations for the Kubernetes ingress-nginx controller.
   # Set to false if a different controller is to be used, in which case the appropriate annotations for that
   # controller need to be specified.
   nginx: true
+  # -- The max body size to allow. Requests exceeding this size will result
+  # in an 413 error being returned to the client.
+  # https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size
+  maxBodySize: 10m
```
nanux, line 150: Should we mention that this size should be at least the same as the configured maximum attachment size (which is 10m by default)?

https://confluence.atlassian.com/adminjiraserver/configuring-file-attachments-938847851.html

dylanrathbone: Im inclined to leave it, Confluence specifies a different default of 100m:

https://confluence.atlassian.com/doc/configuring-attachment-size-150040.html

nanux: Ok, I think it is something that should be mentioned in our CAC documentation.

PR #111, opened 2021-04-23
DCKUBE-205: fix the provisioning error when debug is disabled

badgersow:
The following error was present when debug option was set to false (`-Dhelm.debug=false`)

```
+ helm install -n dcd --wait epyshnograev-confluence-pgsql --values ./src/test/scripts/postgres-values.yaml --set fullnameOverride=epyshnograev-confluence-pgsql --set image.tag=11 --set postgresqlDatabase=confluence --set postgresqlUsername=confluence --set postgresqlPassword=confluence --version 9.4.1 '' bitnami/postgresql
Error: expected at most two arguments, unexpected arguments: bitnami/postgresql
```

The issue was caused by quotes around potentially empty variable.


nanux (approved):
Thanks for fixing this.

BTW I think this would be captured by https://www.shellcheck.net/ - available as IDEA plugin https://plugins.jetbrains.com/plugin/10195-shellcheck

PR #112, opened 2021-04-26
Dckube 208 efk documentation

nghazali:
Added documentation for EFK stack for kubernetes cluster


src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+`Elasticsearch` is a distributed open search and analytics engine for all types of data. 
+
+`Kibana` is an open source frontend application that sits on top of Elasticsearch, providing search and data visualization capabilities for data indexed in Elasticsearch.
+
+There are different methods to deploy EFK stack and here we provide two deployment methods, first deploy EFK using helm chart and deploy local elasticsearch on Kubernetes and second using managed elasticsearch (EKS and AWS ES).
+
+## EFK using local elasticsearch
+
+This solution deploys all EFK stack inside the k8s cluster locally. By enabling fluentd, helm_install script installs fluentd in EKS. It means after running helm_install all product pods will run fluentd which collects all log files and sends them to the fluentd aggregator container.
```
nanux, line 15: We shouldn't mention `helm_install` scripts.

nanux, line 15: I was looking through the charts and I cannot see any centralised logging fluentd pod that would collect the logs. I think every fluentd sidecar is sending the logs directly to elasticsearch.

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+There are different methods to deploy EFK stack and here we provide two deployment methods, first deploy EFK using helm chart and deploy local elasticsearch on Kubernetes and second using managed elasticsearch (EKS and AWS ES).
+
+## EFK using local elasticsearch
+
+This solution deploys all EFK stack inside the k8s cluster locally. By enabling fluentd, helm_install script installs fluentd in EKS. It means after running helm_install all product pods will run fluentd which collects all log files and sends them to the fluentd aggregator container. 
+
+To complete the EFK stack we need to install elasticsearch cluster and kibana and manage to forward the aggregated datalog to elasticsearch by fluentd which is already installed. 
+
+To start first enable `fluentd` and set the `hostname` for elasticsearch in volues.yaml as follows:
```
nanux, line 19: `values.yaml`

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+```yaml
+fluentd:
+   enabled: true
+   elasticsearch:
+     hostname: elasticsearch-master
+```
+
+Now fluentd tries to parse and send the data to elasticsearch but as it is not installed yet then data get lost. At this point we should install elasticsearch and kibana to complete EFK stack deployment.
```
nanux, line 28: Shouldn't we start with the installation of the elasticsearch first? It doesn't have any dependency on the helm charts so I would assume installing it first is a better deal.

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+```shell script
+$ kubectl get pods --namespace=dcd -l app=elasticsearch-master
+NAME                     READY   STATUS    RESTARTS   AGE
+elasticsearch-master-0   1/1     Running   0          100m
+elasticsearch-master-1   1/1     Running   0          100m
+elasticsearch-master-2   1/1     Running   0          100m
+$
+$ kubectl port-forward svc/elasticsearch-master 9200
+```
+make sure elasticsearch cluster working as expected:
```
nanux, line 49: `is working`

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+```shell script
+$ kubectl get deployment
+NAME                               READY   UP-TO-DATE   AVAILABLE   AGE
+helm-operator                      1/1     1            1           23d
+ingress-nginx-release-controller   1/1     1            1           82d
+kibana-kibana                      1/1     1            1           125m
+
+$ kubectl port-forward deployment/kibana-kibana 5601
+```
+Now kibana is accessible in the browser by: http://localhost:5601. To visualise the logs you need to create an index pattern and then see the data in the discovery part. To create the index pattern go to `Management` → `Stack Management` and then select `Kibana` → `Index Patterns`.
```
nanux, line 88: `browser on url: http://...`

nanux, line 88: Regarding the **Index Petterns** - is that something that is serialisable, can we provide it as a configuration?

Basically, the question is, how much we can provide customers product-specific patterns that will help them build an accurate index. Looking through https://hello.atlassian.net/wiki/spaces/DC/blog/2021/01/23/997257266/Log+aggregation+for+DC+products+in+Kubernetes+-+Inno+Week+of+January+2021 I can see that @kennymacleod enabled fluentd logger. Is that the way how Jira and Confluence works out of the box?

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+ingress-nginx-release-controller   1/1     1            1           82d
+kibana-kibana                      1/1     1            1           125m
+
+$ kubectl port-forward deployment/kibana-kibana 5601
+```
+Now kibana is accessible in the browser by: http://localhost:5601. To visualise the logs you need to create an index pattern and then see the data in the discovery part. To create the index pattern go to `Management` → `Stack Management` and then select `Kibana` → `Index Patterns`. 
+
+## EFK in EKS using managed AWS Elasticsearch
+
+In this method elastic search will deploy out of k8s clusters. For this purpose we use fluent bit instead of fluentd which is used for locally deployment for EFK.
```
nanux, line 92: `elasticsearch` will deploy as a managed service and it is outside of the kubernetes cluster.

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+$ kubectl port-forward deployment/kibana-kibana 5601
+```
+Now kibana is accessible in the browser by: http://localhost:5601. To visualise the logs you need to create an index pattern and then see the data in the discovery part. To create the index pattern go to `Management` → `Stack Management` and then select `Kibana` → `Index Patterns`. 
+
+## EFK in EKS using managed AWS Elasticsearch
+
+In this method elastic search will deploy out of k8s clusters. For this purpose we use fluent bit instead of fluentd which is used for locally deployment for EFK. 
+
+When a node inside an EKS cluster needs to call AWS APIs, it needs to provide extended permissions. Amazon provided an image of fluent-bit which supports service accounts and using this we no longer need to follow the traditional way. All we need is to have an IAM role for the service account on EKS cluster so using this service account we can provide AWS permission to the containers in any pod that use that service account. So the pods on that node can call AWS APIs.
```
nanux, line 94: `AWS service accounts`  (to avoid confusion with kubernetes roles.

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+`fluentbit` is used to collect and aggregate the data inside the EKS cluster which will communicate with AWS elasticsearch outside of the cluster. 
+
+First step is configure IRSA for `fluentbit` to make sure we have OIDC identity provider to use IAM roles for the service account in the cluster:
```
nanux, line 98: `IAM Roles for Service Accounts (IRSA)`

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+   -H 'Content-Type: application/json' \
+   -d'
+[
+   {
+     "op": "add", "path": "/backend_roles", "value": ["'${FLUENTBIT_ROLE}'"]
+   }
+]
+'
+```
+Finally, it is time to deploy fluent bit cluster:
```
nanux, line 234: `deamonset`

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
+## Deploy Logging Pattern - EFK Stack
```
nanux, line 1: I would put on the top a disclaimer that this is not officially supported document and it is meant as a guidance.

```
## Disclaimer

**This is not officially supported functionality.** This document serves as an example for users that would like to enable aggregated logging in their Kubernetes cluster. There are many solutions how to provide this capability and this document just showcases some of the options.
```
or something in similar fashion.

nanux (changes requested):
I would summon @tan-ro to help us with English and approachability for users.

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+A common Kubernetes logging pattern is a combination of Fluent, Elasticsearch, and Kibana which is known as *EFK Stack*.
```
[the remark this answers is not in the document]
nghazali: Fluent in general, we use FluentD in one solution and Fluent Bit in second solution for managed elasticsearch

src/main/logging/README.md
```diff
@@ -0,0 +1,244 @@
+## Deploy Logging Pattern - EFK Stack
+
+A common Kubernetes logging pattern is a combination of Fluent, Elasticsearch, and Kibana which is known as *EFK Stack*. 
+
+`Fluent` is an open source and multi-platform log processor which collects data/log from different sources, aggregates and forwards them to multiple destinations and is fully compatible with Docker and Kubernetes environments.
```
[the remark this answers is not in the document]
nghazali: As I mentioned above, at this point I prefer to use Fluent as is more generic word

src/main/logging/README.md
nghazali: I agree, I moved the order

src/main/logging/README.md
nghazali: Agreed, changed the order

src/main/logging/local_es/elasticsearch.yaml
```diff
@@ -0,0 +1,261 @@
+---
```
nanux, line 1: Is it necessary to define all the parameters? Can we rely on defaults and just override the ones that are required?

nanux (approved):
I am approving the PR but would be good to have @tan-ro review as well.

src/main/logging/local_es/elasticsearch.yaml
nghazali: we can do not provide any value file at al1, because all the changes are related to the resource management and there is no specific change which rely to specific deployment itself. 
However, having this complete file gives the better sight to client about their options specially when we are not officially support.

nanux:
Thanks @tan-ro for the review!

PR #122, opened 2021-05-06
DCKUBE-51 Application upgrade

nghazali:
Both Kubernetes upgrade strategies for statefulset (RollingUpdate and OnDelete) is tested successfully on Bitbucket and Confluence helm charts. 
RollingUpgrade with 0 partition is added to the statefulset spec as a guideline. 
Upgrade steps are documented
An environment variable (KUBE_POD_NAME) is introduced to Bitbucket container which is a unique identifier for pod in the cluster. This can be used for pod persistent identifier in ZDU to detect returning nodes. 
Cleaned up docs folder


INSTALL.md
```diff
@@ -133,6 +133,10 @@ products to work correctly in a multi-node configuration). Here is the syntax fo
 kubectl scale statefulsets <statefulsetset-name> --replicas=n
```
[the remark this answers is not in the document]
nghazali: I tested helm upgrade for scale up and then upgraded to the new version but in this case I had some problem with confluence.

docs/bitbucket-upgrade.md
```diff
@@ -0,0 +1,38 @@
...
+    $ helm upgrade -n dcd --wait <release name> --set replicaCount=3 target/helm/bitbucket-0.1.1.tgz --reuse-values
+    ```
+
+1. Upgrade will start by terminating one pod and creating a new pod with an updated version. 
+    ![upgrade-mode](./images/bitbucket-upgrade-2.png)
+
+1. After the new pod is up and running, the next pod will be upgraded until all pods are upgraded to the new version. 
+
+1. After all pods are activated with the new version, finalize the upgrade.
+    ![upgrade-mode](./images/bitbucket-upgrade-3.png)
```
errcode1202, line 29: Should you bullet points increment? They are all `1.`

errcode1202 (changes requested):

docs/bitbucket-upgrade.md
nghazali: The numbers will render correctly in output and the sequence will be in order.

errcode1202 (approved):

src/test/resources/expected_helm_output/bitbucket/output.yaml
```diff
@@ -124,6 +128,10 @@ spec:
             - name: shared-home
               mountPath: "/var/atlassian/application-data/shared-home"
           env:
+            - name: KUBE_POD_NAME
```
nanux, line 131: How this is utilised?

nanux (approved):

src/test/resources/expected_helm_output/bitbucket/output.yaml
nghazali: This will define an environment variable named `KUBE_POD_NAME` in the container. This is done here to address the persistent identifier issue in ZDU. 
There is still a work to do that is  adding a jvm parameter (`-Dcluster.node.name=KUBE_POD_NAME`) in bitbucket startup script and involve in in node persistent identifier logic in ZDU. 
Hazelcast will assign pod name which is unique in each restart to node name in clusterInfo. Then what we have to do is involving node name into detect returning nodes in ZDU to handle the offline nodes display issue after a node is upgraded and is running. We have a ticket already to address this.

PR #125, opened 2021-05-10
Issue 712/dckube 205 make possible to run k8s tests in local cluster

badgersow:
In this pull request we make possible to provision Confluence on local K8s cluster


src/main/charts/confluence/templates/ingress-setup.yaml
```diff
@@ -0,0 +1,37 @@
...
+  name: {{ include "confluence.fullname" . }}-setup
+  labels:
+    {{- include "confluence.labels" . | nindent 4 }}
+  annotations:
+    {{ if .Values.ingress.nginx }}
+    "kubernetes.io/ingress.class": "nginx"
+    "nginx.ingress.kubernetes.io/affinity": "cookie"
+    "nginx.ingress.kubernetes.io/affinity-mode": "persistent"
+    "nginx.ingress.kubernetes.io/proxy-body-size": {{ .Values.ingress.maxBodySize }}
+    "nginx.ingress.kubernetes.io/proxy-connect-timeout": "300"
```
badgersow, line 15: This is the separate Ingress with increasing timeout. It's existing for `/setup` endpoint for Confluence, because those HTTP requests are expected to be slow

nanux: I am hesitant to introduce another `Ingress` just for the purpose of `setup` slow endpoints. This will stay there even after the setup was finished and it might be just confusing for customers. But tbh I didn't find a better workaround for now. Clearly the best solution is to provide fully automated setup in Confluence but that is a long term solution. I was looking into `Jobs` and Helm `hooks`. None will solve our use case to manipulate the values of a resources on certain conditions that will happen after manual intervention. Operator could actually do that but again that is a bigger project.

For now, we can probably at least stick with DRY. We can create `_ingress-template.yaml` file that will contain the current content with the flexibility to change the name and the timeout. That will drastically simplify the two ingresses (because they are mostly duplicated). More information about library charts is here: https://helm.sh/docs/topics/library_charts/

src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -149,6 +149,16 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
 {{- end }}
 
+{{- define "confluence.sysprop.debug" -}}
+{{- if .Values.confluence.jvmDebug.enabled -}}
+-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005
+{{- end }}
+{{- end }}
+
+{{- define "confluence.sysprop.enable.synchrony.by.default" -}}
+-Dsynchrony.by.default.enable.collab.editing.if.manually.managed=true
```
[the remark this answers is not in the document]
badgersow: This only works after Confluence 7.12. Do we mention the minimal supported Connie version somewhere?

src/main/charts/confluence/values.yaml
```diff
@@ -331,3 +335,12 @@ additionalFiles: []
 #    type: secret
 #    key: keystore.jks
 #    mountPath: /var/ssl
+
+# -- Additional host aliases for each pod, equivalent to adding them to the /etc/hosts file. See
+# https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/ for more
+# information.
+additionalHosts: []
```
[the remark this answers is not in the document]
badgersow: This is needed for Confluence to be able to contact Synchrony. In the local execution the ingress lives in `.local` Bonjour address. When Confluence pod needs to contact synchrony, it's going to do that using the Ingress address (`.local` hostname). Hence, we must be sure that Ingress address will not be resolved to loopback (127.0.0.1) and will instead be resolved to the IP address of the ingress Service. Otherwise Connie pod will try to send HTTP calls to itself when it needs to contact synchrony and it will fail. 
This `additionalHosts` annotation does exactly what we need - maven determines the LoadBalancer address of the work laptop and passes it to Kube. Hence, pods know how to resolve the Ingress hostname correctly.
This whole situation seems to be a problem only when ingress lives on local address.

src/test/scripts/helm_install.sh
```diff
@@ -33,6 +33,17 @@ get_current_cluster_type() {
   esac
 }
 
+startNfsServer() {
```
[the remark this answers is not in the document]
badgersow: This is indeed a letfover. Thank you for noticing this, I've removed it and tested that everything is fine.

src/main/charts/confluence/README.md
```diff
@@ -23,13 +23,15 @@ Kubernetes: `>=1.17.x-0`
...
 | confluence.accessLog.mountPath | string | `"/opt/atlassian/confluence/logs"` | The path within the Confluence container where the local-home volume should be mounted in order to capture access logs. |
 | confluence.additionalBundledPlugins | list | `[]` | Specifies a list of additional Confluence plugins that should be added to the Confluence container. These are specified in the same manner as the additionalLibraries field, but the files will be loaded as bundled plugins rather than as libraries. |
 | confluence.additionalEnvironmentVariables | list | `[]` | Defines any additional environment variables to be passed to the Confluence container. See https://hub.docker.com/r/atlassian/confluence-server for supported variables. |
 | confluence.additionalJvmArgs | list | `[]` | Specifies a list of additional arguments that can be passed to the Confluence JVM, e.g. system properties |
+| confluence.jvmDebug.enabled | bool | `false` | If set to `true`, Confluence JVM will be started in debug mode (on port 5005). Kuberneter port forwarding feature could be used to attach a debugger to the JVM. |
```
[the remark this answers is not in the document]
badgersow: I think it will and I agree that we should make it in subsequent PR. I've created DCKUBE-283 to track this.

errcode1202, line 34: Typo here `Kuberneter` should be `Kubernetes`

errcode1202 (changes requested):

src/main/charts/confluence/values.yaml
```diff
@@ -153,6 +153,10 @@ confluence:
   # See https://hub.docker.com/r/atlassian/confluence-server for supported variables.
   additionalEnvironmentVariables: []
 
+  jvmDebug:
+    # -- If set to true, Confluence JVM will be started with debugging port open.
```
errcode1202, line 157: For clarity can we include the port number (5005) in this comment

src/main/charts/confluence/templates/ingress-setup.yaml
badgersow: I've extracted the ingress to the library `_ingress.tpl` and trimmed `ingress.yaml` and `ingress-setup.yaml` to define only parts which are differ.
The repetition is gone. I can't say I like it very much though, because using libraries in Helm is pretty ugly. You have to use a common chart utility to merge 2 trees together, and this adds complexity.
One more thing - because this common merge utility is used, Confluence chart right now has a dependency on a common library. I had to add some code to tests and `helm_install.sh` script which updates the dependency prior to packaging (it will now work otherwire).
All in all, I think it should be OK in the current state. Unit tests are okay and I was also able to provision and test Confluence. Let's see how it goes.

src/main/charts/confluence/README.md
badgersow: Thanks, fixed

src/main/charts/confluence/values.yaml
badgersow: Added

errcode1202 (approved):

.github/workflows/helm_release.yaml
```diff
@@ -1,10 +1,8 @@
+# This workflow releases new version of helm charts
 name: Release Charts
 
 on:
   workflow_dispatch:
-  push:
```
nanux, line 5: Thanks, it was also manually disabled but it is better to remove this.

Btw by default it shouldn't do a release it should verify if there is a new `version` in `chart/Chart.yaml` file and release only a new version.

.github/workflows/maven.yml
```diff
@@ -5,6 +5,9 @@ name: Maven unit tests
 
 on:
   push:
+    branches:
+      - master
+      - /.*-stable/
```
nanux, line 10: I saw this in `prometheus-operator` Github actions. We don't use the `*-stable` pattern, is that something you want to introduce?

src/main/charts/confluence/templates/_helpers.tpl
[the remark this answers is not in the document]
nanux: Can you add it to https://hello.atlassian.net/wiki/spaces/DCD/pages/1108928180/Kubernetes+-+Questions+and+answers

> Which product versions are supported with Kubernetes?

src/main/charts/confluence/templates/ingress-setup.yaml
nanux: I am really sorry for flip-flopping on this one. I think the original intent was to keep it simple and seems that `helm` is not really providing us with the tools to DRY. Using the library is probably adding the same amount of code as removes but also increases complexity and dependency on another library, which is deprecated.

From my side, I would then rather prefer to stick with two very similar ingresses than trying to refactor `yaml` using an external library. Hopefully, this shouldn't be a big change as we have the files already in the git history and we can just fall back to the previous solution.

src/main/charts/confluence/values.yaml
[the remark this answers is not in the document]
nanux: Is that something that will be useful to have in documentation?

src/main/charts/confluence/templates/ingress-setup.yaml
badgersow: I've reverted those commits and now we are back to the old solution

src/main/charts/confluence/templates/_helpers.tpl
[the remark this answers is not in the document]
badgersow: There is already a row for that: `Which product versions are supported with Kubernetes?`

src/main/charts/confluence/values.yaml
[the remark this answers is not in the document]
badgersow: It's already in the [internal documentation](https://hello.atlassian.net/wiki/spaces/DCD/pages/1112917002/How%2Bto%2Binstall%2BConfluence%2Bon%2Blocal%2BKubernetes%2Bcluster#Provision-the-cluster) for the local execution. In general, I don't think it's useful in the external docs, because we don't give people configuration for localhost. If we add this support, then we definitely will look at the internal docs when documenting it for customers.

src/main/charts/confluence/templates/_helpers.tpl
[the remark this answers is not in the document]
nanux: I've added Confluence 7.12 to the row.

nanux:
Thanks, Efim!

nanux (approved):

PR #126, opened 2021-05-10
Lower the NFS server requirements and remove storageClass

nanux:
I decided to move away from the reference infrastructure. I think we are mixing two concerns together - throwaway infrastructure for testing and an example for customers. I think this is keeping it cleaner for customers.

It also separates our requirements:
* need to be able to run in the local environment (use default storage class)
* lower the computational resources (so the builds are not draining our CI capacity)

I haven't made a decision whether to remove the `test-` prefix for the resources in https://github.com/atlassian-labs/data-center-helm-charts/blob/master/reference-infrastructure/storage/nfs/nfs-server.yaml
To some degree, it is still NFS Server for testing purposes. 

Just to learn `kustomize` I've tried to see how this change would look like in it - https://github.com/atlassian-labs/data-center-helm-charts/pull/124
I must say I much prefer the simplicity of this solution. I think I would rather have either very simple yaml files with `sed` or proper full helm chart. Kustomize is somewhere in the middle and I find it clunky.

### DIFF
--- a/src/test/infrastructure/storage/test-nfs-server.yaml
+++ b/src/test/infrastructure/storage/test-nfs-server.yaml
@@ -1,0 +1,135 @@
+### TEST TEMPLATE ONLY (works on AWS only)
+###
+### This template serves as an example how to deploy a very simple NFS server providing persistent volumes
+### in read write many mode (as all the application nodes require to access the shared home from one persistent volume).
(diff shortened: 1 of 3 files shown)


badgersow:
I think we should merge this now to unblock the local execution and maybe experiment later to make it nicer

PR #134, opened 2021-05-17
Allow for the ingress path's to be configurable.

smarch730:
Added ingress.host into values.yaml with corresponding value injection in ingress.yaml for all apps. Defaults to "/"


badgersow:
We should also add this to the Confluence setup ingress [here](https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/main/charts/confluence/templates/ingress-setup.yaml)

nanux (approved):

nanux:
I have created https://github.com/atlassian-labs/data-center-helm-charts/pull/136 to fix missing path in `/setup`.

PR #135, opened 2021-05-19
DCKUBE-267: Resource request definitions

errcode1202:
To facilitate the accurate scheduling of pods to a k8s cluster define the default `cpu` and `memory` request requirements for each product.


src/main/charts/bitbucket/values.yaml
```diff
@@ -104,27 +104,32 @@ bitbucket:
...
+      minHeap: "512m"
 
     # -- Specifies the standard Kubernetes resource requests and/or limits for the Bitbucket container.
     # It is important that if the memory resources are specified here, they must allow for the size of the Bitbucket JVM.
     # That means the maximum heap size, the reserved code cache size, plus other JVM overheads, must be accommodated.
-    # Allowing for (maxHeap+codeCache)*1.5 would be an example.
-    container: {}
+    # Allowing for maxHeap * 1.5 would be an example.
+    container:
     #  limits:
```
nanux, line 120: Should we add some comment for the limits as well?

src/main/charts/bitbucket/values.yaml
```diff
@@ -104,27 +104,32 @@ bitbucket:
...
+      requests:
+        # -- If changing the cpu value update additional JVM arg 'ActiveProcessorCount' below
+        cpu: "1"
+        memory: "2G"
 
   # -- Specifies a list of additional arguments that can be passed to the Bitbucket JVM, e.g. system properties
   additionalJvmArgs:
-  #    - -Dfoo=bar
-  #    - -Dfruit=lemon
+    # -- The value defined for ActiveProcessorCount should correspond to that provided for 'container.requests.cpu'
```
nanux, line 130: @tan-ro this is something we should mention in proper documentation as well.

src/main/charts/confluence/values.yaml
```diff
@@ -105,28 +105,33 @@ confluence:
 
   resources:
     jvm:
+      # -- JVM memory arguments below are based on the defaults defined for the Confluence docker container, see:
+      # https://bitbucket.org/atlassian-docker/docker-atlassian-confluence-server/src/master/#markdown-header-memory-heap-size
```
nanux, line 109: This doesn't work for me - it doesn't scroll.

I believe we would need to use https://bitbucket.org/atlassian-docker/docker-atlassian-confluence-server/src/master/README.md#markdown-header-memory-heap-size

src/main/charts/bitbucket/values.yaml
```diff
@@ -104,27 +104,32 @@ bitbucket:
...
     #  limits:
-    #    cpu: "4"
-    #    memory: "2G"
-    #  requests:
-    #    cpu: "4"
+    #    cpu: "1"
     #    memory: "2G"
+      requests:
+        # -- If changing the cpu value update additional JVM arg 'ActiveProcessorCount' below
+        cpu: "1"
```
nanux, line 125: Where is the default 1 CPU coming from? I believe in the docker environment the containers have access to all cores available to the Docker engine - e.g. the number of cores enabled in the Resources tab in the Docker for Mac settings.

I think restricting to exactly 1 CPU is not necessary - can we put 2 at least here?

src/main/charts/confluence/values.yaml
errcode1202: Hmm works for me. Let me look...

src/main/charts/bitbucket/values.yaml
errcode1202: We decided to settle on 1 cpu based on the testing we did with the updated memory requests. 1 cpu actually performed well, granted this is based on basic testing with little load. I have no problem upping to 2

src/main/charts/bitbucket/values.yaml
errcode1202: I dont think its necessary we dont have one for `requests` and there will be detail for these 2 in the accompanying documentation: https://github.com/atlassian-labs/data-center-helm-charts/pull/137/files#diff-e0b8466f4f553e9957720b26bb44cad7e526abbd0950363ea570f76cd0517fff

src/main/charts/bitbucket/values.yaml
errcode1202: Good point!

src/main/charts/confluence/values.yaml
nghazali: Honestly, none of those links work for me - those don't scroll to the point, both just open the page.

src/main/charts/bitbucket/values.yaml
nghazali: I confirm when we set the cpu request to 1 and don't define the cpu limit, pod have access to all core available in node just like Docker. 
So, we may are safe to leave it as it is also I have no objection to increase it to 2 cores. In this case we may good to increase the memory request and max heap size accordingly by another gigabyte to cover a bit more than minimum dataset for products.

nanux (approved):

src/main/charts/bitbucket/values.yaml
```diff
@@ -104,27 +104,32 @@ bitbucket:
...
+        cpu: "2"
+        memory: "2G"
 
   # -- Specifies a list of additional arguments that can be passed to the Bitbucket JVM, e.g. system properties
   additionalJvmArgs:
-  #    - -Dfoo=bar
-  #    - -Dfruit=lemon
+    # -- The value defined for ActiveProcessorCount should correspond to that provided for 'container.requests.cpu'
+    # see: https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-3B1CE181-CD30-4178-9602-230B800D4FAE
+    - -XX:ActiveProcessorCount=1
```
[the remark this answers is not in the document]
errcode1202: Done

src/main/charts/confluence/values.yaml
errcode1202: Yes auto scrolling to the actual section doesnt work for either link. I think thats an issue on BB's side.

nghazali (approved):

PR #140, opened 2021-05-21
initial commit - changed the max-body-size of request to 250MB for al…

nghazali:
Changed max bosy size of request for all products to 250M and documented it in CONFIG.md

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -160,6 +160,11 @@ If you're not using the `shared-home` volume, then you can declare your own cust
 volume in the "Additional Volumes" section below, then declare the libraries as above
 (but with your custom volume name).
 
+## Request body size
+By default the maximum allowed size for the request body is set to 250MB. If the size in a request exceeds the maximum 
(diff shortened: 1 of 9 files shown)


errcode1202 (approved):
🌮

CONFIG.md
```diff
@@ -160,6 +160,11 @@ If you're not using the `shared-home` volume, then you can declare your own cust
 volume in the "Additional Volumes" section below, then declare the libraries as above
 (but with your custom volume name).
 
+## Request body size
```
nanux, line 163: I think it is worth mentioning that this is specific to NGINX ingress, for example with:

`## Request body size (NGINX ingress only)`

nanux (approved):

CONFIG.md
errcode1202: And its actually called `ingress-nginx` - we should include the URL to this annotation (`proxy-body-size`) https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size

PR #142, opened 2021-05-21
DCKUBE-297: Add rolling upgrade docs for Jira

apawelczyk-atlassian:


docs/confluence-upgrade.md
```diff
@@ -1,8 +1,8 @@
 ## Confluence Rolling Upgrade
-Suppose we have a Confluence cluster with 3 nodes, all running `Confluence xx.xx.0-jdk11`, and an upgrade to 
-`Confluence xx.xx.1-jdk11` is planned. Use these steps to complete the upgrade process: 
+Suppose we have a Confluence cluster with 3 nodes, all running `Confluence xx.xx.0-jdk11`, and an upgrade to
+`Confluence xx.xx.1-jdk11` is planned. Use these steps to complete the upgrade process:
 
-1. Update the `version` and `appVersion` in the Helm chart (`src/main/charts/bitbucket/Chart.yaml`):
+1. Update the `version` and `appVersion` in the Helm chart (`src/main/charts/confluence/Chart.yaml`):
```
nanux, line 5: good catch

docs/jira-upgrade.md
```diff
@@ -0,0 +1,34 @@
...
+
+1. Update the `version` and `appVersion` in the Helm chart (`src/main/charts/jira/Chart.yaml`):
+ ```yaml
+ version: 0.1.1
+ appVersion: xx.xx.1-jdk11
+ ```
+
+2. Create a new version of the Helm package:
+ ```shell script
+ $ helm package src/main/charts/jira --destination target/helm
```
nanux, line 13: As discussed IRL we should not encourage people to package the chart from the source files. Instead, they can set the `image.tag` if they want to change the app version.

Ideally there would be a parameter for `helm upgrade` command but it is currently just an open issue - https://github.com/helm/helm/issues/8194

nanux (dismissed):

docs/confluence-upgrade.md
errcode1202, line 5: should we be referencing local directories here? This is actually a general question as I see we reference local paths throughout our readme's. The reason I ask is based on the assumption that we're using the helm remote repo for the charts and not a clone of this repo, see our instructions [here](https://github.com/atlassian-labs/data-center-helm-charts/blob/master/INSTALL.md#installation). With that in mind how would the path(s) change, Im actually not sure 🤔

docs/confluence-upgrade.md
```diff
@@ -22,12 +22,12 @@ Suppose we have a Confluence cluster with 3 nodes, all running `Confluence xx.xx
  $ helm upgrade -n dcd --wait <release name> --set replicaCount=3 target/helm/confluence-0.1.1.tgz --reuse-values
```
errcode1202, line 22: `-n dcd` should be `-n <namespace>`

docs/jira-upgrade.md
```diff
@@ -0,0 +1,34 @@
...
+ This will create target/helm/jira-0.1.1.tgz
+ ```
+
+3. Go to **Administration > Applications > Jira upgrades** and click **Put Jira into upgrade mode**.
+
+  ![upgrade-mode](./images/jira-upgrade-1.png)
+
+4. Run helm upgrade command with the desired number of nodes after upgrade (replicaCount):
+ ```shell script
+ $ helm upgrade -n dcd --wait <release name> --set replicaCount=3 target/helm/jira-0.1.1.tgz --reuse-values
```
errcode1202, line 23: See comment: https://github.com/atlassian-labs/data-center-helm-charts/pull/142/files#r638356911

docs/confluence-upgrade.md
```diff
@@ -1,8 +1,8 @@
 ## Confluence Rolling Upgrade
-Suppose we have a Confluence cluster with 3 nodes, all running `Confluence xx.xx.0-jdk11`, and an upgrade to 
-`Confluence xx.xx.1-jdk11` is planned. Use these steps to complete the upgrade process: 
+Suppose we have a Confluence cluster with 3 nodes, all running `Confluence xx.xx.0-jdk11`, and an upgrade to
```
errcode1202, line 2: Are these instructions ZDU specific? If thats the case should we call out that ZDU can only be performed on patch versions (at least for Confluence) ?

errcode1202 (approved):

docs/confluence-upgrade.md
apawelczyk-atlassian: We haven't defined who these docs should target. Should it be the users of the charts or people building their own derivative version of these charts? I assumed the latter since they exist here, in the repo, and a next step talks about calling `helm package`.

docs/confluence-upgrade.md
apawelczyk-atlassian: The scope of this PR was Jira so I won't be able to modify Confluence docs further. I just wanted to fix the leftover from copying the file from Bitbucket's docs (and some whitespace). Should I revert these changes to avoid confusion?

apawelczyk-atlassian:
There's some confusion with the scope of this pull request. I'm going to revert to its original - only docs for Jira, based on the existing ones - and open new one where I'll apply your suggestions to all the products.

PR #144, opened 2021-05-24
DCKUBE-317: additional libraries for Synchrony

yzha645:


src/main/charts/confluence/values.yaml
```diff
@@ -188,6 +188,14 @@ synchrony:
   # Confluence service will use to communicate directly with Synchrony, so the URL must be resovable both from inside and
   # outside the Kubernetes cluster.
   ingressUrl:
+  # -- Specifies a list of additional Java libraries that should be added to the Confluence container.
```
badgersow, line 191: Replace `Confluence container` with `Synchrony container`

badgersow (approved):

src/main/charts/confluence/templates/statefulset-synchrony.yaml
```diff
@@ -74,4 +75,5 @@ spec:
           configMap:
             name: {{ include "synchrony.fullname" . }}-entrypoint
             defaultMode: 0744
+        {{ include "confluence.volumes.sharedHome" . | nindent 8 }}
```
nanux, line 78: I believe this is baking in an assumption that your libraries are located in the `shared-home` volume. Currently, if customers want to use a different volume (e.g. they will create a libraryVolume just for jar files), then they cannot use it in synchrony.

In Confluence stateful set we are using additional volumes - https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/main/charts/confluence/templates/_helpers.tpl#L237

This allows you to define `confluence.additionalLibraries` but if you are using different than local/shared home persistent volume, you also need to add the volume definition to `volumes.additional`. It is probably not the most elegant, but it is possible. For synchrony with this change, it won't be possible as `volumes.additional` is not mounted to synchrony statefulSet.

I think we can either add new additional volumes for synchrony or also add the `volumes.additional` the same way as it is currently done for Confluence. I would prefer just mounting the same volumes as it is a niche use case (ideally you use shared-home volume).

```
{{- with .Values.volumes.additional }}
{{- toYaml . | nindent 0 }}
{{- end }}
```

Please read through https://github.com/atlassian-labs/data-center-helm-charts/blob/master/CONFIG.md#additional-libraries--plugins especially:
> If you're not using the shared-home volume, then you can declare your own custom volume in the "Additional Volumes" section below, then declare the libraries as above (but with your custom volume name).

nanux (changes requested):

src/main/charts/confluence/values.yaml
errcode1202, line 191: Not related to this PR, but I realised when looking at the values for our `synchrony` definition that we do not include a `resources` definition for it as we do our other product `StatefulSet's`. I think we probably should, there are specifics that would be nice to tune for the synchrony pod i.e. JVM memory requirements and resource `request` and `limit` sizings, see here: https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/main/charts/confluence/values.yaml#L106

nanux: Thanks @errcode1202, created https://bulldog.internal.atlassian.com/browse/DCKUBE-322

nanux (approved):

PR #148, opened 2021-05-26
DCKUBE-311: Static and dynamic storage provisioning with AWS CSI

errcode1202:
Examples on how the AWS CSI can be used to create and utilise storage both statically and dynamically.


tan-ro (approved):

nanux (approved):

nanux:
As discussed IRL, please remove the dynamic provisioning from EFS and remove static provisioning from EBS.
I think we are not really testing for these use cases and they are more confusing for customers. They are free to use them but I would recommend sticking with static EFS and dynamic EBS.

PR #150, opened 2021-05-27
DCKUBE-329: Add detail regarding use of the config value 'use-forwarded-headers'

errcode1202:
Add detail regarding use of the config value 'use-forwarded-headers'


wkritzinger-atlassian (approved):

nanux:
Should we add an example for this at least for nginx-ingress controller? I think is still a bit confusing and I am worried it is now very easy to skip and being confused by it in product.

errcode1202:
> Should we add an example for this at least for nginx-ingress controller? I think is still a bit confusing and I am worried it is now very easy to skip and being confused by it in product.

If this documentation is confusing I'd rather address that first as opposed to adding an example. Taking that a step further if we still want an example then I'd be inclined to provide the full setup instructions and config we use for our ingress-nginx

nanux (approved):

PR #151, opened 2021-05-27
DCKUBE-322: Add resources stanza for Synchrony and inject values into startup

tarka:
This adds the resources stanza, and injects those values into the generated Synchrony startup script. The resulting running command looks like:
`java -Xss2048k -Xmx1g -XX:ActiveProcessorCount=2 -classpath /opt/atlassian/confluence/confluence/WEB-INF/packages/synchrony-standalone.jar:/opt/atlassian/confluence/confluence/WEB-INF/lib/* synchrony.core sql`

### DIFF
--- a/src/main/charts/confluence/templates/synchrony-start-script.yaml
+++ b/src/main/charts/confluence/templates/synchrony-start-script.yaml
@@ -11,6 +11,8 @@ data:
     #!/usr/bin/env bash
 
     java \
+       -Xss{{ .Values.synchrony.resources.jvm.stackSize }} -Xmx{{ .Values.synchrony.resources.jvm.maxHeap }} \
+       -XX:ActiveProcessorCount={{ .Values.synchrony.resources.container.requests.cpu }} \
        -classpath /opt/atlassian/confluence/confluence/WEB-INF/packages/synchrony-standalone.jar:/opt/atlassian/confluence/confluence/WEB-INF/lib/* \
        synchrony.core \
        sql
--- a/src/main/charts/confluence/templates/synchrony-start-script.yaml
+++ b/src/main/charts/confluence/templates/synchrony-start-script.yaml
@@ -11,7 +11,9 @@ data:
     #!/usr/bin/env bash
 
     java \
-       -Xss{{ .Values.synchrony.resources.jvm.stackSize }} -Xmx{{ .Values.synchrony.resources.jvm.maxHeap }} \
+       -Xms{{ .Values.synchrony.resources.jvm.minHeap }} \
+       -Xmx{{ .Values.synchrony.resources.jvm.maxHeap }} \
+       -Xss{{ .Values.synchrony.resources.jvm.stackSize }} \
        -XX:ActiveProcessorCount={{ .Values.synchrony.resources.container.requests.cpu }} \
        -classpath /opt/atlassian/confluence/confluence/WEB-INF/packages/synchrony-standalone.jar:/opt/atlassian/confluence/confluence/WEB-INF/lib/* \
        synchrony.core \
--- a/src/main/charts/confluence/values.yaml
+++ b/src/main/charts/confluence/values.yaml
@@ -183,6 +183,18 @@ synchrony:
     periodSeconds: 1
     # -- The number of consecutive failures of the Synchrony container readiness probe before the pod fails readiness checks
     failureThreshold: 30
+  resources:
+    jvm:
+      # -- The maximum amount of heap memory that will be used by the Synchrony JVM
+      maxHeap: "1g"
+      # -- The minimum amount of heap memory that will be used by the Synchrony JVM
+      minHeap: "1g"
+      # -- The memory allocated for the Synchrony stack
+      stackSize: "2048k"
+    container: 
+      requests:
+        cpu: "2"
+        memory: "2G"
   # -- The base URL of the Synchrony service.
   # This will be the URL that users' browsers will be given to communicate with Synchrony, as well as the URL that the
   # Confluence service will use to communicate directly with Synchrony, so the URL must be resovable both from inside and
--- a/src/main/charts/confluence/values.yaml
+++ b/src/main/charts/confluence/values.yaml
@@ -188,7 +188,7 @@ synchrony:
       # -- The maximum amount of heap memory that will be used by the Synchrony JVM
       maxHeap: "1g"
       # -- The minimum amount of heap memory that will be used by the Synchrony JVM
-      minHeap: "1g"
+      minHeap: "2g"
       # -- The memory allocated for the Synchrony stack
       stackSize: "2048k"
     container: 
--- a/src/main/charts/confluence/values.yaml
+++ b/src/main/charts/confluence/values.yaml
@@ -185,16 +185,16 @@ synchrony:
     failureThreshold: 30
   resources:
     jvm:
-      # -- The maximum amount of heap memory that will be used by the Synchrony JVM
-      maxHeap: "1g"
       # -- The minimum amount of heap memory that will be used by the Synchrony JVM
-      minHeap: "2g"
+      minHeap: "1g"
+      # -- The maximum amount of heap memory that will be used by the Synchrony JVM
+      maxHeap: "2g"
       # -- The memory allocated for the Synchrony stack
       stackSize: "2048k"
     container: 
       requests:
         cpu: "2"
-        memory: "2G"
+        memory: "2.5G"
   # -- The base URL of the Synchrony service.
   # This will be the URL that users' browsers will be given to communicate with Synchrony, as well as the URL that the
   # Confluence service will use to communicate directly with Synchrony, so the URL must be resovable both from inside and
--- a/src/main/charts/confluence/values.yaml
+++ b/src/main/charts/confluence/values.yaml
@@ -165,7 +165,7 @@ synchrony:
   # -- True if Synchrony (i.e. collaborative editing) should be enabled.
   # This will result in a separate StatefulSet and Service to be created for Synchrony.
   # If disabled, then collaborative editing will be disabled in Confluence.
-  enabled: false
+  enabled: true
   service:
     # -- The port on which the Synchrony Kubernetes service will listen
     port: 80
--- a/src/test/java/test/SynchronyTest.java
+++ b/src/test/java/test/SynchronyTest.java
@@ -41,10 +41,9 @@ class SynchronyTest {
 
     @ParameterizedTest
     @EnumSource(value = Product.class, names = "confluence")
-    void synchrony_params(Product product) throws Exception {
+    void synchrony_entrypoint(Product product) throws Exception {
         final var resources = helm.captureKubeResourcesFromHelmChart(product, Map.of(
-                "synchrony.enabled", "true",
-                "synchrony.ingressUrl", "https://mysynchrony"
+                "synchrony.enabled", "true"
         ));
 
         final var entrypoint = resources.get(Kind.ConfigMap, product.getHelmReleaseName() + "-synchrony-entrypoint")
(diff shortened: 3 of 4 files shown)


wkritzinger-atlassian (approved):

src/main/charts/confluence/values.yaml
```diff
@@ -183,6 +183,18 @@ synchrony:
...
     failureThreshold: 30
+  resources:
+    jvm:
+      # -- The maximum amount of heap memory that will be used by the Synchrony JVM
+      maxHeap: "1g"
+      # -- The minimum amount of heap memory that will be used by the Synchrony JVM
+      minHeap: "1g"
+      # -- The memory allocated for the Synchrony stack
+      stackSize: "2048k"
+    container:
```
wkritzinger-atlassian, line 194: Are these new values documented anywhere?

wkritzinger-atlassian (approved):

src/main/charts/confluence/values.yaml
tarka: Not really; `-Xmx` is pretty standard, and the stack size (`-Xss`) is in template `start-synchrony.sh` script that you're required to edit by hand. I've just ported them across.

wkritzinger-atlassian: Hm yes but these are technically new values you introduced, and they should be documented in `src/main/charts/confluence/README.md` right?

src/main/charts/confluence/values.yaml
```diff
@@ -183,6 +183,18 @@ synchrony:
     periodSeconds: 1
     # -- The number of consecutive failures of the Synchrony container readiness probe before the pod fails readiness checks
     failureThreshold: 30
+  resources:
+    jvm:
+      # -- The maximum amount of heap memory that will be used by the Synchrony JVM
+      maxHeap: "1g"
+      # -- The minimum amount of heap memory that will be used by the Synchrony JVM
+      minHeap: "1g"
+      # -- The memory allocated for the Synchrony stack
```
nghazali, line 192: Don't we need to define reserveCodeCache? The default for this in our Docker Releases is 256m

src/main/charts/confluence/values.yaml
```diff
@@ -183,6 +183,18 @@ synchrony:
...
+      stackSize: "2048k"
+    container: 
+      requests:
+        cpu: "2"
```
errcode1202, line 196: We've included the JVM arg `XX:ActiveProcessorCount` in the other product `values.yaml`  see https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/main/charts/bitbucket/values.yaml#L131

It should probably be set of synchrony too.

src/main/charts/confluence/values.yaml
```diff
@@ -183,6 +183,18 @@ synchrony:
     periodSeconds: 1
     # -- The number of consecutive failures of the Synchrony container readiness probe before the pod fails readiness checks
     failureThreshold: 30
+  resources:
```
errcode1202, line 186: Might be nice to update the test `[output.yaml](https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/test/resources/expected_helm_output/confluence/output.yaml)`  for confluence to include these values so that their presence and corresponding values are tested in future

src/main/charts/confluence/templates/synchrony-start-script.yaml
```diff
@@ -11,7 +11,9 @@ data:
     #!/usr/bin/env bash
 
     java \
+       -Xss{{ .Values.synchrony.resources.jvm.stackSize }} -Xmx{{ .Values.synchrony.resources.jvm.maxHeap }} \
+       -XX:ActiveProcessorCount={{ .Values.synchrony.resources.container.requests.cpu }} \
```
nanux, line 15: @errcode1202 this could be a potential way how to automate the CPU requests & jvm args. Did we test with `m` units? E.g. `-XX:ActiveProcessorCount=250m`?

src/main/charts/confluence/values.yaml
nanux: The `HelmOutputComparisonTest` is verifying an actual output of the processed `helm template` with default values and compare it to the recorded response. If the default is not changed, then the test doesn't see any difference to compare.

The main problem is that the default is `synchrony.enabled: false`. I think we have two options.

1. Enable synchrony by default, this seems like a better option for customers as they get more functionality out of the box. The risk is that there might be some misconfiguration why it doesn't work. @badgersow, what do you think, does it make sense to enable synchrony by default?
2. Add `synchrony.enabled: true` to https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/test/resources/expected_helm_output/confluence/values.yaml and that will enable it at least for the comparison test.

Either way I think we should test for synchrony output as it is fairly chunky part of the solution that is not covered much (we have `SynchronyTest` unit test).

src/main/charts/confluence/values.yaml
nanux: The `README.md` is automatically generated from the `values.yaml` file with https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/main/scripts/generate_helm_docs.sh
So the main thing is to document everything in the `values.yaml` files and we generate the docs later. Right now it is before release but should be automated.

src/main/charts/confluence/values.yaml
```diff
@@ -183,6 +183,18 @@ synchrony:
...
+      # -- The maximum amount of heap memory that will be used by the Synchrony JVM
+      maxHeap: "1g"
+      # -- The minimum amount of heap memory that will be used by the Synchrony JVM
+      minHeap: "1g"
+      # -- The memory allocated for the Synchrony stack
+      stackSize: "2048k"
+    container: 
+      requests:
+        cpu: "2"
+        memory: "2G"
```
nanux, line 197: @badgersow do we need 2CPU and 2G of memory for a single synchrony unit? It seems pretty hungry.

src/main/charts/confluence/values.yaml
errcode1202: Yup - What I was getting at, we should test for synchrony output

src/main/charts/confluence/values.yaml
errcode1202: Ahh this might not be necessary actually because its being set in `synchrony-start-script.yaml `

src/main/charts/confluence/values.yaml
errcode1202: Confuence itself has a `reserverCodeCache` of 256m not sure about synchrony.

src/main/charts/confluence/templates/synchrony-start-script.yaml
errcode1202: Yup millicores were tested. From what I remember it will round up/down to the nearest whole cpu. My question here is why does only `250m` (0.25) of a cpu for synchrony? Why not just give it a whole core?

src/main/charts/confluence/values.yaml
errcode1202: Does it make sense to use the same values as connie? They are based on the confluence container jvm specs: https://bitbucket.org/atlassian-docker/docker-atlassian-confluence-server/src/master/

src/main/charts/confluence/values.yaml
tarka: I've add a test for the entrypoint output

src/main/charts/confluence/values.yaml
tarka: Yeah, it gets set in the generated script, so isn't required here.

Personally I'm not a fan of putting the JVM flags in `values.yml`. IMHO it should hold facts about the infrastructure; deriving settings from those values is the job of the template. Separation of code from data, basically. But I'm not sure what is idiomatic in Helm though.

src/main/charts/confluence/values.yaml
tarka: We don't set it the Quickstarts, nor in the managed Synchrony, so I don't think it's necessary. @badgersow may know more though.

src/main/charts/confluence/values.yaml
errcode1202: @tarka fair point. If we decide to use [helmfile](https://hello.atlassian.net/wiki/spaces/ITPLAT/pages/1170780454/k8s+-+Managing+our+stacks+with+helmfile) it will help on this front.

errcode1202 (approved):
🚀

src/main/charts/confluence/values.yaml
```diff
@@ -183,6 +183,18 @@ synchrony:
     periodSeconds: 1
     # -- The number of consecutive failures of the Synchrony container readiness probe before the pod fails readiness checks
     failureThreshold: 30
+  resources:
+    jvm:
+      # -- The maximum amount of heap memory that will be used by the Synchrony JVM
+      maxHeap: "1g"
```
badgersow, line 189: I think we should have it at 2Gb to keep consistent with how Confluence creates Managed Synchrony by default. Please see [this file (internal repo)](https://stash.atlassian.com/projects/CONFSERVER/repos/confluence/browse/confluence-plugins/confluence-bundled-plugins/synchrony-interop/src/main/resources/env/synchrony.properties#7) with default properties and [this ticket (public Jira)](https://jira.atlassian.com/browse/CONFSERVER-58401) for default heap size change.

src/main/charts/confluence/values.yaml
badgersow: We should definitely enable synchrony by default. The next `7.12.2` version of Confluence should have all changes for Collab editing to be also enabled by default. Confluence builds test the case when Synchrony is enabled.

src/main/charts/confluence/values.yaml
badgersow: We don't use it in managed Synchrony, and I'm for keeping all deployment parameters consistent

src/main/charts/confluence/values.yaml
badgersow: I think we should have separate JVM arg values for Synchrony and Connie. I've seen that large customers spend a lot of time fine-tuning their instances and they will be upset if we don't support separate configuration for Synchrony and Confluence JVM. The typical example for enterprise customer would be to give Connie 16Gb of heap and Synchrony 4Gb of heap - they can be different by the order of magnitude.

badgersow (changes requested):
I've left some comments related to JVM heap size and memory request. Managed synchrony is using 2Gb for heap size. I think we should keep helm charts consistent with that.
Apart from that, everything looks good.

nghazali (approved):

src/main/charts/confluence/values.yaml
tarka: Should I do this as part of this PR, so should we treat that separately?

src/test/java/test/SynchronyTest.java
```diff
@@ -38,4 +38,21 @@ void synchrony_enable(Product product) throws Exception {
                 .hasTextContaining("-Dsynchrony.service.url=https://mysynchrony/v1")
                 .hasTextNotContaining("synchrony.btf.disabled");
     }
+
+    @ParameterizedTest
```
nanux, line 42: Thanks for creating the test. ⭐ 

Minor - we don't really need it parametrized as it will have only one execution for Confluence. But I can see we use `@ParameterizedTest` above as well so we can keep it for consistency.

src/main/charts/confluence/values.yaml
nanux: We can do it as part of this PR but I can see you've already done it. 🙇‍♂️

nanux (approved):

src/test/resources/expected_helm_output/confluence/output.yaml
```diff
@@ -60,6 +106,70 @@ spec:
     app.kubernetes.io/name: confluence
     app.kubernetes.io/instance: unittest-confluence
 ---
+# Source: confluence/templates/statefulset-synchrony.yaml
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: unittest-confluence-synchrony
+  labels:
+    helm.sh/chart: confluence-0.9.0
```
nanux, line 115: The `0.9.0` strings need to be updated to `0.10.0`, currently, the builds are failing.

- [ ] update version to fix the build

badgersow (approved):

PR #152, opened 2021-05-27
DCKUBE-302 Add Bitbucket SSH service and custom annotations.

wkritzinger-atlassian:


src/main/charts/bitbucket/templates/service.yaml
```diff
@@ -22,3 +26,27 @@ spec:
       name: hazelcast
   selector:
     {{- include "bitbucket.selectorLabels" . | nindent 4 }}
+
+{{/*This describes a K8s service for comunicating with the ssh port on Bitbucket*/}}
```
nanux, line 30:
```suggestion
{{/* This describes a K8s service for communicating with the ssh port on Bitbucket */}}
```

src/main/charts/bitbucket/values.yaml
```diff
@@ -52,6 +52,12 @@ bitbucket:
     port: 80
     # -- The type of Kubernetes service to use for Bitbucket
     type: ClusterIP
+    annotations: {}
```
nanux, line 55: This breaks the expected output test - https://github.com/atlassian-labs/data-center-helm-charts/runs/2682877276?check_suite_focus=true

src/main/charts/bitbucket/values.yaml
```diff
@@ -52,6 +52,12 @@ bitbucket:
...
+  sshService:
```
nanux, line 56: I think we definitely need some inline documentation in the values file so users can understand how and when these values should be utilised.

Beyond the inline comments, I think we should have a separate section in configuration explaining SSH setup in kubernetes as it is a bit mindbending experience :)

nanux (changes requested):

errcode1202 (approved):

src/main/charts/bitbucket/values.yaml
bianchi2: @nanux do you mean port forwarding? In my clusters I follow https://kubernetes.github.io/ingress-nginx/user-guide/exposing-tcp-udp-services/ but instructions will vary depending on the ingress itself.

src/test/resources/expected_helm_output/bitbucket/output.yaml
```diff
@@ -39,6 +39,7 @@ metadata:
     app.kubernetes.io/instance: unittest-bitbucket
     app.kubernetes.io/version: "7.12.1-jdk11"
     app.kubernetes.io/managed-by: Helm
+  annotations:
```
nghazali, line 42: if there is no annotation to set then good to have `annotations: {}`

src/main/charts/bitbucket/values.yaml
wkritzinger-atlassian: Fixed

src/main/charts/bitbucket/values.yaml
wkritzinger-atlassian: Added some inline documentation for these properties

src/test/resources/expected_helm_output/bitbucket/output.yaml
wkritzinger-atlassian: this is for the expected output though, right? I can't really control that with the values

src/main/charts/bitbucket/values.yaml
```diff
@@ -52,6 +52,17 @@ bitbucket:
     port: 80
     # -- The type of Kubernetes service to use for Bitbucket
     type: ClusterIP
+    annotations: {}
+  # -- Enable or disable an additional service for exposing SSH for external access.
+  # Disable when the SSH service is exposed through the ingress controller, or enable if the ingress controller does
+  # not support TCP.
```
nanux, line 58: I think here we should link customers to the more substantial documentation describing Ingress configuraion that is going to be created.

src/test/config/bitbucket/values-EKS.yaml
```diff
@@ -7,6 +7,16 @@ ingress:
...
   host: ${helm.release.prefix}-bitbucket.${eks.ingress.domain}
 
+bitbucket:
+  sshService:
+    enabled: true
+    annotations:
+      external-dns.alpha.kubernetes.io/hostname: ${helm.release.prefix}-bitbucket-ssh.${eks.ingress.domain}
+      service.beta.kubernetes.io/aws-load-balancer-internal: 'true'
+  additionalEnvironmentVariables:
+    - name: PLUGIN_SSH_BASEURL
```
nanux, line 17: I believe this is the [system property](https://confluence.atlassian.com/bitbucketserver/configuration-properties-776640155.html#Configurationproperties-plugin.ssh.baseurl) supported by BB. This something customers need to configure as well in case they are using separate domain for ssh (and they don't want to configure it manually). Is this something we will have in the documentation?

src/main/charts/bitbucket/templates/service.yaml
```diff
@@ -22,3 +26,27 @@ spec:
       name: hazelcast
   selector:
     {{- include "bitbucket.selectorLabels" . | nindent 4 }}
+
+{{/* This describes a K8s service for communicating with the ssh port on Bitbucket */}}
```
nanux, line 30: To be consistent with other charts (e.g. synchrony in confluence) I believe this service should be defined in a separate file - `service-ssh.yaml` so it is clearer for anybody who browses the repo.

nanux (changes requested):
Could you please separate the ssh service to a new file?

I think this would also warrant a unit test similar to https://github.com/atlassian-labs/data-center-helm-charts/blob/f02f5debb79f3994ff46a32de40e244e03fb2678/src/test/java/test/SynchronyTest.java

src/main/charts/bitbucket/values.yaml
bianchi2: @nanux in this case it's EKS specific (which is fine). In our tests we create LoadBalancer svc and tell AWS a few additional details. Maybe we need an additional docs page (perhaps not part of this PR) to explain how clouds or standalone ingress controllers can/may handle ssh traffic, say, Nginx port forwarding etc. Maybe just give direction and briefly explain why a standard ingress object won't do?

bianchi2 (approved):

src/main/charts/bitbucket/values.yaml
nanux: It is a good callout. We have discussed it today on our standup that we need to create a dedicated documentation and created https://bulldog.internal.atlassian.com/browse/DCKUBE-343

I will add your note there.

src/test/config/bitbucket/values-EKS.yaml
wkritzinger-atlassian: This is only necessary if you need automatic setup. You can always set this manually in the UI. I think we have a documentation for `bitbucket.properties` somewhere, where this property is documented.

Here it is: https://confluence.atlassian.com/bitbucketserver/configuration-properties-776640155.html#Configurationproperties-plugin.ssh.baseurl

wkritzinger-atlassian: I'll make a note of it in the docs page for this service.

wkritzinger-atlassian:
> Could you please separate the ssh service to a new file?
> 
> I think this would also warrant a unit test similar to https://github.com/atlassian-labs/data-center-helm-charts/blob/f02f5debb79f3994ff46a32de40e244e03fb2678/src/test/java/test/SynchronyTest.java

Done!

PR #158, opened 2021-06-01
Fix broken link in the installation guide

Mishelles:
There was a typo in the installation guide: helm-chart link was broken. We spent some time before actually guess to append one letter to the link :)


nanux (approved):

nanux:
Good catch! Thank you very much @Mishelles.

PR #165, opened 2021-06-03
DCKUBE-345: Enable clustering by default for all products

nanux:
* Switching clustering enabled by default
* The changes that will be applied to the default product manifests are visible in file with the expected output for each product


tarka (approved):

apawelczyk-atlassian:
[ClusteringTest](https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/test/java/test/ClusteringTest.java) doesn't make much sense now. Maybe we should test whether clustering can be *disabled* instead?

apawelczyk-atlassian (approved):
[ClusteringTest](https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/test/java/test/ClusteringTest.java) doesn't make much sense now. Maybe we should test whether clustering can be *disabled* instead?

nghazali (approved):

errcode1202 (approved):

nanux:
I am closing the PR as we are not going to tackle it now to enable users to create an installation with default values.

nanux:
I am closing the PR as we are not going to tackle it now to enable users to create an installation with default values.

PR #166, opened 2021-06-04
Create SSH_BITBUCKET.MD

tan-ro:


docs/examples/ssh/SSH_BITBUCKET.md
```diff
@@ -0,0 +1,61 @@
...
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tcp-services
+  namespace: ingress-nginx
+data:
+  7999: "ssh-test/bitbucket:ssh"
+```
+Nex, we have to edit the `deployment` of the ingress controller and add the `--tcp-services-configmap` option:
```
wkritzinger-atlassian, line 22:
```suggestion
Next, we have to edit the `deployment` of the ingress controller and add the `--tcp-services-configmap` option:
```

tan-ro: Thank, good catch!

docs/examples/ssh/SSH_BITBUCKET.md
```diff
@@ -0,0 +1,61 @@
+# SSH service in Bitbucket on Kubernetes
+
+In addition to providing a service on HTTP(S), Bitbucket also allows remote Git operations over SSH connections. By default, Kubernetes Ingress controllers only work for HTTP connections, but some ingress controllers also support TCP connections.
+
+Depending on the need of your deployment, SSH access can be provided through two mechanisms:
+1. Opening the TCP port through the ingress controller - This option should be used if the SSH service is required to be available on the same DNS name as the HTTP service.
+2. Creating a separate Kubernetes `LoadBalancer` service - This option is available if the ingress controller does not support TCP connections, or if you don’t need your deployment to have the SSH service available on the same DNS name as the HTTP service.
+
+## Setting up the NGINX Ingress Controller to accept SSH connections
```
wkritzinger-atlassian, line 9: I think we have to scrap this option for now. Dylan just tested it again on Kubernetes 1.19 and it does not seem to work as expected. In my testing I ran into a bug that I thought was unrelated to I didn't think much of it, but the combination of using HTTPS with the proxy protocol as well as a TCP service on the same kubernetes service (and therefore ELB on AWS) does not currently work. There's the possibility that it will work on Kubernetes 1.20 with the `appProtocol` field, but this is not tested yet.

bianchi2: This works for me on an on-prem k8s cluster. It will even work on Minikube. Why does k8s matter here if it's just all about port forwarding from Nginx to service directly?

bianchi2 (approved):

docs/examples/ssh/SSH_BITBUCKET.md
bianchi2: So, here's the configmap:

```
eugene:~$ kubectl get cm/tcp-services -n ingress-nginx -o yaml
apiVersion: v1
data:
  "7999": git/bitbucket:7999
kind: ConfigMap
  name: tcp-services
  namespace: ingress-nginx

```
And git clone works just fine:

```
eugene:/tmp$ git clone ssh://git@git.k8sdemo.tk:7999/k8s/kubernetes.git
Cloning into 'kubernetes'...
remote: Enumerating objects: 32, done.
remote: Counting objects: 100% (32/32), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 32 (delta 3), reused 0 (delta 0), pack-reused 0
Receiving objects: 100% (32/32), done.
Resolving deltas: 100% (3/3), done.
```
Since this is my on-prem self made k8s cluster, Nginx listens on host port on the node (host network=true) but it's just a matter of configuration of course. In clouds Nginx service will be LoadBalancer in most cases.

My cluster is v1.18.10

errcode1202: @bianchi2 it would be good to understand how you have configured your `ingress-nginx` controller. We're basing the installation and config on our ingress readme (still in review) here: https://github.com/atlassian-labs/data-center-helm-charts/blob/dckube-338-eks-setup-examples/docs/examples/ingress/INGRESS_NGINX.md#2-create-a-custom-config-yaml

wkritzinger-atlassian (approved):

docs/examples/ssh/SSH_BITBUCKET.md
bianchi2: @errcode1202 like I said, my cluster is an on-prem self made one. Nginx is deployed with hostNetwork=true. As per docs, you need to add this additional ssh port to Nginx service. I haven't tried it in EKS, though it did work for me on Digital Ocean and self made K8s clusters with Nginx deployed (hostNetwork). I'll give it a go myaself on EKS and open a PR if my tests are successful.

wkritzinger-atlassian: Yep, if it's on-prem it should work, however, we are giving examples for AWS and with that it won't work. If we specifically call out that it won't work with AWS (and potentially other cloud providers) then we might be able to keep this section.

docs/examples/ssh/SSH_BITBUCKET.md
```diff
@@ -0,0 +1,24 @@
+# SSH service in Bitbucket on Kubernetes
+
+In addition to providing a service on HTTP(S), Bitbucket also allows remote Git operations over SSH connections. By default, Kubernetes Ingress controllers only work for HTTP connections, but some ingress controllers also support TCP connections.
+
+You can provide SSH access by creating a separate Kubernetes `LoadBalancer` service, but you can do this only if you don’t need your deployment to have the SSH service available on the same DNS name as the HTTP service.
```
nanux, line 5: Should we swap the condition? It is a big _if_.

> If you don’t need your deployment to have the SSH service available on the same DNS name as the HTTP service you can provide SSH access ....

docs/examples/ssh/SSH_BITBUCKET.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Creating a separate Kubernetes LoadBalancer service for SSH connections on AWS
+In the values file for the helm chart, the extra SSH service can be enabled like this:
+```
+bitbucket:
+  sshService:
+    enabled: true
+```
+On a deployment using AWS, assuming you have [external-dns](https://github.com/kubernetes-sigs/external-dns) configured, you can add these annotations to automatically set up the DNS name for the SSH service:
```
nanux, line 14: It is not clear if this is optional or not. I would say if we are talking about having a different domain name for the ssh service then this is not optional or if we make `external-dns` optional, then we need to also say that if you don't have it, you need to create DNS configuration for the newly created loadbalancer.

docs/examples/ssh/SSH_BITBUCKET.md
nanux, line 5: When I am thinking about this, could we structure the document like:

## Purpose/goal/...

In addition to providing...

## Prerequisites

1. You can run your SSH service on a different domain name
2. You have external-dns (?optional/recommended)

## Steps
1. Enable SSH service
2. (Optional) Add DNS annotation
3. Apply new values

tan-ro: Yeah, good call!

docs/examples/ssh/SSH_BITBUCKET.md
wkritzinger-atlassian: Yep also a good call - `external-dns` just helps with setting up the DNS name, but you can also do it manually of course

docs/examples/ssh/SSH_BITBUCKET.md
wkritzinger-atlassian: yep that also looks good, much better structure

docs/examples/ssh/SSH_BITBUCKET.md
errcode1202: Okay - In the meantime Im testing the alternative approach to see if thats a viable fallback for AWS

docs/examples/ssh/SSH_BITBUCKET.md
```diff
@@ -0,0 +1,32 @@
...
+   ```
+   bitbucket:
+      annotations:
+         external-dns.alpha.kubernetes.io/hostname: bitbucket-ssh.example.com
+   ```
+   This step is optional, you can set up the DNS name manually.
+
+3. Apply new values: 
+
+   ```
```
nanux, line 27: This should be part of step 2.

```
bitbucket:
  annotations:
    external-dns.alpha.kubernetes.io/hostname: bitbucket-ssh.example.com
  additionalEnvironmentVariables:
    - name: PLUGIN_SSH_BASEURL
      value: ssh://bitbucket-ssh.example.com/
```

Step 3 should be just **Apply new values**.

tan-ro: Thanks, done.

nanux (approved):

PR #173, opened 2021-06-13
Defining the following values in the helpers template for each chart,…

LouisZschaler:
We use the Atlassian Helm Charts in several umbrella charts but we'd like to get rid of the antipattern in our umbrella Helm charts where we implement chart logic via the chart's values file. This prevents us from adding more values where the umbrella Helm charts are actually used, e.g. defining more additionalInitContainers, because this would override the additionalInitContainers from our umbrella Helm charts.

In order to achieve that I applied some small, non-breaking changes to the upstream Atlassian Helm charts. I chose variables where I expect this procedure to be useful for us and others.


nanux:
@pathob and @LouisZschaler, thanks for bringing this use case to our attention. I don't see why we could not merge this PR as it is making it flexible for some use cases (e.g. where the helm charts are used as a dependency) and not impacting current users.

We are just in process of setting up a full CI pipeline for this branch so we can be confident that there aren't some uncovered cases. Part of the CI is running outside of Github Actions and we need to set it up manually for this branch as it is coming from a fork.

We will come back with the results (and 👍 if everything is 🍏) shortly.

errcode1202 (approved):
These changes look good with our internal testing: 
https://server-syd-bamboo.internal.atlassian.com/browse/DCNG-K8SHELMTEST215-4

There is a failing test, although it is also currently seen on `master` indicating it is not related to these changes:
https://server-syd-bamboo.internal.atlassian.com/browse/DCNG-K8SHELMTEST-BBSAWSDEFAULTS-319

Approving and merging

nanux (approved):

PR #192, opened 2021-06-28
DCKUBE-424 latest lts product helm chart

nghazali:
Introduced a script to detect latest LTS of products in scope (Stash, Confluence and Jira-software) and put it as docker image version in helm_install.sh


errcode1202:
Just an FYI Nasser, @nanux asked if we could keep the PR titles and commits in line with his commentary here:
https://atlassian.slack.com/archives/C017S5V4QG3/p1623221711153000

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,63 @@
+import sys
```
errcode1202, line 1: How is this script `get_product_lts_version.py` ultimately called?  I expected to see it in `helm_install.sh` but I dont? 🤔

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,63 @@
...
+def get_lts_version(argv):
+	lts_version=''
+	product = argv[0].lower()
+	if product == 'jira':
+		product = 'jira-software'
+	elif product == 'bitbucket' :
+		product = 'stash'
+
+	if product in known_lts_version:
+		url = f"https://my.atlassian.com/download/feeds/archived/{product}.json"
```
errcode1202, line 20: Just looking at the raw data from these respective product URL's, from what they present I dont see how the LTS version can be calculated based on the code that parses it below?

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,63 @@
...
+		try:
+			fl = urllib.request.urlopen(url)
+			fdata = fl.read()
+			jsdata = json.loads(fdata[10:len(fdata)-1].decode("utf-8"))
+			enterprise_edition = [x for x in jsdata if x['edition'].lower() == 'enterprise']
+			sortedData = sorted(enterprise_edition, key=lambda k:cversion(k['version']), reverse=True)
+
+			if len(sortedData)>0 :
+				lts_version = sortedData[0]['version']
+			# print (f"Succeed!  LTS Version for {argv[0]} is {lts_version}")
```
errcode1202, line 31: not sure if this is intended to be commented out? If so we should remove it

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,63 @@
...
+			else:
+				lts_version = known_lts_version[product]
+			# print(f"Failed to find the LTS Version for {product} dynamically. It is replaced by '{known_lts_version[product]}'")
```
errcode1202, line 34: not sure if this is intended to be commented out? If so we should remove it

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,63 @@
+import sys
+import urllib.request
+import json
+
+known_lts_version = {
+	'jira-software':'8.13.7',
```
errcode1202, line 6: Assuming this map should be using the latest LTS product versions? If so the current LTS for Jira is `8.13.8`.

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,63 @@
...
+	'confluence':'7.12.2',
```
errcode1202, line 7: Latest LTS for confluence is `7.4.9`

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,63 @@
...
+	'stash':'7.12.1',
```
errcode1202, line 8: Current LTS for Stash is `7.6.7`

src/test/scripts/get_product_lts_version.py
errcode1202: Im also not sure how this script `get_product_lts_version.py` is ultimately used?  I expected to see it in `helm_install.sh` but I dont? 🤔

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,63 @@
+import sys
+import urllib.request
+import json
+
+known_lts_version = {
```
errcode1202, line 5: These appear to be the the latest product releases, and not the current LTS versions for each product. Current LTS versions are: `8.13.8`/`7.4.9`/`7.6.7`

src/test/scripts/get_product_lts_version.py
nghazali: This scripts fetch all releases of product, filter Enterprise editions (LTS versions) and get the latest version by sorting them and picking the first one. 

Also this script is called by HelmTestPlanSpec.java (through helm_install_chart) in bamboo plan (see https://stash.atlassian.com/projects/CSI/repos/kubernetes-ci/pull-requests/151/overview)

src/test/scripts/get_product_lts_version.py
nghazali: Check edition field (Standard/None means none LTS and Enterprise is LTS version)

src/test/scripts/get_product_lts_version.py
nghazali: We can remove these but I left this on purpose to be used for debug if needed in future

src/test/scripts/get_product_lts_version.py
nghazali: The same as above

src/test/scripts/get_product_lts_version.py
nghazali: This part is not needed at all but as currently LTS version of Confluence and Bitbucket does not support Kubernetes, so I added a logic to feed bamboo with a valid release until next LTS release for those products.

src/test/scripts/get_product_lts_version.py
errcode1202: Got ya, so `Enterprise` == `LTS`

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,61 @@
...
+		product = 'stash'
+
+	if product in known_lts_version:
+		url = f"https://my.atlassian.com/download/feeds/archived/{product}.json"
+
+		try:
+			fl = urllib.request.urlopen(url)
+			fdata = fl.read()
+			jsdata = json.loads(fdata[10:len(fdata)-1].decode("utf-8"))
+			enterprise_edition = [x for x in jsdata if x['edition'].lower() == 'enterprise']
```
errcode1202, line 26: Based on the previous comments, and to make it clear, it might be nice to change this variable name to `lts_edition`

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,61 @@
...
+			sortedData = sorted(enterprise_edition, key=lambda k:cversion(k['version']), reverse=True)
+
+			if len(sortedData) > 0:
+				lts_version = sortedData[0]['version']
+			else:
+				lts_version = known_lts_version[product]
+
+			# as currently latest lts versoin of bitbucket and confluence don't support k8s
```
errcode1202, line 34: typo here - `versoin` should be `version`

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,61 @@
...
+			# we use none-lts version of those products in the test
```
errcode1202, line 35: another subtle typo - `none-lts` should be `non-lts`

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,61 @@
...
+
+	return lts_version
+
+
+def cversion(version):
+	vers = [int(v) for v in version.split(".")]
+	mapped_ver = ''
+	for i in range(max(len(vers)-1, 4)):
+		if len(vers) > i:
+			mapped_ver = f'{mapped_ver}{str(vers[i]).zfill(5)}'
```
errcode1202, line 53: Its not very clear what this expression (`f'{mapped_ver}{str(vers[i]).zfill(5)}'`) is doing, could we extract it into its owned named function?

src/test/scripts/get_product_lts_version.py
```diff
@@ -0,0 +1,61 @@
...
+		else:
+			mapped_ver = f'{mapped_ver}00000'
```
errcode1202, line 55: Same as comment above, why we set `mapped_ver` to `f'{mapped_ver}00000'` isnt overly clear. It might also be better in a named function.

src/test/scripts/get_product_lts_version.py
nghazali: zfill(n) method in python add zeros to the left side of a number to make the length of string n (like padLeft). The purpose of this is making the version an unified string comparable
'7.12.1' => '00007000120000100000'
In this case we can apply string comparison in version and sort the list based on version correctly.

errcode1202 (changes requested):
Some fixes required.

src/test/scripts/get_product_lts_version.py
errcode1202: Wouldnt it be better to pull in a library to handle version parsing and comparison, this looks useful:
https://pypi.org/project/semantic-version/

src/test/scripts/get_product_lts_version.py
nghazali: This is for the case that the version missing minor, patch, or build - so replace with '00000' to makes the version unified for string comparison:

'7.12'             => '0000700120000000000'
'7.12.39'        => '0000700120003900000' 
'7.12.39.124' => '0000700120003900124'

errcode1202: Neat. But I think we should use a librabry for this, its much clearer:

```
>>> from packaging import version
>>> version.parse("2.3.1") < version.parse("10.1.2")
True
```
> https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python

src/test/scripts/get_product_lts_version.py
errcode1202: I think we should remove them

src/test/scripts/get_product_lts_version.py
nghazali: fixed

src/test/scripts/get_product_lts_version.py
nghazali: changed to `lts_versions`

src/test/scripts/get_product_lts_version.py
nghazali: fixed

src/test/scripts/get_product_lts_version.py
nghazali: I have implemented by this approach initially, but this is not the best for our case so I reverted this approach and refactored this part. This brings a dependency to packaging module which is not installed by default in ubuntu build server. If we go with this approach we need to install additional one which I preferred to add cversion method to script rather than add additional dependency.

src/test/scripts/get_product_lts_version.py
nghazali: Already removed them

src/test/scripts/product_versions.py
```diff
@@ -0,0 +1,61 @@
+import sys
+import urllib.request
+import json
+
+known_lts_version = {
```
yzha645, line 5: Instead of `known_lts_version`, will something like minimum supported version be better suited the purpose here?

src/test/scripts/get_product_lts_version.py
nghazali: packaging library is additional module dependency and we need to install in bamboo spec at some point. I preferred to not go to that direction and instead add a simple local method to do the job.

src/test/scripts/product_versions.py
nghazali: Good suggestion, I will change it.

errcode1202 (approved):

PR #195, opened 2021-07-02
DCKUBE-436: Should use NFS Service IP - POC

errcode1202:
🚧  These changes are for POC purposes.


src/main/charts/bitbucket/templates/shared-home-pv.yaml
```diff
@@ -22,6 +22,7 @@ spec:
     - rsize=32768
     - wsize=32768
     - _netdev
+    - nolock
```
errcode1202, line 25: Although this approach works where the NFS can be mounted via the Service, we have an issue when BB comes up and cannot aquire a lock on shared home. When this flag is added lock on shared home can be aquired. This of course is not the desired behaviour - shared home locking is required so that data is not inadvertently mutated. Need to understand why locking is not working. This could be down to the NFS base docker image that is being used i.e Centos

wkritzinger-atlassian:
Support for this was added in #205

PR #196, opened 2021-07-02
DCKUBE-429 Increase timeout for waiting for pods to come up.

wkritzinger-atlassian:
Sometimes the cluster-autoscaler takes too long to spin up new
nodes, this timeout should give enough time for it to spin up
nodes and then for the application to come up.

### DIFF
--- a/src/test/scripts/helm_install.sh
+++ b/src/test/scripts/helm_install.sh
@@ -114,7 +114,7 @@ bootstrap_database() {
   # Use the product name for the name of the postgres database, username, and password.
   # These must match the credentials stored in the Secret preloaded into the namespace,
   # which the application will use to connect to the database.
-  helm install -n "${TARGET_NAMESPACE}" --wait \
+  helm install -n "${TARGET_NAMESPACE}" --wait --timeout 15m \
      "$POSTGRES_RELEASE_NAME" \
      --values "$THISDIR/postgres-values.yaml" \
      --set fullnameOverride="$POSTGRES_RELEASE_NAME" \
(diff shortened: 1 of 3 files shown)


badgersow (approved):

tarka (approved):

src/test/scripts/helm_install.sh
```diff
@@ -114,7 +114,7 @@ bootstrap_database() {
   # Use the product name for the name of the postgres database, username, and password.
   # These must match the credentials stored in the Secret preloaded into the namespace,
   # which the application will use to connect to the database.
-  helm install -n "${TARGET_NAMESPACE}" --wait \
+  helm install -n "${TARGET_NAMESPACE}" --wait --timeout 15m \
```
errcode1202, line 117: Minor comment - is it worth turning these timeout values to constants?

errcode1202 (approved):

src/test/scripts/shared_home_browser_install.sh
```diff
@@ -8,4 +8,4 @@ TARGET_NAMESPACE=$1
 
 echo Starting shared home browser...
 kubectl apply -n "${TARGET_NAMESPACE}" -f "./../../../target/config/shared-home/shared-home-browser.yaml"
-kubectl wait -n "${TARGET_NAMESPACE}" --for=condition=Ready pod/shared-home-browser
+kubectl wait -n "${TARGET_NAMESPACE}" --for=condition=Ready pod/shared-home-browser --timeout 900s
```
errcode1202, line 11: same as previous comment

src/test/scripts/helm_install.sh
wkritzinger-atlassian: Not sure if it's worth it, the timeouts are scattered around a few different files, doing it "right" would mean adding it to the `helm_parameters` file and that one is quite messy already IMHO.

PR #198, opened 2021-07-05
Involved current feed in finding latest LTS version of products

nghazali:


src/test/scripts/product_versions.py
```diff
@@ -15,18 +15,25 @@ def get_lts_version(argv):
...
-			fl = urllib.request.urlopen(url)
-			fdata = fl.read()
-			jsdata = json.loads(fdata[10:len(fdata)-1].decode("utf-8"))
+			# load archived feeds
+			fdata_a = urllib.request.urlopen(url_archived).read()
+			jsdata = json.loads(fdata_a[10:len(fdata_a)-1].decode("utf-8"))
+
+			# load current feeds and append to archive
+			fdata_c = urllib.request.urlopen(url_current).read()
+			jsdata += json.loads(fdata_c[10:len(fdata_c)-1].decode("utf-8"))
```
wkritzinger-atlassian, line 29: Hmm not that the JSONP callback is likely to change anytime soon, but stripping the characters off just like that doesn't seem to robust. We could define something like a JSONP callback name constant and have a method that strips the padding from the JSON? We already use it at least twice here now

src/test/scripts/product_versions.py
```diff
@@ -15,18 +15,25 @@ def get_lts_version(argv):
 		product = 'jira-software'
 	elif product == 'bitbucket':
 		product = 'stash'
-
 	if product in known_supported_version:
```
wkritzinger-atlassian, line 18: The latest LTS version of Bitbucket (7.6) is not supported by this, will this be taken into account?

nghazali: We have the known_supported_versions defined to handle this case. If LTS version come from this script is less than known_supported_version, then it will be ignored.

src/test/scripts/product_versions.py
nghazali: Agree... refactored this part.

yzha645 (approved):

src/test/scripts/product_versions.py
```diff
@@ -15,18 +16,25 @@ def get_lts_version(argv):
...
-		url = f"https://my.atlassian.com/download/feeds/archived/{product}.json"
+		url_archived = f"https://my.atlassian.com/download/feeds/archived/{product}.json"
+		url_current = f"https://my.atlassian.com/download/feeds/current/{product}.json"
 
 		try:
-			fl = urllib.request.urlopen(url)
-			fdata = fl.read()
-			jsdata = json.loads(fdata[10:len(fdata)-1].decode("utf-8"))
+			# load archived feeds
```
errcode1202, line 24: Minor request can we be a bit more descriptive on the naming of the 4 variables below i.e.
`fdata_a`
`jsdata`
`fdata_c`
`jsdata`

errcode1202 (approved):

src/test/scripts/product_versions.py
```diff
@@ -45,6 +53,13 @@ def get_lts_version(argv):
 	return lts_version
 
 
+def loadJSON(fdata):
+	result = re.search("\[(.*?)\]", fdata.decode("utf-8"))
```
wkritzinger-atlassian, line 57: This regex will fail if there are any `]` in the objects, the `.*` pattern should probably be greedy here (drop the `?` at the end). This way this pattern will match from the first `[` to the last `]`

src/test/scripts/product_versions.py
```diff
@@ -45,6 +53,13 @@ def get_lts_version(argv):
...
+	if result is None:
+		return ""
+	return json.loads(f"[{result.group(1)}]")
```
wkritzinger-atlassian, line 60: If you use "group 0", it'll use the whole match (and you wouldn't need to use a matching group at all), and you don't need to surround it with `[]`.
```suggestion
	return json.loads(result.group(0))
```

src/test/scripts/product_versions.py
nghazali: 👍

src/test/scripts/product_versions.py
```diff
@@ -45,6 +53,13 @@ def get_lts_version(argv):
 	return lts_version
 
 
+def loadJSON(fdata):
+	result = re.search("\[.*\]", fdata.decode("utf-8"))
+	if result is None:
+		return ""
```
wkritzinger-atlassian, line 59: Since this is returning a list of values, shouldn't it return `[]` in the error case?

nghazali: I tested with invalid data and returns None

wkritzinger-atlassian: `result` is `None` but you specifically return `""` if `result` is `None`. In the code above, you have the line

```
feeds += loadJSON(current_feeds)
```

If `loadJSON` returns `""`, it'll append an empty string to a list, and I think python will coerce the value into a string, which is probably not what you want.

```suggestion
		return []
```

PR #199, opened 2021-07-05
DCKUBE-437: Wording improvement for warning in NOTES when PV is not used

yzha645:
When helm chart is installed without persistent volume, a warning will be printed out in the post-install note. 
However, the warning was for Shared Home only. Now added local home PV checking as well, and made separate warnings for each, which hopefully will reduce ambiguity.

### DIFF
--- a/src/main/charts/bitbucket/templates/NOTES.txt
+++ b/src/main/charts/bitbucket/templates/NOTES.txt
@@ -41,7 +41,7 @@ Get the {{ title .Chart.Name }} URL by running these commands in the same shell:
 
 {{ if not ( or .Values.volumes.localHome.persistentVolumeClaim.create .Values.volumes.localHome.customVolume ) }}
 #################################################################################
-######      WARNING: Persistent volume is not used for Local Home!!!       #####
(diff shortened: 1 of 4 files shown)


tan-ro (approved):

src/main/charts/bitbucket/templates/NOTES.txt
```diff
@@ -39,9 +39,15 @@ Get the {{ title .Chart.Name }} URL by running these commands in the same shell:
 {{- end }}
 {{- end }}
 
+{{ if not ( or .Values.volumes.localHome.persistentVolumeClaim.create .Values.volumes.localHome.customVolume ) }}
+#################################################################################
+######      WARNING: Persistent volume is not used for Local Home!!!       #####
```
errcode1202, line 44: Minor indentation issue here you need one more space so that the hashes `#` are vertically aligned :) This applies to the other `NOTES.txt`

yzha645: Good spot!! Fixed~

nghazali (approved):

PR #202, opened 2021-07-08
DCKUBE-474: Added Jira Proxy information to helm chart

nghazali:
Added default value for ingress.path and also introduced ALT_PROXY_NAME and ALT_PROXY_PORT for Jira helm chart

### DIFF
--- a/src/main/charts/jira/templates/_helpers.tpl
+++ b/src/main/charts/jira/templates/_helpers.tpl
@@ -34,6 +34,24 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
+Create default value for ingress port
+*/}}
+{{- define "jira.ingressPort" -}}
+{{ default (ternary "443" "80" .Values.ingress.https) .Values.ingress.port -}}
+{{- end }}
+
+{{/*
+Create default value for ingress path
+*/}}
+{{- define "jira.ingressPath" -}}
+{{- if .Values.ingress.path -}}
+{{- .Values.ingress.path -}}
+{{- else -}}
+{{ default ( "/" ) .Values.jira.service.contextPath -}}
+{{- end }}
+{{- end }}
+
+{{/*
 The name of the service account to be used.
 If the name is defined in the chart values, then use that,
 else if we're creating a new service account then use the name of the Helm release,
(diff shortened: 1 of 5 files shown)


src/main/charts/jira/values.yaml
```diff
@@ -313,9 +313,9 @@ ingress:
    
   # -- The base path for the Ingress Resource. For example '/jira'. Based on a 
   # 'ingress.host' value of 'company.k8s.com' this would result in a URL of 
-  # 'company.k8s.com/jira'
+  # 'company.k8s.com/jira'. Default value is 'jira.service.contextpath'
   #
-  path: "/"
+  #path: '/'
```
nanux, line 318: I don't think we comment out the values that are defaulted. You can leave it as `path: ` because providing the`path: /` example doesn't do more than the current default. Leaving the key there but without value is more consistent to rest of the keys.

When you change it to `path:`, please make sure it does still work with the `if` and ternary statements.

nanux (changes requested):

nanux:
@nghazali can you please also add a unit test for the default ingress path (that it will pick up context path) and the fact that we fill the environmental variables now?

You can follow `test.IngressTest` example.

errcode1202 (approved):

src/main/charts/jira/values.yaml
nghazali: done

src/main/charts/jira/templates/_helpers.tpl
```diff
@@ -33,6 +33,24 @@ Create chart name and version as used by the chart label.
...
+{{/*
+Create default value for ingress path
+*/}}
+{{- define "jira.ingressPath" -}}
+{{- if .Values.ingress.path -}}
+{{- .Values.ingress.path -}}
+{{- else -}}
+{{ default ( "/" ) .Values.jira.service.contextPath -}}
+{{- end }}
+{{- end }}
```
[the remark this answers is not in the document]
nghazali: The tests are added

errcode1202 (approved):

apawelczyk-atlassian (approved):

nanux (approved):

PR #205, opened 2021-07-19
DCKUBE-436 Bitbucket NFS server chart.

wkritzinger-atlassian:
TODO: 
* Fix docker image that's used for the tag
* Tune some of the variables for the nfs chart


src/test/infrastructure/nfs-server/values.yaml
```diff
@@ -0,0 +1,75 @@
...
+image:
+  repository: docker.atl-paas.net/atlassian/nfs-server-test
+  #pullPolicy: IfNotPresent
+  pullPolicy: Always
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "wkritzinger"
+
+imagePullSecrets:
+  - name: regcred
+nameOverride: ""
```
nanux, line 16: I believe you can leave out `""` for empty values.

tarka (approved):

nanux (approved):

src/test/infrastructure/nfs-server/values.yaml
```diff
@@ -0,0 +1,75 @@
...
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: docker.atl-paas.net/atlassian/nfs-server-test
+  #pullPolicy: IfNotPresent
+  pullPolicy: Always
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "wkritzinger"
```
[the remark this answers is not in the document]
wkritzinger-atlassian: Yep once the Docker image was merged and pushed I'll update this

src/test/infrastructure/nfs-server/values.yaml
wkritzinger-atlassian: This came from the helm chart template, as did most of these values in this file so I might leave it as-is

pom.xml
```diff
@@ -118,6 +118,14 @@
 
         <persistence.enabled>true</persistence.enabled>
         <clustering.enabled>true</clustering.enabled>
+        <!--
+            Number of instance replicas to start during initial setup. This value is currently only used by Bitbucket,
+            other products' values.yaml ignore this property. Other products need to perform additional setup steps on
+            a single node after Helm chart installation.
+        -->
+        <replica.count>2</replica.count>
```
wkritzinger-atlassian, line 126: I made this change so the setup for Bitbucket becomes easier. It does not scale up the instance after deployment anymore and instead starts up with 2 nodes. All other products still work the same way, this variable is only used for the Bitbucket templates, as is mentioned in the comment above.

src/test/config/bitbucket/values.yaml
```diff
@@ -1,5 +1,7 @@
 # This file contains overrides for the Bitbucket Helm chart's values.yaml file
 
+replicaCount: ${replica.count}
```
wkritzinger-atlassian, line 3: This is how this value is only applied to Bitbucket, this line is missing in the other products.

wkritzinger-atlassian:
Fixed the TODOs and got green builds using this branch. Need a new release before we get green builds again for Bitbucket.

PR #206, opened 2021-07-20
DCKUBE-475: Initial troubleshooting document 

nanux:
Ideally, we would start collecting examples of common failures and problems in a form of:

* Symptoms
* Investigation
* Potential fixes

But for now, I don't have a concrete list to put here. I am open to suggestions.

### DIFF
--- a/docs/docs/README.md
+++ b/docs/docs/README.md
@@ -36,6 +36,7 @@ The diagram below provides a high level overview of what a typical deployment mi
 * [Configuration](installation/CONFIGURATION.md) - a deep dive into the configuration parameters
 * [Platforms support](platforms/PLATFORMS.md) - how to allow support for different platforms
 * [Examples](examples/EXAMPLES.md) - various configuration examples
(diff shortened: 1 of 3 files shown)


errcode1202:
I wonder if we should caveat the guide itself stating that we cannot provide guidance on anything outside of the product install itself and its configuration. Anything else i.e general k8s issues is out of scope.

docs/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+```shell
+kubectl get RELEASE_NAME -n NAMESPACE -o wide
+```
+
+Get details about a specific pod:
+
+```shell
+kubectl describe POD_NAME -n NAMESPACE
+```
```
errcode1202, line 67: Should we add some steps on examining the output for the storage (local/shared home) i.e. `describe` for `PV` / `PVC` . A lot of installation issues come from misconfigured storage and looking at the pod is not enough to understand why the pod wont start

tan-ro (approved):

PR #208, opened 2021-07-21
Add updated architectural diagram via upload

tan-ro:


tarka:
Clipped text ...

tarka:
Separate from this update, the relationship between e.g the PV/PVCs and EFS/shared-FS haven't been shown, but I suppose that's not really important at this point?

tan-ro:
Fixed clipped text

tarka (approved):

nanux:
Thank you @tan-ro. I wanted to make a comment that the new diagram doesn't have an alpha channel and the background is not transparent but I've checked the current diagram with dark theme (I am using white by default).

Screen Shot 2021-07-23 at 10 37 07 am
Screen Shot 2021-07-23 at 10 37 15 am

I believe having a white background is definitely a better option.

nanux:
Would it be possible to also add the updated xml source file? https://github.com/atlassian-labs/data-center-helm-charts/blob/master/docs/images/k8s%20overview.xml

tan-ro:
Uploaded k8s_overview.xml

nanux (approved):

errcode1202 (approved):

PR #210, opened 2021-07-22
DCKUBE-527: Fix Synchrony resource requests and limits

nanux:
* added resources into Synchrony container
* added tests
* enabled synchrony by default for the expected outcome test

### DIFF
--- a/src/main/charts/confluence/templates/statefulset-synchrony.yaml
+++ b/src/main/charts/confluence/templates/statefulset-synchrony.yaml
@@ -49,6 +49,10 @@ spec:
             initialDelaySeconds: {{ .Values.synchrony.readinessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.synchrony.readinessProbe.periodSeconds }}
             failureThreshold: {{ .Values.synchrony.readinessProbe.failureThreshold }}
+          {{- with .Values.synchrony.resources.container }}
+          resources:
+          {{- toYaml . | nindent 12 }}
+          {{- end }}
           env:
             - name: SYNCHRONY_BIND
               valueFrom:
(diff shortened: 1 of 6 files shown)


src/test/java/test/SynchronyTest.java
```diff
@@ -39,6 +40,19 @@ void synchrony_enable(Product product) throws Exception {
...
+    @EnumSource(value = Product.class, names = "confluence")
+    void synchrony_disabled(Product product) throws Exception {
+        final var resources = helm.captureKubeResourcesFromHelmChart(product, Map.of(
+                "synchrony.enabled", "false"
+        ));
+
+        final var sysProps = resources.get(Kind.ConfigMap, product.getHelmReleaseName() + "-jvm-config")
+                .getNode("data", "additional_jvm_args");
+
+        assertThat(sysProps).hasTextContaining("synchrony.btf.disabled");
```
nanux, line 53: Because Synchrony is now disabled by default for the comparison test, I've added this test to verify that the property is added if we disable synchrony.

src/main/charts/confluence/templates/statefulset-synchrony.yaml
```diff
@@ -49,6 +49,10 @@ spec:
             initialDelaySeconds: {{ .Values.synchrony.readinessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.synchrony.readinessProbe.periodSeconds }}
             failureThreshold: {{ .Values.synchrony.readinessProbe.failureThreshold }}
+          {{- with .Values.synchrony.resources.container }}
```
wkritzinger-atlassian, line 52: So the values for this already existed, but they were not put into the statefulset?

nanux: Correct, they were just not utilised.

badgersow (approved):

tarka (approved):

PR #212, opened 2021-07-23
DCKUBE-529: Update product versions to latest

nanux:
Update product versions

### DIFF
--- a/src/main/charts/bitbucket/README.md
+++ b/src/main/charts/bitbucket/README.md
@@ -1,6 +1,6 @@
 # bitbucket
 
-![Version: 0.13.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 7.12.1-jdk11](https://img.shields.io/badge/AppVersion-7.12.1--jdk11-informational?style=flat-square)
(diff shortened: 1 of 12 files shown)


nanux:
Waiting for the [build results](https://server-syd-bamboo.internal.atlassian.com/browse/DCNG-K8SHELMTEST255)

src/main/charts/bitbucket/README.md
```diff
@@ -1,6 +1,6 @@
 # bitbucket
 
-![Version: 0.13.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 7.12.1-jdk11](https://img.shields.io/badge/AppVersion-7.12.1--jdk11-informational?style=flat-square)
+![Version: 0.13.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 7.14.1-jdk11](https://img.shields.io/badge/AppVersion-7.14.1--jdk11-informational?style=flat-square)
```
wkritzinger-atlassian, line 3: Are we updating the 0.13.0 version?

badgersow (approved):

src/main/charts/bitbucket/README.md
nanux: There will be a new release created after this. I can see what you mean, the README.md is not updated until there is a new version released and then it will sow `0.14.0`. I don't want to couple this PR with solving that problem as it was happening before as well and I am planning to run the release today but it is something we should consider.

tarka (approved):

src/main/charts/bitbucket/README.md
wkritzinger-atlassian: Ahhh I see now, yep that makes sense 👍

wkritzinger-atlassian (approved):

PR #214, opened 2021-07-23
DCKUBE-525: Externalise the Elasticsearch cluster for func-tests

tarka:
This also contains some tweaks to make the script a bit more robust for future refactoring.


src/test/scripts/elasticsearch-values.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
nanux, line 1: I think this could be moved to `/src/test/infrastructure` so it is next to `nfs-server` helm chart. Arguably we should move the `postgres-values.yaml` there as well.

src/test/scripts/elasticsearch-values.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+master:
+  replicas: 1
+  podAnnotations:
+    "atlassian.com/business_unit": "server_engineering"
```
nanux, line 6: Is this necessary in our cluster?

nanux (approved):

src/test/scripts/helm_install.sh
```diff
@@ -123,17 +127,33 @@ bootstrap_database() {
...
-     bitnami/postgresql > $LOG_DOWNLOAD_DIR/helm_install_log.txt
+     bitnami/postgresql >> $LOG_DOWNLOAD_DIR/helm_install_log.txt
 
   if [[ "$DB_INIT_SCRIPT_FILE" ]]; then
     kubectl cp -n "${TARGET_NAMESPACE}" $DB_INIT_SCRIPT_FILE $POSTGRES_RELEASE_NAME-0:/tmp/db-init-script.sql
     kubectl exec -n "${TARGET_NAMESPACE}" ${POSTGRES_RELEASE_NAME}-0 -- /bin/bash -c "psql postgresql://$PRODUCT_NAME:$PRODUCT_NAME@localhost:5432/$DB_NAME -f /tmp/db-init-script.sql"
   fi
 }
 
+bootstrap_elasticsearch() {
```
errcode1202, line 138: Out of interest, are we using basic auth for ES? If so where are the credentials defined?

src/test/scripts/elasticsearch-values.yaml
tarka: Only for running in KITT. It's probably not necessary in day-to-day, or for CI, but there's no harm in adding it for future use.

src/test/scripts/helm_install.sh
tarka: We don't use auth or HTTPS (yet). The open-source version of ES doesn't support it, and the ways of hacking it in are not ideal. I wrote up some notes here: https://hello.atlassian.net/wiki/spaces/DCD/pages/1252068866/Notes+Bitbucket+Elasticsearch+recommendations

Medium term it is _very_ likely we will support OpenSearch for external ES (as AWS will be moving to it), so we can migrate to that then.

PR #217, opened 2021-07-28
DCKUBE-534: Expand post install tests

tarka:


src/test/java/test/postinstall/ElasticSearchInstallTest.java
```diff
@@ -0,0 +1,64 @@
...
+@EnabledIf("isBitbucket")
+class ElasticSearchInstallTest {
+    static boolean isBitbucket() {
+        return productIs(Product.bitbucket);
+    }
+
+    private static final KubeClient client = new KubeClient();
+
+    // See helm_install.sh for where this host is generated.
+    static final String esIngressBase = "https://" + getRelease() + "-elasticsearch-master-0.kube.deplops.com";
```
wkritzinger-atlassian, line 36: `kube.deplops.com` is that too hardcoded? Other K8s clusters will have different domains

tarka: Good point. I'll make it configurable.

wkritzinger-atlassian: Can you use the value from the `helm_parameters` file?

tarka: Yep, that's what I've done.

src/test/java/test/postinstall/ElasticSearchInstallTest.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Test
+    void elasticSearchBeingUsed() {
+        // Relies on the backdoor ingress controller installed by helm_install.sh.
+        // If this changes an alternative would be to use the fabric8 client ExecWatch/ExecListener to
+        // invoke curl from a pod.
+        final var indexURL = esIngressBase + "/_cat/indices?format=json";
+        when().get(indexURL).then()
+                .body("findAll { it.index == 'bitbucket-index-version' }[0]", hasEntry("docs.count", "1"));
```
wkritzinger-atlassian, line 58: I wonder if we need to retry this as Bitbucket can take a while to create those indices. In practice it probably won't be a problem as `helm install` waits for the pods to start up. We'll need to keep an eye on it and see if it's flaky.

wkritzinger-atlassian (approved):
Great changes! Will be good to have this tested during CI

nghazali (approved):

PR #218, opened 2021-07-28
DCKUBE-453: Add support for providing a custom fluentd start command

errcode1202:


src/main/charts/jira/templates/_fluentd_templates.tpl
```diff
@@ -2,7 +2,7 @@
 {{ if .Values.fluentd.enabled }}
 - name: fluentd
   image: {{ .Values.fluentd.imageName }}
-  command: ["fluentd", "-c", "/fluentd/etc/fluent.conf", "-v"]
+  command: ["sh", "-c", {{ include "fluentd.start.command" . | quote }}]
```
bianchi2, line 5: @errcode1202 this way fluentd process won't have pid 1 and when the pod is shut down it's sh that will receive SIGTERM or SIGKILL (if not shut down gracefully). Chances are that child processes spawned by fluentd won't be terminated as fluentd isn't the main container process, as a result some zombie processes may show up in workers.

I do not claim it will necessarily happen, but sh -c approach isn't among best practices. As far as I understand, you want to let users run any arbitrary command which you wrap in sh -c?

errcode1202: @bianchi2 Yes the intention is to allow users to supply an arbitrary command here. Is there a better way of doing this as opposed the `"sh", "-c"` approach?

errcode1202: @bianchi2 the approach used here was based on how custom commands are defined for the [nfs-permission-fixer](https://github.com/atlassian-labs/data-center-helm-charts/blob/master/src/main/charts/jira/templates/statefulset.yaml#L44). The explanation you provided makes a good point though, and I wonder if we need to re-assess our approach to how these `custom` commands are being executed and whether they are safe within the context that they are used. Let me have a chat with @tarka today to see what the recommendations on this are.

tarka: Yep, nested shells in startup can cause the shutdown signal to be lost, which will result in a non-clean process termination. However you can avoid this by doing `exec <my-sub-process>` at key points. We ship modified startup scripts for Bitbucket that do this for this reason. So you _might_ get away with `["sh", "-c", "exec {{ include "fluentd.start.command" . | quote }}"]`, or simply document the issue.

More details about the many ways of getting this wrong here: https://hynek.me/articles/docker-signals/

bianchi2: thanks @tarka. `exec "$@"` i what I usually add to entrypoint scripts. In this particular case, indeed exec in the command itself should work (entrypoint in docker terms, while args is cmd). I have created a simple pod this way:

```
kind: Pod
apiVersion: v1
metadata:
  name: tomcat
spec:
  containers:
    - name: tomcat
      image: tomcat
      command:
        - sh
        - -c 
        - exec /usr/local/tomcat/bin/catalina.sh run
```
And Tomcat java process has PID 1:

```
    PID TTY      STAT   TIME COMMAND
      1 ?        Ssl    0:01 /usr/local/openjdk-11/bin/java -Djava.util.logging.config.file=/usr/local/tomcat/conf/logging.properties -Djava.util.logging.manager=org.apache.juli.ClassLoaderLogManager -Djdk.tls.ephemeralDHKeySize=2048 -Djava.protocol.handler.pkgs=org.apache.catalina.webresources -Dorg.apache.catalina.security.SecurityListener.UMASK=0027 -Dignore.endorsed.dirs= -classpath /usr/local/tomcat/bin/bootstrap.jar:/usr/local/tomcat/bin/tomcat-juli.jar -Dcatalina.base=/usr/local/tomcat -Dcatalina.home=/usr/local/tomcat -Djava.io.tmpdir=/usr/local/tomcat/temp org.apache.catalina.startup.Bootstrap start
     37 pts/0    Ss     0:00 sh -i -c TERM=xterm sh
     43 pts/0    S      0:00 sh
     44 pts/0    S      0:00 bash
     47 pts/0    R+     0:00 ps ax
```

errcode1202: Thanks guys for the great explanations! Let me modify the changes to use `exec` and test this approach. Ill also update the `nfs-permission-fixer` to make use of `exec` too.

gjacobsatl (approved):

src/main/charts/jira/templates/_fluentd_templates.tpl
errcode1202: Discussing and experimenting with @tarka, we think the best way to support custom commands, and also ensure signals (`SIGTERM`) are received as appropriate, is to make sure that the custom command is crafted appropriately. That is, the command should invoke the _**primary**_ process with `exec`. As part of this change the commentary for the `fluentd.command` property in `values.yaml` (see change) has been updated to call out this detail i.e.

```bash
# -- The command used to start Fluentd. If not supplied the default command 
# will be used: "fluentd -c /fluentd/etc/fluent.conf -v"
#
# Note: The custom command can be free-form, however pay particular attention to
# the process that should ultimately be left running in the container. This process
# should be invoked with 'exec' so that signals are appropriately propagated to it,
# for instance SIGTERM. An example of how such a command may look is:
# "<command 1> && <command 2> && exec <primary command>"
```

If we deploy Jira with the following:

```yaml
fluentd:
  enabled: true
  imageName: fluent/fluentd-kubernetes-daemonset:v1.11.5-debian-elasticsearch7-1.2
  command: "echo Hi there! && apt update && apt install curl -y && curl https://jira-perf.k8s.deplops.com/status && exec fluentd -c /fluentd/etc/fluent.conf -v"
```

each command is executed successfully with the final fluentd process being executed as `PID 1`...

```bash
root@jira-0:/home/fluent# ps -eaf
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  1 01:22 ?        00:00:00 ruby /fluentd/vendor/bundle/ruby/2.6.0/bin/fluentd -c /fluentd/etc/fluent.conf -v
root       433     1  1 01:22 ?        00:00:00 /usr/local/bin/ruby -Eascii-8bit:ascii-8bit /fluentd/vendor/bundle/ruby/2.6.0/bin/fluentd -c /fluentd/etc/fluent.conf -v --under-supervisor
root       434     0  0 01:22 pts/0    00:00:00 /bin/bash
root       444   434  0 01:22 pts/0    00:00:00 ps -eaf
```

@bianchi2 it would be great to get your take on this.

tarka: :+1:

tarka (approved):

src/main/charts/jira/templates/_fluentd_templates.tpl
bianchi2: LGTM! Thanks for making these changes.

PR #237, opened 2021-08-12
DCKUBE-593: use default storage class of the NFS server

badgersow:
Local execution has been broken since this change. It was trying to provision an NFS Server using the gp2 Storage Class, which is available in AWS, but not in Docker for Mac.

This pull request uses the default Storage Class for NFS Server instead. It fixes the local execution, and it's also supposed to work in AWS, since the gp2 storage class is default there.

* [Confluence build](https://server-syd-bamboo.internal.atlassian.com/browse/CONFA3CLUSTER-K8S37-2)
* [Bitbucket build](https://server-syd-bamboo.internal.atlassian.com/browse/BSERV-K8S17-4)

### DIFF
--- a/src/test/infrastructure/nfs-server/templates/statefulset.yaml
+++ b/src/test/infrastructure/nfs-server/templates/statefulset.yaml
@@ -125,7 +125,7 @@ spec:
         {{- end }}
       spec:
         accessModes: [ "ReadWriteOnce" ]
-        {{- if .Values.persistence.storageClassName }}
+        {{- if not (kindIs "invalid" .Values.persistence.storageClassName) }}
         storageClassName: {{ .Values.persistence.storageClassName | quote }}
         {{- end }}
         resources:
--- a/src/test/infrastructure/nfs-server/templates/statefulset.yaml
+++ b/src/test/infrastructure/nfs-server/templates/statefulset.yaml
@@ -125,7 +125,7 @@ spec:
         {{- end }}
       spec:
         accessModes: [ "ReadWriteOnce" ]
-        storageClassName: {{ .Values.persistence.storageClassName | default "gp2" }}
+        storageClassName: {{ .Values.persistence.storageClassName }}
         resources:
           requests:
             storage: {{ .Values.persistence.size | default "1Gi" }}
--- a/src/test/infrastructure/nfs-server/templates/statefulset.yaml
+++ b/src/test/infrastructure/nfs-server/templates/statefulset.yaml
@@ -125,7 +125,9 @@ spec:
         {{- end }}
       spec:
         accessModes: [ "ReadWriteOnce" ]
-        storageClassName: {{ .Values.persistence.storageClassName }}
+        {{- if .Values.persistence.storageClassName }}
+        storageClassName: {{ .Values.persistence.storageClassName | quote }}
+        {{- end }}
         resources:
           requests:
             storage: {{ .Values.persistence.size | default "1Gi" }}
--- a/src/test/infrastructure/nfs-server/values.yaml
+++ b/src/test/infrastructure/nfs-server/values.yaml
@@ -57,7 +57,7 @@ persistence:
   annotations: {}
   size: 5Gi
   # If set to nonempty string value, this will specify the storage class to be used.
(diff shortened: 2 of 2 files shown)


badgersow:
@wkritzinger-atlassian Please take a look at this PR. I wonder if it can break something in Bitbucket.

errcode1202:
@wkritzinger-atlassian this reminds me of a similar problem we encountered during the blitz, where we had to set the `StorageClassName` to `""` (see below for `values.yaml` snippet`) for things to work otherwise we got a storage class conflict error. Im not sure now which is the better approach, defaulting to `gp2` or setting it blank `""`

```yaml
sharedHome:
    persistentVolume:
      create: true
      mountOptions: []
      nfs:
        server: "10.100.31.122"
        path: "/srv/nfs" 
    persistentVolumeClaim:
      create: true
      storageClassName: ""
```

src/test/infrastructure/nfs-server/templates/statefulset.yaml
```diff
@@ -125,7 +125,7 @@ spec:
         {{- end }}
       spec:
         accessModes: [ "ReadWriteOnce" ]
-        storageClassName: {{ .Values.persistence.storageClassName | default "gp2" }}
+        storageClassName: {{ .Values.persistence.storageClassName }}
```
nanux, line 128: What about not including the `storageClassName`property in case the `.Values.peristance.storageClassName` is empty? I assume it should then use the default storage class:

> If storageClassName is not specified in the PVC, the default storage class will be used for provisioning.

https://kubernetes.io/blog/2017/03/dynamic-provisioning-and-storage-classes-kubernetes/

badgersow: Yes, I think this is a better idea. To be honest, I wonder why it works now, because the empty string is supposed to disable the dynamic provisioning, according to the documentation.

pbruski: Maybe you want to do something like this:
`  {{- if not (kindIs "invalid" .Values.volumes.sharedHome.persistentVolumeClaim.storageClassName) }}
  storageClassName: {{ .Values.volumes.sharedHome.persistentVolumeClaim.storageClassName | quote }}
  {{- end }}`
(from src/main/charts/bitbucket/templates/shared-home-pvc.yaml)

badgersow: Thanks for the tip. I've done this and added a comment to explain the usage.

badgersow:
@errcode1202 Thanks for your comment. I've changed the PR to not include the SC clause by default instead of using the empty string.

src/test/infrastructure/nfs-server/values.yaml
```diff
@@ -56,7 +56,10 @@ service:
 persistence:
   annotations: {}
   size: 5Gi
-  storageClassName: "gp2"
+  # If set to nonempty string value, this will specify the storage class to be used.
```
errcode1202, line 59: minor `nonempty` should probably be `non-empty`

errcode1202 (approved):

src/test/infrastructure/nfs-server/values.yaml
```diff
@@ -56,7 +56,10 @@ service:
...
+  # If left without value, the default Storage Class will be utilised.
```
errcode1202, line 60: Can we stipulate what the default storage class is i.e. `gp2`

src/test/infrastructure/nfs-server/values.yaml
```diff
@@ -56,7 +56,10 @@ service:
...
+  # Alternatively, can be set to the empty string, to indicate that no Storage Class should be used here.
```
errcode1202, line 61: can we also indicate what we mean by empty string i.e. `""`

src/test/infrastructure/nfs-server/values.yaml
nanux: This probably depends on the cluster, I am not sure if we assume a default class here.

nanux (approved):

PR #249, opened 2021-08-19
DCKUBE-613: Configurable grace periods

tarka:
I couldn't find the graceful periods for Crowd and Jira, so went with the Kubernetes default.


errcode1202 (approved):

errcode1202 (changes requested):
Hey Steve, you need to regenerate the docs so that the new property `terminationGracePeriodSeconds` is rendered as part of them, see: https://github.com/atlassian-labs/data-center-helm-charts/blob/1e8b150023ab419e0ac8d11be3c9babbc50f45d2/DEVELOPMENT.md

src/main/charts/bitbucket/values.yaml
```diff
@@ -35,6 +35,11 @@
 #
 replicaCount: 1
 
+# -- The termination grace period for pods during shutdown. This
+# should be set the Bitbucket internal grace period (default 30
```
errcode1202, line 39: Not sure if this sentence reads correctly: `should be set the Bitbucket internal grace period ` should it be `should will set the Bitbucket internal grace period ` ? 

This needs to be fixed for the other products too.

src/main/charts/bitbucket/templates/statefulset.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         {{- include "bitbucket.selectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "bitbucket.serviceAccountName" . }}
-      terminationGracePeriodSeconds: 1
+      terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
```
errcode1202, line 28: Should we add tests to ensure that the correct grace period (whats defined in `values.yaml`) is applied?

tarka: This is done via the expected-output tests.

src/main/charts/bitbucket/values.yaml
tarka: Bitbucket has it's own internal grace-period for SSH etc, which the customer can override. This setting is for the Kubernetes grace setting, which is separate. The Kubernetes one should be slightly higher than the BB one, so that BB can complete it's shutdown before the Kubernetes grace period times out.

But there is a missing word, I'll fix that.

nghazali (approved):

src/main/charts/confluence/templates/statefulset-synchrony.yaml
```diff
@@ -21,7 +21,7 @@ spec:
         {{- include "synchrony.selectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "confluence.serviceAccountName" . }}
-      terminationGracePeriodSeconds: 1
+      terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
```
badgersow, line 24: Just to let you know - I'm working on a similar PR in #250 and I will likely move this property to different value (because I also want to add shutdown command). I will wait until you merge and then start on top of your changes.

badgersow (approved):

src/main/charts/confluence/templates/statefulset-synchrony.yaml
tarka: :+1:

errcode1202 (approved):

src/main/charts/bitbucket/README.md
```diff
@@ -28,23 +28,29 @@ Kubernetes: `>=1.19.x-0`
...
 | bitbucket.ports.hazelcast | int | `5701` | The port on which the Hazelcast listens for client traffic |
 | bitbucket.ports.http | int | `7990` | The port on which the Bitbucket container listens for HTTP traffic |
 | bitbucket.ports.ssh | int | `7999` | The port on which the Bitbucket container listens for SSH traffic |
+| bitbucket.resources.container | object | `{"requests":{"cpu":"2","memory":"2G"}}` |  container. It is important that if the memory resources are specified here,  they must allow for the size of the Bitbucket JVM. That means the maximum heap  size, the reserved code cache size, plus other JVM overheads, must be  accommodated. Allowing for (maxHeap+codeCache)*1.5 would be an example. |
 | bitbucket.resources.container.requests.cpu | string | `"2"` | Initial CPU request by Bitbucket pod |
 | bitbucket.resources.container.requests.memory | string | `"2G"` | Initial Memory request by Bitbucket pod |
+| bitbucket.resources.jvm | object | `{"maxHeap":"1g","minHeap":"512m"}` |  defaults defined for the Bitbucket docker container. https://bitbucket.org/atlassian-docker/docker-atlassian-bitbucket-server/src/master/ |
```
badgersow, line 50: Any ideas how so much additional `object` documentation lines were created? My pre-commit hook doesn't generate them.

PR #250, opened 2021-08-19
DCKUBE-612: proper shutdown for Confluence

badgersow:
In this pull request we introduce proper shutdown procedure for Confluence. This is going to remove race conditions during shutdown process, which could lead to errors in logs and potentially incorrect shutting down procedure.

The changes can be summarized as:

1. Use tomcat stop script instead of sending SIGTERM to the JVM process
2. Wait until the tomcat process is killed


nghazali:
DCKUBE-613 covers this change so I think we can close this ticket with no change after DCKUBE-613 is merged.

badgersow:
> DCKUBE-613

I'll wait when Steve merges the PR and then consolidate our changes

src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -178,6 +178,15 @@ The command that should be run by the nfs-fixer init container to correct the pe
 {{- end }}
 {{- end }}
 
+{{/*
+The command that will be run to properly shutdown Confluence tomcat
+*/}}
+{{- define "confluence.shutdown.command" -}}
+{{- if .Values.confluence.shutdown.command }}
+{{ .Values.confluence.shutdown.command }}
+{{- else }}export CONNIE_PID=$(pgrep -f org.apache.catalina.startup.Bootstrap | sort | head -n 1); echo $CONNIE_PID > /opt/atlassian/confluence/work/catalina.pid && /bin/su confluence -c /opt/atlassian/confluence/bin/stop-confluence.sh; tail --pid=$CONNIE_PID -f /dev/null{{- end }}
```
badgersow, line 187: This is the command to shutdown Confluence properly

1. Get the PID of Confluence (the PID of the only running tomcat process)
2. Write it to the file and launch the tomcat stop script
3. Wait until the process has been killed

tarka: This can be simplified, there is a script /opt/atlassian/support/common.sh that will do the pgrep part and populate $APP__PID.

That said, I'm going to modify the entrypoint util to dump a pid in the app home, so we can use that from now on. And I'm also going to add a dedicated shutdown util to each image.

I'd say merge this for now, and I'll port this across to the Docker image when I do that.

tarka (approved):

src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -185,7 +185,7 @@ The command that will be run to properly shutdown Confluence tomcat
 {{- if .Values.confluence.shutdown.command }}
 {{ .Values.confluence.shutdown.command }}
 {{- else }}
-{{- printf "echo $(pgrep -f org.apache.catalina.startup.Bootstrap | sort | head -n 1) > /opt/atlassian/confluence/work/catalina.pid && /bin/su confluence -c /opt/atlassian/confluence/bin/stop-confluence.sh" }}
+echo $(pgrep -f org.apache.catalina.startup.Bootstrap | sort | head -n 1) > /opt/atlassian/confluence/work/catalina.pid && /bin/su confluence -c /opt/atlassian/confluence/bin/stop-confluence.sh
```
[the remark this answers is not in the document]
badgersow: It's not waiting and this is the reason why I've had to add the `tail` command in the end

nghazali (approved):

PR #255, opened 2021-08-22
DCKUBE-580 Add some content for DEVELOPMENT.md document

nghazali:


DEVELOPMENT.md
```diff
@@ -14,6 +47,45 @@ Repository contains repository hook definition to auto-generate chart documentat
 When you have the pre-commit hook installed and `commit` a file, it will run the pre-commit hook. If there is a change generated in the `README.md`, the hook will fail and notify you about the new change. You can review the changes and run the `commit` again, now it should pass.
 
 
+## Build and test the project
```
errcode1202, line 50: Im really not sure if we should be advertising this stuff. `helm_install.sh` is something **we use only for CI**, and isnt used for "development" of the charts per se. It's caused problems/confusion in the past and Ive stopped using it myself. To be honest if a turn key solution is what we're trying to provide here then we should wait for the official Terraform scripts.

DEVELOPMENT.md
```diff
@@ -1,5 +1,38 @@
...
+## Prerequisites 
+Before deploy and setup Atlassian's Data Center products the following resources are required:
+* A Kubernetes cluster
+* An Ingress Controller
+* A database
+* Have a `shared-home` volume
+* Setup `local-home` volume
+
+To see how to create and setup those resources follow [Prerequisites Documentation](https://atlassian-labs.github.io/data-center-helm-charts/userguide/PREREQUISITES/).
+
```
errcode1202, line 35: A lot of this content feels like its duplicated from the PREREQUISITE doc Would it not be better to add this detail to it and then reference it from here?

errcode1202 (changes requested):

DEVELOPMENT.md
```diff
@@ -14,6 +47,45 @@ Repository contains repository hook definition to auto-generate chart documentat
...
+```shell script
+./src/test/scripts/helm_install.sh target/config/[product]/helm_parameters
+```
+
+To uninstall the test product and cleanup the resources run `helm_uninstall.sh`:
+
+```shell script
+./src/test/scripts/helm_uninstall.sh target/config/[product]/helm_parameters
+```
+
```
errcode1202, line 88: Im really not sure if we should be advertising this stuff. helm_install.sh is something we use only for CI, and isnt used for "development" of the charts per se. It's caused problems/confusion in the past and Ive stopped using it myself. To be honest if a turn key solution is what we're trying to provide here then we should wait for the official Terraform scripts.

errcode1202 (changes requested):

nghazali:
As result of the team discussion I am closing this PR.

PR #261, opened 2021-08-24
Dckube 581 provide value for set permissions

yzha645:
Adding `setPermissions` to values.yaml file gives user flexibility to set environment variable `SET_PERMISSIONS` on container level. 
- Confluence: already had this field. Changed the default value to `true` to keep consistent with the default value of Docker container. 
- Crowd: was hardcoding `SET_PERMISSIONS` to false in statefulset.yaml. Added `setPermissions` to values.yaml and referenced it in statefulset.yaml. 
- Jira & Bitbucket: was not able to configure this value. Added `setPermissions` to values.yaml and referenced it in statefulset.yaml.


src/main/charts/bitbucket/values.yaml
```diff
@@ -451,7 +451,12 @@ bitbucket:
     # -- The GID used by the Bitbucket docker image
     #
     gid: "2003"
-    
+
+  # -- Boolean to define whether to set home directory permissions on startup
```
nanux, line 455: I think we can add `local` to the description => `to set local home directory`
This will make it more explicit and avoid potential confusion between local or shared home directory.

src/main/charts/bitbucket/values.yaml
```diff
@@ -451,7 +451,12 @@ bitbucket:
...
+  # of Bitbucket container. Set to 'false' to disable this behaviour.
+  #
+  setPermissions: true
```
nanux, line 458: I was considering whether `setLocalHomePermissions` won't be more explicit but I am not sure if that is not just unnecessary complicating things. It just feels that `setPermissions` can mean anything but you can read the comment to understand it so I guess that should be fine.

nanux:
You will need to fix the expected output test to fix the build. You can run the test locally with `mvn clean test`

I think it is also worth adding unit test as well for the `SET_PERMISSIONS` env variable - you can use https://github.com/atlassian-labs/data-center-helm-charts/blob/main/src/test/java/test/ContextPathTest.java as an inspiration.

nanux (changes requested):

src/main/charts/bitbucket/values.yaml
yzha645: yea, I agree with you on making it clear on what this variable does. In terms of the naming, I feel there is also some benefits to keep the name consistent (despite the uppercase /camel case thing), especially for developers. So more inclined to the current version.

src/test/java/test/ContextPathTest.java
```diff
@@ -14,7 +14,7 @@
 import static test.jackson.JsonNodeAssert.assertThat;
 
 /**
- * Tests the various permutations of the "<product>.license" value structure in the Helm charts
+ * Tests the various permutations of the "<product>.contextPath" value structure in the Helm charts
```
nanux, line 17: 🥇

nanux:
https://server-syd-bamboo.internal.atlassian.com/browse/DCNG-K8SHELMTEST300-2 Looks happy as well :)

nanux (approved):

nghazali (approved):

PR #263, opened 2021-08-24
DCKUBE-620: Added ContextPath to NOTES.txt and also involved context path to Confluence ingress

nghazali:


src/main/charts/confluence/templates/ingress-setup.yaml
```diff
@@ -29,14 +29,14 @@ spec:
     - host: {{ .Values.ingress.host }}
       http:
         paths:
-          - path: "{{ trimSuffix "/" .Values.ingress.path }}/setup"
+          - path: "{{ include "confluence.ingressPath" . }}/setup"
```
badgersow, line 32: If `Values.ingress.path` and `.Values.confluence.service.contextPath`, this will result to `//setup` URL and the ingress will be broken

src/test/java/test/IngressTest.java
```diff
@@ -210,4 +210,56 @@ void jira_ingress_path_default(Product product) throws Exception {
...
+        final var resources = helm.captureKubeResourcesFromHelmChart(product, Map.of(
+                "ingress.create", "true",
+                "ingress.host", "myhost.mydomain"));
+
+        final var ingresses = resources.getAll(Kind.Ingress);
+        Assertions.assertNotEquals(0, ingresses.size());
+
+        for (KubeResource ingress : ingresses) {
+            assertThat(ingress.getNode("spec", "rules").required(0).path("http").path("paths").required(0).path("path"))
+                    .hasTextContaining("/");
```
badgersow, line 260: We should use `hasTextEqualTo` instead of `hasTextContaining`. For example: `"/foo/".hasTextContaining("/")` is going to pass, and this is not what we want.

src/test/java/test/IngressTest.java
```diff
@@ -210,4 +210,56 @@ void jira_ingress_path_default(Product product) throws Exception {
         }
     }
 
+    @ParameterizedTest
+    @EnumSource(value = Product.class, names = "confluence")
+    void confluence_ingress_path_contextPath(Product product) throws Exception {
```
badgersow, line 215: We should add one more unit test where `confluence.service.contextPath` is not provided, but `ingress.path` is provided. All 4 combination should be tested.

badgersow (changes requested):

src/main/charts/confluence/templates/ingress-setup.yaml
badgersow: Actually it wound be awesome to have the unit test which is going to assert the exact path of Confluence ingresses. This way we will never be able to have something like `//setup` again

nghazali: wow, we see at the same time, I just uploaded a patch to fix this

src/test/java/test/IngressTest.java
nghazali: I think this test is not really needed as based on the logic always path has more priority and if a test is existed and passed with context and path, the result with another test with path and no context would be identical. (if path is existed the workflow does not consider context at all, while the value is provided or not)

src/test/java/test/IngressTest.java
nghazali: You are right, but the reason for using `hasTextContains` in this case and all other tests is we have more than one Node. One is `<path>` and another is `<path>/setup` (consider this assertion is in the loop)
I have changed the code slightly to use `equal` assertion and in the same time don't violate the test conditions in loop by dropping the `/setup` from the path if any is existed.

src/test/java/test/IngressTest.java
nghazali: Anyhow, having an extra test doesn't hurt - so I added an extra test for that scenario.

PR #267, opened 2021-08-25
Allow specifying custom ingress-nginx class name.

Tommyf:
Sometimes clusters have multiple ingress-nginx controllers with
different names, or they simply do not use the default name of
'nginx'.

This change allows the class name to be specified while still using
the built-in Helm Chart functionality of 'nginx: true' instead of
having to redefine all the annotations included with 'nginx:true'


errcode1202 (changes requested):
@Tommyf please see comments in-line.

nanux:
@Tommyf, I will close this one as it became stale. It is now replaced with https://github.com/atlassian/data-center-helm-charts/pull/313. Please let me know if you have any feedback and if it is solving your use case.

PR #281, opened 2021-08-30
Issue/dckube 670 fix synchrony ingress path

badgersow:
Synchrony path of the Ingress has been broken after pull request #263 

This pull request changes the unit tests to verify **all** paths of **all** ingresses, which exposes the current problem (CI fails on first commit). Also, this PR fixes the issue by removing the extra quote in the ingress file.

### DIFF
--- a/src/main/charts/confluence/templates/ingress.yaml
+++ b/src/main/charts/confluence/templates/ingress.yaml
@@ -37,7 +37,7 @@ spec:
                 port:
                   number: {{ $.Values.confluence.service.port }}
           {{ if .Values.synchrony.enabled }}
-          - path: {{ trimSuffix "/" (include "confluence.ingressPath" .) }}/synchrony"
+          - path: {{ trimSuffix "/" (include "confluence.ingressPath" .) }}/synchrony
             pathType: Prefix
             backend:
               service:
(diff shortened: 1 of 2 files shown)


src/test/java/test/IngressTest.java
```diff
@@ -226,70 +227,167 @@ void jira_ingress_path_default(Product product) throws Exception {
...
+
+        org.assertj.core.api.Assertions.assertThat(ingressPaths).containsExactlyInAnyOrder(
+                "/",
+                "/setup",
+                "/bootstrap");
+    }
+
+    private List<String> extractAllPaths(Traversable<KubeResource> ingresses) {
+        return ingresses
+                .flatMap(ingress -> ingress.getNode("spec", "rules"))
```
badgersow, line 388: This makes tests better, because they assert all ingress paths instead of the first path for each ingress. This is why the problem was missed - because synchrony path is not the first, hence it was not covered by tests

src/main/charts/confluence/templates/ingress.yaml
```diff
@@ -37,7 +37,7 @@ spec:
                 port:
                   number: {{ $.Values.confluence.service.port }}
           {{ if .Values.synchrony.enabled }}
-          - path: {{ trimSuffix "/" (include "confluence.ingressPath" .) }}/synchrony"
```
badgersow, line 40: This is the actual fix

tarka (approved):

nghazali (approved):

PR #282, opened 2021-08-30
DCKUBE-621: Prestop hooks 

tarka:
Add prestop hooks to all statefulsets, and normalise the shutdown values for all products.


badgersow (approved):

badgersow:
@tarka We can merge this pull request only after all docker images have been updated. Otherwise pre-stop hook is going to fail and the process will be killed immediately by SIGKILL.

badgersow:
As discussed with @errcode1202 , we should also change our `image.pullPolicy` to `Always`

> Always
>    every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image digest. If the kubelet has a container image with that exact digest cached locally, the kubelet uses its cached image; otherwise, the kubelet pulls the image with the resolved digest, and uses that image to launch the container.

Currently if the customers upgrade, their cached image will be old and will not have the shutdown script in the container root. PreStop hook is going to fail and JVM will be killed by SIGKILL

src/main/charts/jira/values.yaml
```diff
@@ -461,6 +461,19 @@ jira:
     #
     secretKey: license-key
 
+  shutdown:
```
errcode1202, line 464: For consistency can we:

- include a general comment on what this `shutdown` stanza controls 
- leave a carriage return after `shutdown`

src/main/charts/bitbucket/values.yaml
```diff
@@ -567,6 +562,17 @@ bitbucket:
       #
       passwordSecretKey: password
 
+  shutdown:
```
errcode1202, line 565: For consistency can we:

- include a general comment on what this shutdown stanza controls
- leave a carriage return after shutdown

src/main/charts/bitbucket/values.yaml
```diff
@@ -567,6 +562,17 @@ bitbucket:
...
+    # -- The termination grace period for pods during shutdown. This
+    # should be set to the Bitbucket internal grace period (default 30
+    # seconds), plus a small buffer to allow the JVM to fully terminate.
+    terminationGracePeriodSeconds: 35
+
+    # By default, Bitbucket will be stopped by the script supplied by the image.
+    # If any other shutdown behaviour is needed, it can be achieved by overriding this value.
+    command:
+      - "/shutdown-wait.sh"
```
errcode1202, line 574: Similar to the `fluentd.command` should we not just default to the `/shutdown-wait.sh` script in the respective helper function, meaning it doesn't need to be explicitly defined here? Looking at the helper for this there isnt actually any need to supply this value here
```
{{/*
The command that will be run to properly shutdown Bitbucket
*/}}
{{- define "bitbucket.shutdown.command" -}}
{{- if .Values.bitbucket.shutdown.command }}
{{- .Values.bitbucket.shutdown.command }}
{{- else }}["/shutdown-wait.sh"]{{- end }}
{{- end }}

```

errcode1202:
> As discussed with @errcode1202 , we should also change the default value of our `image.pullPolicy` to `Always`
> 
> > Always
> > every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image digest. If the kubelet has a container image with that exact digest cached locally, the kubelet uses its cached image; otherwise, the kubelet pulls the image with the resolved digest, and uses that image to launch the container.
> 
> Currently if the customers upgrade, their cached image will be old and will not have the shutdown script in the container root. PreStop hook is going to fail and JVM will be killed by SIGKILL

Based on @nanux latest [comment](https://atlassian.slack.com/archives/GFTNVLQQ4/p1630383517101400?thread_ts=1630380019.091000&cid=GFTNVLQQ4) I think we should leave this for now as is.

src/main/charts/bitbucket/values.yaml
```diff
@@ -567,6 +562,17 @@ bitbucket:
       #
       passwordSecretKey: password
 
+  shutdown:
+    # -- The termination grace period for pods during shutdown. This
+    # should be set to the Bitbucket internal grace period (default 30
+    # seconds), plus a small buffer to allow the JVM to fully terminate.
+    terminationGracePeriodSeconds: 35
+
+    # By default, Bitbucket will be stopped by the script supplied by the image.
```
errcode1202, line 571: Should we expand on this comment and include a reference/url to what this script is?

errcode1202 (changes requested):

src/main/charts/confluence/templates/statefulset.yaml
```diff
@@ -155,4 +155,4 @@ spec:
         {{ end }}
         {{ include "confluence.volumes" . | nindent 8 }}
         {{ include "fluentd.config.volume" . | nindent 8 }}
-  {{ include "confluence.volumeClaimTemplates" . | nindent 2 }}
\ No newline at end of file
+  {{ include "confluence.volumeClaimTemplates" . | nindent 2 }}
```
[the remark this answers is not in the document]
badgersow: I think that new line was added to the end of the file, hence it appeared on the diff

tarka:
> @tarka We can merge this pull request only after all docker images have been updated. Otherwise pre-stop hook is going to fail and the process will be killed immediately by SIGKILL.

All images have been updated, and I tested that each product has an `/shutdown-wait.sh` and it is being called.

src/main/charts/bitbucket/values.yaml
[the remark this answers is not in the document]
tarka: The underlying Kubernetes API expects an array of strings (see [ExecAction](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#execaction-v1-core), so I've exposed that.

I'm happy to change this for consistency, but the command needs be wrapped in a shell script, which rapidly descends into quote-hell.

src/main/charts/confluence/templates/statefulset.yaml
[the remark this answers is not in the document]
tarka: This.

src/main/charts/bitbucket/values.yaml
tarka: We provide many of the default values in the `values.yaml`. To me having the command blank in that context implies no command is run by default, which is not true. We should probably be consistent about this, or add a note about what the default is and where it is set.

src/main/charts/bitbucket/templates/_helpers.tpl
```diff
@@ -123,6 +123,15 @@ The command that should be run by the nfs-fixer init container to correct the pe
 {{- end }}
 {{- end }}
 
+{{/*
+The command that will be run to properly shutdown Bitbucket
+*/}}
+{{- define "bitbucket.shutdown.command" -}}
+{{- if .Values.bitbucket.shutdown.command }}
+{{- .Values.bitbucket.shutdown.command }}
+{{- else }}["/shutdown-wait.sh"]{{- end }}
```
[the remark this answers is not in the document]
tarka: This goes to @errcode1202 's comment below about default values. I lean to having them in the `values.yaml` for clarity, but we should be consistent.

src/main/charts/bitbucket/values.yaml
tarka: Done

src/main/charts/crowd/values.yaml
```diff
@@ -213,6 +213,22 @@ crowd:
     #
     usePodNameAsClusterNodeName: true
 
+  shutdown:
+
+    # -- The termination grace period for pods during shutdown. This
+    # should be set to the internal grace period, plus a small buffer
+    # to allow the JVM to fully terminate.
```
errcode1202, line 220: nit - for consistency can we include a `#` after the the property name for all of these i.e.

```
# -- The termination grace period for pods during shutdown. This
# should be set to the internal grace period, plus a small buffer
# to allow the JVM to fully terminate.
#
terminationGracePeriodSeconds: 30
```

errcode1202 (changes requested):

src/main/charts/crowd/values.yaml
tarka: Done

nanux (approved):

errcode1202 (approved):
🍻

nghazali (approved):

PR #283, opened 2021-08-31
Issue/dckube 654 more configuration for synchrony

badgersow:
This pull request addresses the following problems:

* Synchrony pod writes logs to ephemeral storage on a docker container. The logs are lost after pod restart
* Helm charts don't support passing additional JVM arguments to synchrony process
* There is no way to create persistent configuration files for synchrony pod (e.g. file with logging levels config)

To fix it, we

* Introduce a new volume `synchronyHome` to act as a "local home for Synchrony"
* Use `emptyDir` by default for `synchronyHome`, but allow the same config flexibility as elsewhere
* Add a volume mount for the volume into default location `/var/atlassian/application-data/confluence`
* Introduce `synchorny.additionalJvmArgs` analogous to Confluence
* Introduce `synchrony.setPermissions` flag that sets the permissions to `synchronyHome` volume (the same as other products)

The changes were tested by a combination of unit tests and following the scenarios described [here](https://confluence.atlassian.com/doc/configuring-synchrony-858772125.html).


src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -332,6 +340,16 @@ For each additional plugin declared, generate a volume mount that injects that l
 {{- end }}
 {{- end }}
 
+{{- define "synchrony.volumes" -}}
+{{ if not .Values.volumes.synchronyHome.persistentVolumeClaim.create }}
+{{ include "synchrony.volumes.synchronyHome" . }}
+{{- end }}
+{{ include "confluence.volumes.sharedHome" . }}
```
badgersow, line 347: This is the existing behaviour. We mount shared home to make it easy to pass JDBC driver using `additionalLibraries` value.

src/main/charts/confluence/templates/NOTES.txt
```diff
@@ -51,5 +51,11 @@ Get the {{ title .Chart.Name }} URL by running these commands in the same shell:
 ######            Data will be lost when the pod is terminated.             #####
 #################################################################################
 {{- end }}
+{{ if ( and .Values.synchrony.enabled ( not ( or .Values.volumes.synchronyHome.persistentVolumeClaim.create .Values.volumes.synchronyHome.customVolume ))) }}
```
badgersow, line 54: If no persistent volume is specified, we write a warning. This is done only if synchrony is enabled

src/main/charts/confluence/templates/ingress.yaml
```diff
@@ -37,7 +37,7 @@ spec:
                 port:
                   number: {{ $.Values.confluence.service.port }}
           {{ if .Values.synchrony.enabled }}
-          - path: {{ trimSuffix "/" (include "confluence.ingressPath" .) }}/synchrony"
+          - path: {{ trimSuffix "/" (include "confluence.ingressPath" .) }}/synchrony
```
badgersow, line 40: This is the bugfix for ingress typo that has been addressed in #281

src/main/charts/confluence/values.yaml
```diff
@@ -286,8 +286,58 @@ volumes:
       # here.
       #
       command:
-  
-  # -- Defines additional volumes that should be applied to all Confluence pods.
+
+  # Each synchrony pod needs its own volume for 'synchrony-home'. Synchrony process will write logs to that location
+  # and any configuration files can be placed there.
+  #
+  synchronyHome:
```
badgersow, line 293: This copies the structure of `localHome`

src/main/charts/confluence/values.yaml
```diff
@@ -286,8 +286,58 @@ volumes:
...
+    # https://kubernetes.io/docs/concepts/storage/persistent-volumes/#static
+    #
+    customVolume: { }
+    # persistentVolumeClaim:
+    #   claimName: "<pvc>"
+
+    # -- Specifies the path in the Synchrony container to which the synchrony-home volume will be
+    # mounted.
+    #
+    mountPath: "/var/atlassian/application-data/confluence"
```
badgersow, line 338: Because synchrony uses a docker container from Confluence, this is the location of log files

src/main/charts/confluence/templates/NOTES.txt
```diff
@@ -51,5 +51,11 @@ Get the {{ title .Chart.Name }} URL by running these commands in the same shell:
...
+#################################################################################
+######      WARNING: Persistent volume is not used for Synchrony Home!!!    #####
+######            Data will be lost when the pod is terminated.             #####
```
nanux, line 57: This looks very scary. I think we can say "Synchrony logs will be lost...". We are not losing Confluence or collaborative editing data.

badgersow: I agree with that. I've modified the warning message.

src/main/charts/confluence/templates/ingress.yaml
yzha645: Shall we merge main into this branch in this case?

src/main/charts/confluence/values.yaml
```diff
@@ -155,7 +155,7 @@ database:
     
 # REQUIRED - Volume configuration
 #
-# By default, the charts will configure the local-home and shared-home as ephemeral 
+# By default, the charts will configure the local-home, synchrony-home and shared-home as ephemeral
 # volumes i.e. 'emptyDir: {}'. This is fine for evaluation purposes but for production 
 # deployments this is not ideal and so local-home and shared-home should both be configured
```
yzha645, line 160: Since we mentioned synchrony-home above, maybe here as well?

src/main/charts/confluence/values.yaml
```diff
@@ -286,8 +286,58 @@ volumes:
...
+      # created for each pod based on the 'StorageClassName' supplied below.
+      #
+      create: false
+
+      # -- Specify the name of the 'StorageClass' that should be used for the synchrony-home
+      # volume claim.
+      #
+      storageClassName:
+
+      # -- Specifies the standard K8s resource requests and/or limits for the synchrony-home
```
yzha645, line 312: I understand that this comment is everywhere in our chart: `Specifies the standard K8s resource requests and/or limits for .. `, but unlike pod/container, pv, pvc and volumeClaimTemplate do not have option for requests.limits. Maybe we can make this comment more tailored for volume claim? At least omit the `and/or limits` part here.

src/main/charts/confluence/values.yaml
badgersow: Fixed

src/main/charts/confluence/values.yaml
badgersow: Thank you, fixed for all volumes

src/main/charts/confluence/values.yaml
```diff
@@ -517,9 +567,9 @@ confluence:
     container: 
     
       requests:
-        # -- Initial CPU request by Confluence pod
-        #
-        cpu: "2" # -- If changing the cpu value update additional JVM arg 'ActiveProcessorCount' below
+        # -- Initial CPU request by Confluence pod.
+        # If changing the cpu value update additional JVM arg 'ActiveProcessorCount' below
+        cpu: "2"
```
errcode1202, line 572: For consistency can we include a `#` above this property

errcode1202 (changes requested):

src/main/charts/confluence/values.yaml
```diff
@@ -286,8 +286,58 @@ volumes:
       # here.
       #
       command:
-  
-  # -- Defines additional volumes that should be applied to all Confluence pods.
+
+  # Each synchrony pod needs its own volume for 'synchrony-home'. Synchrony process will write logs to that location
```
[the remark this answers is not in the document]
badgersow: Fixed

src/main/charts/confluence/values.yaml
badgersow: Fixed

errcode1202 (approved):
👍🏽

src/main/charts/confluence/values.yaml
```diff
@@ -157,7 +157,7 @@ database:
 #
 # By default, the charts will configure the local-home, synchrony-home and shared-home as ephemeral
 # volumes i.e. 'emptyDir: {}'. This is fine for evaluation purposes but for production 
-# deployments this is not ideal and so local-home and shared-home should both be configured 
+# deployments this is not ideal and so local-home, synchrony-home and shared-home should both be configured
```
yzha645, line 160: should **all** be configured?

badgersow: Thanks, fixed

yzha645 (approved):

PR #288, opened 2021-09-01
Remove redundant README

errcode1202:
Our main landing page `README.md` is under `docs/docs`. This one should no longer be needed.


nanux:
The `README.md` in the root has a different content and it is supposed to be a landing page for the repository. For now, we are not going to remove it as it gives us the flexibility to have different landing for user docs - with mkdocs and Github repository landing page.

errcode1202:
Should its details not reflect those of the one in docs/docs?

PR #292, opened 2021-09-02
Dckube 672 test bitbucket helm chart

yzha645:
While testing Bitbucket chart, found that some links in `values.yaml` file commentary are no longer reachable. Updated these links to the latest documentation.


tan-ro:
What was the reason you moved some of the links from the installation steps to the prerequisites? Did it make more sense when going through them?

src/main/charts/bitbucket/values.yaml
```diff
@@ -8,14 +8,15 @@
 # - volumes
 #
 # Additional details on pre-provisioning these required resources can be found here:
-# https://atlassian.github.io/data-center-helm-charts/installation/INSTALLATION/#3-configure-database
-# https://atlassian.github.io/data-center-helm-charts/installation/INSTALLATION/#5-configure-persistent-storage
+# https://atlassian.github.io/data-center-helm-charts/userguide/PREREQUISITES/#provision-a-database
```
yzha645, line 11: @tan-ro I think I misunderstood the `pre-provisioning` here. I was thinking pre-provisioning db/volumes themselves, instead of pre-configure those resources in the product. 
So what you mean here is how to configure those stanza, right? Then links to installation steps make more sense. I'll update them.

tan-ro: Great. Let me know if you need a hand

tan-ro (approved):

nanux (approved):

nanux:
Thank you @yzha645, that is a great catch!

PR #297, opened 2021-09-06
Additional documentation updates

errcode1202:
Fixes with:

* Formatting
* Styling
* Typos
* Some re-wording

### DIFF
--- a/docs/docs/examples/logging/efk/EFK.md
+++ b/docs/docs/examples/logging/efk/EFK.md
@@ -5,7 +5,7 @@
 
 ## EFK stack
 
-A common Kubernetes logging pattern is the combination of ***Elasticsearch***, ***Fluentd***, and ***Kibana***, known as *EFK Stack*. 
+A common Kubernetes logging pattern is the combination of `Elasticsearch`, `Fluentd`, and `Kibana`, known as *EFK Stack*. 
 
 [Fluentd](https://www.fluentd.org/){.external} is an open-source and multi-platform log processor that collects data/logs from different sources, aggregates, and forwards them to multiple destinations. It is fully compatible with Docker and Kubernetes environments. 
 
@@ -48,9 +48,9 @@ $ curl localhost:9200
 }
 ```
 
-### 2. Enable fluentd
+### 2. Enable Fluentd
 
-Now enable `fluentd` and set the `hostname` for Elasticsearch in `values.yaml` as follows:
+Now enable `Fluentd` and set the `hostname` for Elasticsearch in `values.yaml` as follows:
 
 ```yaml
 fluentd:
@@ -86,16 +86,16 @@ kubectl port-forward deployment/kibana-kibana 5601
 
 ## Managed EFK stack
 
-In this solution [Elasticsearch is deployed as a managed AWS service](https://aws.amazon.com/elasticsearch-service/){.external} and lives outside of the Kubernetes cluster. This approach uses [fluentBit](https://fluentbit.io/){.external} instead of `fluentd` for log processing.
+In this solution [Elasticsearch is deployed as a managed AWS service](https://aws.amazon.com/elasticsearch-service/){.external} and lives outside of the Kubernetes cluster. This approach uses [fluentBit](https://fluentbit.io/){.external} instead of `Fluentd` for log processing.
 
-???+ info "fluentbit"
+???+ info "Fluentbit"
 
-    `fluentbit` is used to collect and aggregate log data inside the EKS cluster. It then sends this to an AWS Elasticsearch instance outside of the cluster.
+    `Fluentbit` is used to collect and aggregate log data inside the EKS cluster. It then sends this to an AWS Elasticsearch instance outside of the cluster.
 
 
-When a node inside an EKS cluster needs to call an AWS API, it needs to provide extended permissions. Amazon provides an image of `fluentbit` that supports AWS service accounts,and using this you no longer need to follow the traditional way. All you need is to have an IAM role for the AWS service account on an EKS cluster. Using this service account, an AWS permission can be provided to the containers in any pod that use that service account. The result is that the pods on that node can call AWS APIs.
+When a node inside an EKS cluster needs to call an AWS API, it needs to provide extended permissions. Amazon provides an image of `Fluentbit` that supports AWS service accounts,and using this you no longer need to follow the traditional way. All you need is to have an IAM role for the AWS service account on an EKS cluster. Using this service account, an AWS permission can be provided to the containers in any pod that use that service account. The result is that the pods on that node can call AWS APIs.
 
-Your first step is to configure IAM roles for Service Accounts (IRSA) for `fluentbit`, to make sure you have an OIDC identity provider to use IAM roles for the service account in the cluster:
+Your first step is to configure IAM roles for Service Accounts (IRSA) for `Fluentbit`, to make sure you have an OIDC identity provider to use IAM roles for the service account in the cluster:
 
 ```shell
 eksctl utils associate-iam-oidc-provider \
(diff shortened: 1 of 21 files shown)


docs/docs/examples/logging/efk/EFK.md
```diff
@@ -238,13 +252,13 @@ $ curl -sS -u "${ES_DOMAIN_USER}:${ES_DOMAIN_PASSWORD}" \
 ]
 '
 ```
-Finally, it is time to deploy Fluent Bit DaemonSet:
+Finally, it is time to deploy the `Fluentbit` DaemonSet:
```
errcode1202, line 255: @nghazali where does the `fluentbit.yaml` config file come from in this case?

docs/docs/examples/logging/efk/EFK.md
```diff
@@ -238,13 +252,13 @@ $ curl -sS -u "${ES_DOMAIN_USER}:${ES_DOMAIN_PASSWORD}" \
...
 ```shell
-$ kubectl apply -f src/main/logging/fluentbit.yaml
+kubectl apply -f src/main/logging/fluentbit.yaml
```
errcode1202, line 257: @nghazali where does the fluentbit.yaml config file come from in this case?

nghazali: That's right. This file is moved to this path `docs/docs/examples/logging/efk/managed_es/`

docs/docs/examples/logging/efk/EFK.md
```diff
@@ -5,44 +5,27 @@
...
-elasticsearch-master-1   1/1     Running   0          100m
-elasticsearch-master-2   1/1     Running   0          100m
-$
-$ kubectl port-forward svc/elasticsearch-master 9200
-```
-Make sure Elasticsearch cluster is working as expected:
+Install Elasticsearch using the instructions [documented here](../../elasticsearch/BITBUCKET_ELASTICSEARCH.md). Once installed make sure Elasticsearch cluster is working as expected:
 
 ```shell
 $ curl localhost:9200
```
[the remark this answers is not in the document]
nghazali: I think is better to move `port-forward` command in right before `curl` command. In fact, we should forward the port to be able to use it in localhost, but this is not make sense to put in in the block related to get elasticsearch pods. 

Move line 43 => before line 48

docs/docs/examples/logging/efk/EFK.md
```diff
@@ -5,44 +5,27 @@
...
-$ kubectl get pods --namespace=dcd -l app=elasticsearch-master
-NAME                     READY   STATUS    RESTARTS   AGE
-elasticsearch-master-0   1/1     Running   0          100m
-elasticsearch-master-1   1/1     Running   0          100m
-elasticsearch-master-2   1/1     Running   0          100m
-$
-$ kubectl port-forward svc/elasticsearch-master 9200
-```
-Make sure Elasticsearch cluster is working as expected:
+Install Elasticsearch using the instructions [documented here](../../elasticsearch/BITBUCKET_ELASTICSEARCH.md). Once installed make sure Elasticsearch cluster is working as expected:
```
nghazali, line 28: The elastic installation pointed on this line is for Bitbucket search. This is very similar and it probably works, but in installing it adds an `image.tag` which is not needed when we provision an elasticsearch for EFK stack. 

Also we should change the `image.tag="7.9.3" in line 15 of `BITBUCKET_ELASTICSEARCH.md` as the default version for bitbucket is different in out helm chart (7.15.1)

docs/docs/examples/logging/efk/EFK.md
```diff
@@ -75,51 +60,62 @@ fluentd:
...
-Your first step is to configure IAM roles for Service Accounts (IRSA) for `fluentbit`, to make sure you have an OIDC identity provider to use IAM roles for the service account in the cluster:
+
+When a node inside an EKS cluster needs to call an AWS API, it needs to provide extended permissions. Amazon provides an image of `Fluentbit` that supports AWS service accounts,and using this you no longer need to follow the traditional way. All you need is to have an IAM role for the AWS service account on an EKS cluster. Using this service account, an AWS permission can be provided to the containers in any pod that use that service account. The result is that the pods on that node can call AWS APIs.
+
+Your first step is to configure IAM roles for Service Accounts (IRSA) for `Fluentbit`, to make sure you have an OIDC identity provider to use IAM roles for the service account in the cluster:
 
 ```shell
-$ eksctl utils associate-iam-oidc-provider \
+eksctl utils associate-iam-oidc-provider \
      --cluster dcd-ap-southeast-2 \
```
nghazali, line 102: Maybe good to change cluster name to something else ( `<cluster-name>` or `my-k8s-cluster` ). WDYT?

docs/docs/examples/logging/efk/EFK.md
```diff
@@ -206,27 +209,38 @@ $ cat <<EOF> ~/environment/logging/elasticsearch_domain.json
...
+```shell
+eksctl get iamserviceaccount --cluster dcd-ap-southeast-2
+```
+The output of this command should look similar to this:
 ```shell
-$ eksctl get iamserviceaccount --cluster dcd-ap-southeast-2
-[ℹ] eksctl version 0.37.0
-[ℹ] using region ap-southeast-2
 NAMESPACE    NAME                ROLE ARN
 kube-system cluster-autoscaler   arn:aws:iam::887464544476:role/eksctl-dcd-ap-southeast-2-addon-iamserviceac-Role1-1RSRFV0BQVE3E
```
nghazali, line 231: I think we should redact or change the dcd account number with a random number as this documents are public (here and anywhere else we used it in samples)

docs/docs/examples/logging/efk/EFK.md
errcode1202: Thanks

docs/docs/examples/logging/efk/EFK.md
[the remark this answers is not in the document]
errcode1202: Done.

docs/docs/examples/logging/efk/EFK.md
errcode1202: changes to `<cluster-name>`

docs/docs/examples/logging/efk/EFK.md
errcode1202: I actually think the `image.tag` is required, as taken from our ES install page:

> Choose a version of Elasticsearch that is supported by the [version of Bitbucket you are installing](https://confluence.atlassian.com/bitbucketserver/supported-platforms-776640981.html#Supportedplatforms-additional-toolsAdditionaltools)

errcode1202: As for the `image.tag="7.9.3`, based on the current [app version defined in the BB Chart.yaml](https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bitbucket/Chart.yaml#L6) i.e. `appVersion: 7.15.1-jdk11`, then Elasticsearch `7.9.3` [should be compatible](https://confluence.atlassian.com/bitbucketserver0715/supported-platforms-1082271274.html)?

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -299,7 +308,9 @@ jira:
         memory: "8G"
 ``` 
 
-By default, no container-level resource limits or requests are set.
+!!!tip ""
+
+    By default, the Helm Charts have no container-level resource limits or requests are set.
```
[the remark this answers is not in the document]
errcode1202: Good spot. Fixed.

docs/docs/examples/logging/efk/EFK.md
nghazali: Right, We should have `image.tag` for elasticsearch in BB which currently the latest supported version is 7.10.2, but my point is this version is Bitbucket specific supported version. 
When we talk about EFK is a bit different because the elasticsearch is not interacting with our product directly so the limitation of BB for supporting versions does not applied. 

However, having 'image.tag' and the either `7.9.3` or `7.10.2` versions will works for EFK stack, but there is no logic behind that why we chosen this specific version for EFK and not any other version. The EFK stack works with ES 7.14 as well as version 7.9.

nghazali: I will leave it to you to decide if we need to redundant the installation commands or link to the existing one wrote for BB. The differences is pointing a specific versions which is needed for BB but is not for EFK stack.

nghazali (approved):

nanux (approved):

docs/docs/examples/logging/efk/EFK.md
errcode1202: I think for consistency and simplification let's leave the tag as defined in the ES installation instructions this now points to.

PR #299, opened 2021-09-08
DCKUBE-686: decrease Confluence failover time

badgersow:
If the Confluence node becomes unhealthy, it takes 2.5 minutes of failing `/status` checks for load balancer to exclude the node.
It's because the readiness probe has a period of 5 seconds and `failureThreshold` of 30. The product of those values is 150 seconds, which corresponds to 2 minutes and 30 seconds
In practice this means that the customer, whose session was pointing to the failed node, will continue experiencing errors for at least 2 minutes and 30 seconds until they can work with the application again.
2:30 seems like too much. There is no reason I know failover should be this long. In this pull request I'm changing the chart defaults to have 30 seconds of failover - this coincides with the reasonable timeout for HTTP request and seems like a better value for failover.
This change can also be useful for other products (Jira, BBS, Crowd).


nghazali:
Some unit tests related to failover is failed and need to be fixed.

badgersow:
> Some unit tests related to failover is failed and need to be fixed.

Thank you, I've fixed them

nghazali (approved):

nanux (approved):

PR #301, opened 2021-09-16
DCKUBE-678: Add scheduleName to StatefulSet

yzha645:
Kubernetes scheduler was not configurable in our Helm charts, which prevented user from using their custom scheduler.  
Added `schedulerName` to values.yaml file and StatefulSet template. User now has the flexibility to configure which scheduler to use.


nghazali (approved):

src/main/charts/bitbucket/values.yaml
```diff
@@ -777,6 +777,10 @@ tolerations: []
 #        topologyKey: topology.kubernetes.io/zone
 affinity: {}
 
+# -- Standard K8s schedulerName that will be applied to all Bitbucket pods
```
nanux, line 780: Maybe link to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podspec-v1-core
or https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#specify-schedulers-for-pods

nanux (approved):

errcode1202 (approved):

PR #309, opened 2021-09-23
DCKUBE-677: Make securityContext more flexible

yzha645:
`securityContext` had only `fsGroup` configurable by user supplying `gid` in values.yaml file. However, gid is not a K8s standard field, and we have to convert gid to fsGroup within _helpers template. 
Now made securityContext fully configurable by user supplying the whole section in values.yaml file, so that more fields than `fsGroup` are able to be configured. 

Moreover, there are two levels of securityContext in K8s world: pod level and container level. This PR also separated these two levels by defining `securityContext` stanza for pod, and `containerSecurityContext` stanza for container. User now have 100% flexibility to define securityContext on both pod and container level.    

`Values.bitbucket.securityContext.enabled` and `Values.bitbucket.securityContext.gid` will de deprecated. Breaking changes will have to be introduced for existing customers who wants to upgrade Helm chart.

### DIFF
--- a/src/main/charts/bitbucket/templates/_helpers.tpl
+++ b/src/main/charts/bitbucket/templates/_helpers.tpl
@@ -119,7 +119,7 @@ The command that should be run by the nfs-fixer init container to correct the pe
 {{- if .Values.volumes.sharedHome.nfsPermissionFixer.command }}
 {{ .Values.volumes.sharedHome.nfsPermissionFixer.command }}
 {{- else }}
-{{- printf "(chgrp %s %s; chmod g+w %s)" .Values.bitbucket.securityContext.gid .Values.volumes.sharedHome.nfsPermissionFixer.mountPath .Values.volumes.sharedHome.nfsPermissionFixer.mountPath }}
+{{- printf "(chgrp %v %s; chmod g+w %s)" .Values.bitbucket.securityContext.fsGroup .Values.volumes.sharedHome.nfsPermissionFixer.mountPath .Values.volumes.sharedHome.nfsPermissionFixer.mountPath }}
 {{- end }}
 {{- end }}
 
--- a/src/main/charts/bitbucket/templates/statefulset.yaml
+++ b/src/main/charts/bitbucket/templates/statefulset.yaml
@@ -26,12 +26,8 @@ spec:
     spec:
       serviceAccountName: {{ include "bitbucket.serviceAccountName" . }}
       terminationGracePeriodSeconds: {{ .Values.bitbucket.shutdown.terminationGracePeriodSeconds }}
-      {{ if .Values.bitbucket.securityContext.enabled }}
       securityContext:
-        # This is intended to ensure that the shared-home volume is group-writeable by the GID used by the Bitbucket container.
-        # However, this doesn't appear to work for NFS volumes due to a K8s bug: https://github.com/kubernetes/examples/issues/260
-        fsGroup: {{ .Values.bitbucket.securityContext.gid }} # Makes sure that the shared-home volume mount is writeable by the bitbucket user's GID
-      {{ end }}
+        {{ toYaml .Values.bitbucket.securityContext | nindent 8 }}
       initContainers:
         {{- include "bitbucket.additionalInitContainers" . | nindent 8 }}
         {{- if .Values.volumes.sharedHome.nfsPermissionFixer.enabled }}
@@ -69,6 +65,10 @@ spec:
             periodSeconds: 5
             failureThreshold: 60
             initialDelaySeconds: 10
+          {{- with .Values.bitbucket.containerSecurityContext}}
+          securityContext:
+            {{- toYaml . | nindent 12}}
+          {{- end}}
           volumeMounts:
             - name: local-home
               mountPath: {{ .Values.volumes.localHome.mountPath | quote }}
--- a/src/main/charts/bitbucket/templates/tests/test-shared-home-permissions.yaml
+++ b/src/main/charts/bitbucket/templates/tests/test-shared-home-permissions.yaml
@@ -17,8 +17,8 @@ spec:
       imagePullPolicy: IfNotPresent
       securityContext:
         # Slightly dodgy; we assume that the UID and GID used by the product images are the same, which in practice they are
-        runAsUser: {{ .Values.bitbucket.securityContext.gid }}
-        runAsGroup: {{ .Values.bitbucket.securityContext.gid }}
+        runAsUser: {{ .Values.bitbucket.securityContext.fsGroup }}
+        runAsGroup: {{ .Values.bitbucket.securityContext.fsGroup }}
       volumeMounts:
         - name: shared-home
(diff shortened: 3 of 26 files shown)


src/main/charts/bitbucket/templates/tests/test-shared-home-permissions.yaml
```diff
@@ -17,8 +17,8 @@ spec:
       imagePullPolicy: IfNotPresent
       securityContext:
         # Slightly dodgy; we assume that the UID and GID used by the product images are the same, which in practice they are
-        runAsUser: {{ .Values.bitbucket.securityContext.gid }}
-        runAsGroup: {{ .Values.bitbucket.securityContext.gid }}
+        runAsUser: {{ .Values.bitbucket.securityContext.fsGroup }}
```
nanux, line 20: What would happen if somebody overrides the full security context and leave out the `fsGroup`? When we say that users have full flexibility to change the `securityContext`, they can probably also include the context without `fsGroup. We should either cater for the default or document that.

src/main/charts/bitbucket/templates/tests/test-shared-home-permissions.yaml
```diff
@@ -17,8 +17,8 @@ spec:
       imagePullPolicy: IfNotPresent
       securityContext:
         # Slightly dodgy; we assume that the UID and GID used by the product images are the same, which in practice they are
```
nanux, line 19:
```suggestion
        # We assume that the UID and GID used by the product images are the same, which in practice they are
```

src/main/charts/bitbucket/values.yaml
```diff
@@ -445,20 +445,27 @@ bitbucket:
...
-    #
-    enabled: true
-    
+
     # -- The GID used by the Bitbucket docker image
     #
-    gid: "2003"
+    fsGroup: 2003
+
+#    # -- Set to 'true' to enable the security context (Deprecated)
```
nanux, line 457: With the current implementation, this is not deprecated, rather it is not working anymore. Deprecated would mean that you can still use it but it might be removed in future.

src/main/charts/bitbucket/values.yaml
```diff
@@ -445,20 +445,27 @@ bitbucket:
...
+#    #
+#    enabled: true
+#
+#    # -- The GID used by the Bitbucket docker image (Deprecated)
+#    #
+#    gid: "2003"
+
+
+  # -- Standard K8s field that holds security configurations that will be applied to a container.
```
nanux, line 466: Include link to https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

src/main/charts/bitbucket/values.yaml
```diff
@@ -445,20 +445,27 @@ bitbucket:
     # needs extra annotations.
     #
     annotations: {}
-    
-  # Enable or disable security context in StatefulSet template spec.
-  # Enabled by default with UID 2003. Disable when deploying to OpenShift,
-  # unless anyuid policy is attached to service account.
+
+  # Standard K8s field that holds pod-level security attributes and common container settings.
```
nanux, line 449: Include link to https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

src/main/charts/bitbucket/templates/_helpers.tpl
```diff
@@ -119,7 +119,7 @@ The command that should be run by the nfs-fixer init container to correct the pe
 {{- if .Values.volumes.sharedHome.nfsPermissionFixer.command }}
 {{ .Values.volumes.sharedHome.nfsPermissionFixer.command }}
 {{- else }}
-{{- printf "(chgrp %s %s; chmod g+w %s)" .Values.bitbucket.securityContext.gid .Values.volumes.sharedHome.nfsPermissionFixer.mountPath .Values.volumes.sharedHome.nfsPermissionFixer.mountPath }}
+{{- printf "(chgrp %v %s; chmod g+w %s)" .Values.bitbucket.securityContext.fsGroup .Values.volumes.sharedHome.nfsPermissionFixer.mountPath .Values.volumes.sharedHome.nfsPermissionFixer.mountPath }}
```
nanux, line 122: See my other comment - users are now able to define `securityContext` without fsGroup. We probably should include default value.

src/test/java/test/SecurityContextTest.java
```diff
@@ -0,0 +1,50 @@
...
+import test.model.Product;
+
+import java.util.Map;
+
+import static test.jackson.JsonNodeAssert.assertThat;
+
+/**
+ * Tests the various permutations of the "<product>.securityContext" and "<product>.containerSecurityContext" value structure in the Helm charts
+ */
+class SecurityContextTest {
```
nanux, line 18: We need to add a test for a case when `securityContext.fsGroup` is missing.

src/test/java/test/SecurityContextTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @BeforeEach
+    void initHelm(TestInfo testInfo) {
+        helm = new Helm(testInfo);
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = Product.class)
+    void test_pod_security_context(Product product) throws Exception {
+
+        final var resources = helm.captureKubeResourcesFromHelmChart(product, Map.of(
```
nanux, line 30: Please add another securityContext value as well, e.g. runAsUser: 1000. We were using the `fsGroup` even before this change.

src/main/charts/bitbucket/templates/statefulset.yaml
```diff
@@ -26,12 +26,8 @@ spec:
     spec:
       serviceAccountName: {{ include "bitbucket.serviceAccountName" . }}
       terminationGracePeriodSeconds: {{ .Values.bitbucket.shutdown.terminationGracePeriodSeconds }}
-      {{ if .Values.bitbucket.securityContext.enabled }}
       securityContext:
-        # This is intended to ensure that the shared-home volume is group-writeable by the GID used by the Bitbucket container.
```
nanux, line 31: I wonder if we are not losing useful information, if the bug still persists.

nanux (changes requested):

errcode1202:
More of a reminder to myself, once merged into `main` ill need to merge these changes into #316

errcode1202 (changes requested):
@yzha645 now that the Bamboo chart is merged it probably makes sense to also update it with these changes.

errcode1202:
@yzha645 can these changes also be applied the bamboo-agent: https://github.com/atlassian/data-center-helm-charts/tree/main/src/main/charts/bamboo-agent

BREAKING_CHANGES.md
```diff
@@ -0,0 +1,15 @@
+# Breaking Changes
+
+## Release 2.0.0
```
nanux, line 3:
```suggestion
## Version 2.0.0
```

nanux (approved):

PR #313, opened 2021-09-28
DCKUBE-722: Enable configuring ingress.class name

nanux:
This replaces https://github.com/atlassian/data-center-helm-charts/pull/267

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -9,8 +9,6 @@ on:
       - main
       - /.*-stable/
   pull_request:
-    branches:
-      - main
 
 jobs:
   build:
--- a/src/main/charts/bamboo/README.md
+++ b/src/main/charts/bamboo/README.md
@@ -1,5 +1,3 @@
-🚧 These charts are currently under construction! 🚧
-
 # bamboo
 
 ![Version: 0.0.1](https://img.shields.io/badge/Version-0.0.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 8.0.1-jdk11](https://img.shields.io/badge/AppVersion-8.0.1--jdk11-informational?style=flat-square)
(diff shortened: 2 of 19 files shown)


.github/workflows/maven.yml
```diff
@@ -9,8 +9,6 @@ on:
       - main
       - /.*-stable/
   pull_request:
-    branches:
```
nanux, line 12: We should run the build for any PR. Currently, it won't run for example if we are targeting `release/x.y.z`

src/main/charts/bamboo/values.yaml
```diff
@@ -275,6 +275,14 @@ ingress:
   # pre-provisioned Ingress Controller being available.
   # 
   create: false
+
+  # -- The class name used by the ingress controller if it's being used.
```
nanux, line 279: Looking for feedback for the documentation text in the `values.yaml` files. 🙏 

This is a relevant documentation - https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class but I didn't include it as it promotes `IngressClass` objects that are not created by default. TBH, I think it a bit confusing. K8s has introduced `ingressClassName` into ingress `spec` but `ingress.class` annotation still t[akes a preference if it is there](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#ingressspec-v1-networking-k8s-io). For example, Nginx and ALB ingress controllers are still listing `ingress.class` in their installation documentation.

src/main/charts/bamboo/README.md
```diff
@@ -76,6 +74,7 @@ Kubernetes: `>=1.19.x-0`
 | image.repository | string | `"atlssmith/bamboo"` | The Bamboo Docker image to use https://hub.docker.com/r/atlassian/bamboo-server |
 | image.tag | string | `""` | The docker image tag to be used - defaults to the Chart appVersion |
 | ingress.annotations | object | `{}` | The custom annotations that should be applied to the Ingress Resource  when NOT using the K8s ingress-nginx controller. |
+| ingress.className | string | `"nginx"` | The class name used by the ingress controller if it's being used. Please follow documenation of your ingress controller. If the cluster  contains multiple ingress controllers, this setting allows you to control which of them is used for Atlassian application traffic. |
```
nanux, line 77: 🎍

errcode1202:
More of a reminder to myself, once merged into `main` ill need to merge these changes into #316

errcode1202 (changes requested):
@nanux now that the Bamboo chart is merged it probably makes sense to also update it with these changes.

errcode1202:
> @nanux now that the Bamboo chart is merged it probably makes sense to also update it with these changes.

Ignore me, I see you've already done this!

errcode1202 (approved):

nghazali (approved):

PR #314, opened 2021-09-28
DCKUBE-634: Bitbucket - set context path

nanux:
This change enables to use of a context path for Bitbucket.


errcode1202 (approved):

src/main/charts/bitbucket/templates/_helpers.tpl
```diff
@@ -108,6 +108,13 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{ with .Values.ingress.port }}:{{ . }}{{ end }}
 {{- end }}
 
+{{/*
+Create default value for ingress path
```
errcode1202, line 112: Any reason why we dont use the same pattern here as we do for Jira, confluence and Bamboo i.e.

https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/jira/templates/_helpers.tpl#L44

nanux: I am happy to bring it here but I am not really aware of any use case when this is necessary - i.e. having a different context path and ingress path.

Let me fix it at least for consistency reasons.

nghazali (approved):

errcode1202 (approved):

Issue #312, opened 2021-09-26
[Suggestion] - Automatically roll StatefulSet if changes occur in ConfigMaps


dennisfischer:
#### Problem
The helm charts ship with at least one ConfigMap for each application, e.g., https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/jira/templates/config-jvm.yaml which can be updated by providing the respective values in `values.yaml`. However, the StatefulSet will not perform an update as it is not modified.

#### Details
An administrator might want to change the JVM memory options. A change to `jvm.maxHeap`, `jvm.minHeap` and `jvm.reservedCodeCache` will not roll the StatefulSet, because only the ConfigMap receives an update. A change to `requests` and `limits` of a container will roll the StatefulSet, because the StatefulSet gets updated. This seems like an inconsistent behavior, given that the StatefulSet will always roll if container `requests` and `limits` get updated.

The helm documentation shows an example how ConfigMaps can force updates: https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

#### Suggestion
Add the respective annotations such that changes to ConfigMaps perform updates to the StatefulSet

I can provide the necessary changes in a PR if wanted.

### Product

Jira, Confluence, Bitbucket

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

nanux:
@dennisfischer, I think you have a good point. Would you be able to raise a PR with the changes?

bianchi2:
As a side note, there's a project https://github.com/stakater/Reloader which does the same. Helm built-in approach is better though. Changes to mounted cms and secrets should trigger rollout restart.

dennisfischer:
@nanux Will raise a PR with the changes once ready. I have already prepared the changes in my fork w.r.t. adding the checksums. Tests still tbd.

PR #315, opened 2021-09-28
Roll Statefulset Pods if ConfigMap changes

dennisfischer:
This PR implements the suggested changes in #312 

A fluentd configuration for synchrony seems to be missing. As such, this PR does not calculate the fluentd ConfigMap checksum for the `statefulset-synchrony.yaml`.


nanux:
Thanks, @dennisfischer for the contribution. We have tested this and the pods are correctly restarted and picking up the configmap changes. Feel free to merge.

PR #316, opened 2021-09-29
BAMBK8S-47: Bamboo agent Helm chart

errcode1202:
A chart for deploying bamboo agents. Agents are:

- ephemeral in nature
- short-lived
-  not public
-  only communicate with the pre-provisioned [bamboo server](https://github.com/atlassian/data-center-helm-charts/tree/main/src/main/charts/bamboo). 

As such:

- they do not have a `database`
- they do not have `local-home`
- they do not have `shared-home`
- they do not have an `ingress`
- `fluentd` support is deliberately not provided - Log capture for ephemeral agents seems overkill. Happy to discuss this further though.

Additionally the chart does not support the following as the other product charts do:

- `additionalLibraries`
- `additionalBundledPlugins`
- `additionalVolumeMounts`
- `additionalJvmArgs` - image does not support `JVM_SUPPORT_RECOMMENDED_ARGS`

Notes about tests:

Due to the fact that many of these features are not supported some of the unit tests have been updated to skip testing the `agent` charts. This is done with the annotation:

```java
@EnumSource(value = Product.class, names = {"bamboo_agent"}, mode = EnumSource.Mode.EXCLUDE)
```

**Note this change also includes some minor updates to the bamboo server chart including additional tests**


src/main/charts/agent/templates/tests/test-application-status.yaml
```diff
@@ -0,0 +1,31 @@
+{{/*TODO: Once status endpoint is available re-instate this test*/}}
```
errcode1202, line 1: Bamboo agents to not have a `/status` endpoint yet. Ticket created to be actioned by Bamboo team:
https://bulldog.internal.atlassian.com/browse/BAMBK8S-84

src/main/charts/agent/templates/deployment-agent.yaml
```diff
@@ -0,0 +1,84 @@
...
+              valueFrom:
+                configMapKeyRef:
+                  key: max_heap
+                  name: {{ include "agent.fullname" . }}-jvm-config
+            {{- include "agent.additionalEnvironmentVariables" . | nindent 12 }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.agent.ports.http }}
+              protocol: TCP          
+{{/*              TODO: Enable readiness probe once available*/}}
```
errcode1202, line 53: Bamboo agents to not have a /status endpoint yet. Ticket created to be actioned by Bamboo team:
https://bulldog.internal.atlassian.com/browse/BAMBK8S-84

src/main/charts/agent/values.yaml
```diff
@@ -0,0 +1,233 @@
...
+replicaCount: 1
+
+# Image configuration 
+#
+image:
+
+  # -- The Bamboo agent Docker image to use
+  # https://hub.docker.com/r/atlassian/bamboo-agent-base
+  #
+  # TODO: This needs to be updated when docker porting work is complete
```
errcode1202, line 28: Will be updates when docker updates have been merged

src/main/charts/agent/values.yaml
```diff
@@ -0,0 +1,233 @@
...
+
+  # Port definitions
+  #
+  ports:
+  
+    # -- The port on which the Bamboo agent listens for HTTP traffic
+    #
+    http: 80
+     
+  # TODO: As of yet a /status endpoint does not exist for agents. Once delivered the implementation for this
```
errcode1202, line 104: Bamboo agents to not have a /status endpoint yet. Ticket created to be actioned by Bamboo team:
https://bulldog.internal.atlassian.com/browse/BAMBK8S-84

src/test/java/test/ContainersTest.java
```diff
@@ -27,7 +27,7 @@ void initHelm(TestInfo testInfo) {
     }
 
     @ParameterizedTest
-    @EnumSource
+    @EnumSource(value = Product.class, names = {"agent"}, mode = EnumSource.Mode.EXCLUDE)
```
nanux, line 30: There are a fair bit of Exclusions in the unit tests. Would it make sense to leave out bamboo agent out of the Product enum? It is not really a product anyway and I think going forward they will probably diverge more than converge.
You have a specialised test for it and it can be potentially expanded if we are missing something critical from the other parametrised tests.

src/main/charts/agent/values.yaml
```diff
@@ -0,0 +1,233 @@
...
+    # size, plus other JVM overheads, must be accommodated. 
+    # Allowing for (maxHeap)*1.5 would be an example.
+    #
+    container:
+      
+      requests:
+      
+        # -- Initial CPU request by Bamboo agent pod
+        #
+        cpu: "2" # If changing the cpu value update 'ActiveProcessorCount' below
```
nanux, line 174: Are those recommended values for Bamboo agent?

src/main/charts/agent/values.yaml
```diff
@@ -0,0 +1,233 @@
...
+    
+  # Enable or disable security context in StatefulSet template spec. Enabled 
+  # by default with UID 2005. Disable when deploying to OpenShift, unless anyuid 
+  # policy is attached to a service account.
+  #
+  securityContext:
+  
+    # -- Set to 'true' to enable the security context
+    #
+    enabled: true
```
nanux, line 90: This has been changed by @yzha645 in https://github.com/atlassian/data-center-helm-charts/pull/309 I would recommend following the same approach.

src/main/charts/agent/values.yaml
```diff
@@ -0,0 +1,233 @@
...
+  # -- For Docker images hosted in private registries, define the list of image pull 
+  # secrets that should be utilized by the created ServiceAccount
+  # https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
+  #
+  imagePullSecrets: []
+  # - name: secretName
+
+# Bamboo agent configuration 
+#
+agent:
```
nanux, line 65: I think we can pull this stanza to the top after `replicaCount` it will be the one part that needs to be changed so it is better when we lead with it.

src/test/java/test/ContextPathTest.java
```diff
@@ -36,8 +36,14 @@ void test_context_path(Product product) throws Exception {
                 .assertHasValue("ATL_TOMCAT_CONTEXTPATH", "/" + product.name());
 
 
-        assertEquals(resources.getStatefulSet(
-                product.getHelmReleaseName()).getContainer().get("readinessProbe").get("httpGet").get("path").asText(),
-                "/" + product.name() + "/status");
+        if(product.name().equals("bamboo")) {
```
nanux, line 39: I would extract it to a different test. Having a conditional based on the input is making the test more opaque.

src/test/java/test/IngressTest.java
```diff
@@ -222,7 +222,103 @@ void jira_ingress_path_default(Product product) throws Exception {
...
+
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = Product.class, names = "bamboo")
+    void bamboo_ingress_path_value(Product product) throws Exception {
+        final var resources = helm.captureKubeResourcesFromHelmChart(product, Map.of(
+                "ingress.create", "true",
+                "ingress.host", "myhost.mydomain",
+                "bamboo.service.contextPath", "/context",
```
nanux, line 283: I was recently thinking about that. Is there a real use case to have `ingresPath` and `contextPath` different? I didn't implement this use case for Bitbucket helm chart in https://github.com/atlassian/data-center-helm-charts/pull/314 because I am not sure if there is a need for it. Is that different for Bamboo?

src/main/charts/agent/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
```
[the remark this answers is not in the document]
errcode1202: Yup can do this, would `bamboo_agent` be okay? Reason being, the tests [utilize an enum](https://github.com/atlassian/data-center-helm-charts/blob/main/src/test/java/test/model/Product.java) as part of their execution. Each chart needs to be represented in this enum and as `-` hyphens are not valid for enum keys an `_` underscore is what we need here.

src/main/charts/agent/templates/NOTES.txt
```diff
@@ -0,0 +1,29 @@
...
+is advisable to wait for the tests to pass before continuing.
+
+To see the custom values you used for this release:
+
+  $ helm get values {{ .Release.Name }} -n {{ .Release.Namespace }}
+  
+{{ if not .Values.agent.securityToken -}}
+#################################################################################
+######      WARNING: Security token not provided for authentication!!!      #####
+######      Agents will not be able to authenticate with Bamboo server.     #####
```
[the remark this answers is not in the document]
errcode1202: I know what you mean, however the actual bamboo server docker image is also named as such: https://bitbucket.org/atlassian-docker/docker-bamboo-server/src

src/main/charts/agent/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - Bamboo DC
+  - Atlassian
+home: https://www.atlassian.com/software/bamboo
+icon: https://atlassian-labs.github.io/data-center-helm-charts/icons/bamboo.svg
+sources:
+  - https://github.com/atlassian-labs/data-center-helm-charts
+  - https://bitbucket.org/atlassian-docker/docker-bamboo-agent-base
+deprecated: false
+annotations:
+  artifacthub.io/containsSecurityUpdates: "true"
```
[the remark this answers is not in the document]
errcode1202: Which part?

src/main/charts/agent/templates/_helpers.tpl
```diff
@@ -0,0 +1,128 @@
...
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+The base url of the Bamboo DC server
+*/}}
+{{- define "agent.serverBaseUrl" -}}
+{{- if .Values.agent.server }}
+{{- printf "http://%s" .Values.agent.server }}
```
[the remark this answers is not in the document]
errcode1202: So, this is the internal K8s DNS name designated to the bamboo server service, it is not exposed publicly and nor is it over `https`. What I can do is rename this helper to something like `bambooServerServiceDns`?

src/main/charts/agent/values.yaml
errcode1202: There are only [recommendations for memory requirements](https://bitbucket.org/dchevell/docker-atlassian-bamboo-base-agent/src/master/) but not CPU. I figured a CPU count of `2` seemed reasonable but maybe not?

src/main/charts/agent/values.yaml
errcode1202: To be honest I think its better left where it is:

- Consistent positioning with the other charts
- Arguably the `image` stanza which comes after `replicaCount` takes priority
- and following on from the previous point so too does the `serviceAccount` stanza

src/main/charts/agent/values.yaml
errcode1202: Okay I think Im going to have to wait until #309 is merged - seems its still in flux.

src/test/java/test/IngressTest.java
errcode1202: Its no different for bamboo, but I noted confluence was testing for it, so in the interests of flexibility I thought making sure had the flexibility too. To be honest Im not sure if this is a real use case but its not exactly like we can stop people from providing different values for these properties in `values.yaml`?

src/main/charts/agent/values.yaml
bianchi2: 2 vcpu is more than enough for the agent itself but there's certainly no standards here, especially when it comes to what kind of jobs the agent will take. In fact, 2 vcpu can be a lot fo ask (as a default), provided that one has 4cpu/16GB workers for example.

bianchi2:
@errcode1202 nice work! I played a bit with a [k8s Operator](https://github.com/bianchi2/bamboo-operator) that launches agents this way based on Bamboo job queue state. It's just a fun project though.

Quick question about agent registration and scaling up/down. Since agent home isn't persisted, it'll be a new agent every time a new pod starts or restarts. Thus, we can have situations when a pod is killed if it exceeds mem cap, evicted, and so on. Also, if you scale up/down this deployment 10 times a day, you'll have 10 offline agents a day and potentially lots of them, if some kind of automation does the scaling like a cron job.

We have seen customers with 1000s of registered agents in the database which is caused by the usage pattern - creating and deleting agents on demand like this. This causes big issues with Bamboo performance since at times Bamboo needs to check all agent's capabilities.

Ideally, when such a stateless agent pod restarts, we need to "unregister" an agent (post shutdown hook). Or, maybe run a k8s job that does that every now and then - delete offline agents that have a particular naming pattern. In both cases though such a job/hook will need api token or username/pass. [My lame attempts to do that](https://github.com/bianchi2/bamboo-operator/blob/f1d0fa31d51dbb61a632966c58d73d633db69007/rest/rest_client.go#L58)

Maybe it's worth adding a note to docs and check with Bamboo team to confirm that such a usage pattern can lead to performance issues over time.

src/main/charts/agent/values.yaml
errcode1202: @bianchi2 2 does feel a bit excessive as a default, I think we can change this to `1` and leave it up to the user of the chart to tune this as they see fit.

errcode1202:
> @errcode1202 nice work! I played a bit with a [k8s Operator](https://github.com/bianchi2/bamboo-operator) that launches agents this way based on Bamboo job queue state. It's just a fun project though.
> 
> Quick question about agent registration and scaling up/down. Since agent home isn't persisted, it'll be a new agent every time a new pod starts or restarts. Thus, we can have situations when a pod is killed if it exceeds mem cap, evicted, and so on. Also, if you scale up/down this deployment 10 times a day, you'll have 10 offline agents a day and potentially lots of them, if some kind of automation does the scaling like a cron job.
> 
> We have seen customers with 1000s of registered agents in the database which is caused by the usage pattern - creating and deleting agents on demand like this. This causes big issues with Bamboo performance since at times Bamboo needs to check all agent's capabilities.
> 
> Ideally, when such a stateless agent pod restarts, we need to "unregister" an agent (post shutdown hook). Or, maybe run a k8s job that does that every now and then - delete offline agents that have a particular naming pattern. In both cases though such a job/hook will need api token or username/pass. [My lame attempts to do that](https://github.com/bianchi2/bamboo-operator/blob/f1d0fa31d51dbb61a632966c58d73d633db69007/rest/rest_client.go#L58)
> 
> Maybe it's worth adding a note to docs and check with Bamboo team to confirm that such a usage pattern can lead to performance issues over time.

Hey @bianchi2 thanks for highlighting this behaviour. This is actually something I had noted, admittedly I assumed that Bamboo would clean out these "stale" agents (at some point) and remove them from its database. I have a demo tonight with the team on the current charts so I can definitely show case this as a potential issue. We have a dedicated engineer for this work so we may be able to add this to Bamboo server itself as an improvement. Failing, and depending on what the Bamboo folks think, we'll take our cues from you ;) and look at implementing a job, or shutdown hook, which I think would be a cleaner option. Very cool [operator](https://github.com/bianchi2/bamboo-operator) by the way!

src/main/charts/agent/Chart.yaml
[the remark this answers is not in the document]
nanux: `artifacthub.io/containsSecurityUpdates: "true"`
I don't think the first release should have "Contain Security Updates" in the changelog.

src/main/charts/agent/templates/_helpers.tpl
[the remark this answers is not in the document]
nanux: I see. In that case, I wouldn't use `baseUrl` name as it has a different established meaning. `bambooServerServiceUrl` is fine?

nanux (approved):

src/main/charts/agent/templates/NOTES.txt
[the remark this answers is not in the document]
tarka: The docker image is published as both "bamboo" and "bamboo-server", with the former being preferred. 

However, there is "Bamboo Server", the (EOLed) Atlassian product, and "the Bamboo server", which is what the agent communicates with. In this case I think the terminology is correct.

tarka (approved):

PR #322, opened 2021-10-07
Make it possible to provision bamboo-agent-capabilities

bianchi2:
This PR introduces a few changes that make it possible to define Bamboo agent capabilities by just dropping **bamboo-capabilities.capabilities** file in the root of a directory where helm upgrade is called.

For some reason, mounting a CM directly to /var/atlassian/application-data/bamboo-agent/bin/bamboo-capabilities.properties in bamboo-agent container makes bin directory empty. The installer jar either can't write to it or smth else non-obvious happens. So, I implemented it as an init container which mounts a cm and copies it to /var/atlassian/application-data/bamboo-agent/bin which is declared as a runtime volume between init and agent containers. The cm gets created only if Helm finds a local file bamboo-capabilities.capabilities.

Test prop file:

```
system.jdk.JDK\ 1.11=/opt/java/openjdk/bin/java
JDK\ 11=/opt/java/openjdk/bin/java
Python=/usr/bin/python3
Python\ 3=/usr/bin/python3
Git=/usr/bin/git
system.builder.ant.Ant=/opt/apache-ant-1.7.1
system.myNewCapability=/my/new/path
```
All capabilities are shows as sources from prop file:

image


errcode1202:
@bianchi2 the way I understand it is, updating the capabilities in this way doesnt actually give the agent these capabilities all its doing is "advertising" its capabilities. For the agent to actually support these, the underlying docker agent base image would need to be updated to support them?

errcode1202:
> > @bianchi2 the way I understand it is, updating the capabilities in this way doesn't actually give the agent these capabilities, all its doing is "advertising" the fact that the agent can run these types of job? For the agent to actually support these, the underlying docker agent base image would also need to be updated to support them?
> 
> @errcode1202 correct. You install tools you need and [declare capabilities for an agent](https://confluence.atlassian.com/bamboo/configuring-remote-agent-capabilities-using-bamboo-capabilities-properties-289276849.html). Job may then have requirements and are only picked up by agents that have those capabilities.
> 
> Also, capabilities/executables are then available in UI when creating job tasks.

Thanks for the clarification @bianchi2. Similar to the my [comment on your Docker-in-Docker PR ](https://github.com/atlassian/data-center-helm-charts/pull/323#issuecomment-939560599) we'll address this as part of the 2nd release for the Bamboo charts.

PR #323, opened 2021-10-07
Make it possible to run jobs in a sidecar docker container

bianchi2:
This PR adds an optional docker-in-docker container to Bamboo remote agent pod to make it possible to run jobs in Docker environment but avoid doing that in an agent container.

A few pre-reqs:
* docker needs to be installed (not started) in the agent image, for example
```
FROM atlassian/bamboo-agent-base:latest
RUN curl -s https://get.docker.com/ | bash
```
* a sidecar container needs to run in a **privileged** mode to make docker-in-docker available.

Security (apart from dind container run in a privileged mode):
* Docker itself can run in secure (default) and non secure mode, and certs dir is mounted into agent container accordingly.

File sharing:
* Xml-data and temp from BVamboo agent home are mounted into both containers so that dind has access to build dir and temp scripts. Docker image version is configurable and should likely be more or less the same version as in agent image

Use cases:
* run jobs in Docker
* use Docker in script tasks
* use Docker type tasks in jobs

Limitations:
* dind container can grow in size as /var/lib/docker grows, so it's recommended to restart agent every now and then or run some jobs that clean up unused images and exited containers
* if this container exceeds mem cap the entire pod will be restarted and since agent home isn't persisted it'll mean a new agent comes up
* not extensively tested :)

image

image

image


errcode1202:
> @errcode1202 of course! I tested this in MiniKube (running jobs in Docker and running docker tasks in agent environment). Here's a short video: https://drive.google.com/file/d/131O1DKZbubJU-ImY28odF23hA9vGZJrW/view
> 
> The only real k8s related limitation is securityContext.privileged: true. If it's forbidden cluster wide, dind container won't start of course. Needless to say, we should mention that dind is definitely not among security best practices in containers world. Other than that, it's just a sidecar container that shares a few volume mounts with agent container.
> 
> Once again, the scenarios that I tested:
> 
> * job run in agent environment: using Docker cli in script tasks
> * job run in agent environment: using Docker tasks
> * job run in docker environment: using any non docker tasks
> * job run in docker environment: using Docker tasks
> 
> The latter is the most interesting use case. **Using docker cli and docker tasks** in a **job that runs in Docker itself**. It is possible by mounting `/var/run/docker.sock` and agent home/temp directory. We'll be looking at docker-in-docker-in-docker (dind itself - job container - docker in job container) :) No changes in Helm chart are required to achieve that, just add an additional /var/run/docker.sock mount should be added to job docker settings. I have tested this as well. Needless to say, the image that you use for your job needs to have docker installed (not started), and /var/run/docker.sock mount will make sure that Docker in dind container is invoked when Bamboo runs an exec into a job container. Having tested this, I can't help posting:
> 
> image

@bianchi2 we have a few priorities on the backlog that we need to get through for the first release, once out we'll tee this up as a priority, so that it can be effectively tested and documented. Xzibit was always ahead of the curve :) so getting this in is a must!

bianchi2:
@errcode1202  of course! No worries. I'll try to keep my fork in "able to merge" state. And like I said, this is some experimental stuff that works, but sure thing a [PBC plugin](https://bitbucket.org/atlassian/per-build-container/src) should do a better job. Nonetheless, didn is doing what it is supposed to do, as well gives some flexibility since you don't need to install stuff in agent image (other than Docker)

PR #328, opened 2021-10-11
Update the way that readiness and startup probes are called

errcode1202:


errcode1202:
> Does this fix a problem?

When viewed through K9s I noticed that without the agents started immediately at the same time. With the update I noticed things behaving as I'd expect, i.e. it took longer for the agents to be recognised as started. This change also makes things consistent with how the other helm charts invoke commands.

PR #333, opened 2021-10-14
[confluence]: add topologySpreadConstraints

janfuhrer:
**What this PR does:**
New feature which alows to add `topologySpreadConstraints` to the Confluence or Synchrony pod.
More infos: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/

I've tested the changes locally with the examples in the values-file:

```yaml
confluence:
  topologySpreadConstraints: 
   - maxSkew: 1
     topologyKey: kubernetes.io/hostname
     whenUnsatisfiable: ScheduleAnyway
     labelSelector:
       matchLabels:
         app.kubernetes.io/name: confluence-synchrony
synchrony:
  enabled: true
  topologySpreadConstraints: 
   - maxSkew: 1
     topologyKey: kubernetes.io/hostname
     whenUnsatisfiable: ScheduleAnyway
     labelSelector:
       matchLabels:
         app.kubernetes.io/name: confluence-synchrony
```

PR #334, opened 2021-10-14
add new feature additionalVolumeClaimTemplates

janfuhrer:
**What this PR does:**
New feature which allows to define additional `volumeClaimTemplates` for the Confluence and Synchrony pod.
This is needed, if we f.e. want to deploy Confluence on NFS, but we [have to use BlockStorage (or everything else instead of NFS) for the lucene-index](https://confluence.atlassian.com/kb/health-check-lucene-index-files-location-873918046.html).

The feature allows to define a list of `additionalVolumeClaimTemplates`. For each `volumeClaimTemplate`, we can define if we want to enable it in Confluence, Synchrony or both of them (use of `enableInConfluence` or `enableInSynchrony`).
It works both, with `volumes.localHome.persistentVolumeClaim.create` enabled or disabled.
Works too with Synchrony.

**Tests:**
I've tested the changes locally with following values (with clustering & 3 replicas, works fine for me) :

```yaml
confluence:
  additionalVolumeMounts:
  - mountPath: /var/atlassian/application-data/confluence/index
    name: index
  clustering:
    enabled: true
volumes:
  additionalVolumeClaimTemplates:
  - enableInConfluence: true
    enableInSynchrony: false
    name: index
    resources:
      requests:
        storage: 1Gi
    storageClassName: block
  localHome:
    persistentVolumeClaim:
      create: true
      resources:
        requests:
          storage: 1Gi
      storageClassName: nfs
  sharedHome:
    persistentVolumeClaim:
      create: true
      resources:
        requests:
          storage: 1Gi
      storageClassName: nfs
```

Let me know what you think or if I need to change anything.


nanux:
Thank you @janfuhrer, I can see how this can be useful in some use cases. It is a similar case to https://github.com/atlassian/data-center-helm-charts/pull/333 PR, we try to maintain feature parity between the product helm charts so it would be great if we can implement it for the other products as well. I can see the `enableInConfluence` and `enableInSynchrony` are fairly specific for Confluence as it is deployed in 2 stateful sets. Is the `additionalVolumeClaimTemplates` useful for Synchrony as well or did you include it for completeness?

janfuhrer:
I only did it for completeness. I think it's enough to do it only for the confluence-statefulset, because the Lucene index is not needed for Synchrony, right?
There is probably a better solution to implement this without the specific `enableInConfluence` and `enableInSynchrony`.

nanux:
I see, thanks for thinking about Synchrony as well but I don't think it is necessary. If we make it available only on product StatefulSet, the implementation will be then consistent across different  Do you think you could remove the `enableInConfluence` and `enableInSynchrony` and use only in `confluence.volumeClaimTemplates`? If you can then use the same approach for other products and again add some unit tests for the change.

Thank you

janfuhrer:
@nanux Yes I'll look at this.

jjeongatl (approved):

nanux (approved):

nghazali (approved):

nghazali:
@janfuhrer , Thanks for the contribution.

PR #341, opened 2021-10-21
Define podAnnotations as template to allow overrides

pathob:
Dear Atlassian team,

similar to pull request #173 that we contributed, we need pod annotations to be able to be overridden dynamically in our Helm charts that are using your Helm charts as sub charts. Currently, Kubernetes cannot detect that changes made to our own files should trigger a redeploy of your applications. With this pull request, we could override your template definitions and add custom checksums so that we can control when your applications should be redeployed.

We would very much appreciate if this feature could be merged.

I also improved the comments for all the template definitions a bit.

Thanks, and please let me know if anything needs to be adjusted.


pathob:
Also see https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

But I saw this PR has conflicts now, will resolve them next week...

nanux:
Hi @pathob, could you please elaborate a bit on the situation when Kubernetes doesn't pick up the changes and doesn't restart the pods? Is this related to your specific setup or is that applicable to the general use of the helm charts?

We have merged this contribution - https://github.com/atlassian/data-center-helm-charts/pull/315 that is using the checksum annotation to make sure the pods are rolled over when the config map definition changes. Is that not sufficient for your case?

pathob:
Hi @nanux , yes, it is more related to our specific setup as we are using your charts as subcharts and want to inject more checksums of our own configmaps.

errcode1202:
Created in house pipeline to verify these changes: https://server-syd-bamboo.internal.atlassian.com/browse/DCD-K8SHELMTEST43-1

errcode1202 (approved):

nghazali (approved):

nanux (approved):

PR #344, opened 2021-11-02
DCKUBE-731: Stabilize CI

nanux:
There were multiple issues causing CI instability. Mainly two main causes of CI failures were:
1. Problematic state of the CI testing cluster that wasn't managed by IaC scripts
2. `Parallel` option for the Bitbucket Data Center

There is a parallel (no pun intended) work on the first issue - we have now cluster managed with code and was recreated from scratch. This cluster has a predictable state and this PR contains some changes that are necessary to utilise the cluster.

The second problem with the `Parallel` option is that the nodes were coming up in the same time and the second pod got stuck with SQL exception:

```
Detail: Key (prop_key)=(last.licensed.user.count) already exists.; nested exception is org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "app_property_pkey"
```

I have switched the option back to `OrderedReady` (the previous default) but users can still switch to parallel to use it with mirror farm.

Note: I was thinking about rebasing the changes to clean up history but I think this documents the debugging that was done and because we are squashing the commits when the PR is merged, I decided to leave it as is.


nanux:
CI build results - https://server-syd-bamboo.internal.atlassian.com/browse/DCD-K8SHELMTEST24-13

src/main/charts/bitbucket/values.yaml
```diff
@@ -635,11 +635,11 @@ bitbucket:
     #
     command: "/shutdown-wait.sh"
 
-  # -- Pod management strategy. Bitbucket can start and destroy multiple nodes at the
-  # same time therefore it can utilize the "Parallel" option.
+  # -- Pod management strategy. Bitbucket Data Center require "OrderedReady" value but for Bitbucket Mirrors you
```
errcode1202, line 638:
```suggestion
  # -- Pod management strategy. Bitbucket Data Center requires the "OrderedReady" value but for Bitbucket Mirrors you
```

tarka (approved):

src/test/scripts/helm_install.sh
```diff
@@ -281,6 +281,8 @@ wait_for_ingress() {
   echo "Waiting for $INGRESS_URI to be ready"
   for (( i=0; i<10; ++i ));
   do
+     echo "checking status: i = ${i}"
```
errcode1202, line 284:
```suggestion
     echo "checking ingress status: i = ${i}"
```

errcode1202: .

errcode1202 (changes requested):

nghazali (approved):

PR #352, opened 2021-11-30
set ActiveProcessorCount based on ...requests.cpu

domruf:
Instead of telling the user to manually update ActiveProcessorCount,
simply add the setting automatically based on
.Values.jira.resources.container.requests.cpu.


src/main/charts/jira/templates/config-jvm.yaml
```diff
@@ -10,6 +10,11 @@ data:
     {{- range .Values.jira.additionalJvmArgs }}
     {{ . }}
     {{- end }}
+    {{- if hasSuffix "m" .Values.jira.resources.container.requests.cpu }}
+    -XX:ActiveProcessorCount={{ div (trimSuffix "m" .Values.jira.resources.container.requests.cpu) 1000 | default 1 }}
+    {{ else }}
+    -XX:ActiveProcessorCount={{ .Values.jira.resources.container.requests.cpu }}
+    {{ end }}
```
nanux, line 17:
```suggestion
    {{- end }}
```
to remove the empty line

nanux:
@domruf, thank you for the improvements. This will make the configuration less error-prone. Would you mind adding the same implementation to the other products as well?

Can you also fix the `test.RequestsTest.sts_resource_requests_and_limits` test please? It would be great to add some additional unit tests to cover this logic.

domruf:
@nanux okay I'll look into it.

nanux (approved):

nghazali (approved):

nanux:
Thanks @domruf !

PR #359, opened 2021-12-15
BAMBK8S-194: Support for unattended setups

errcode1202:
Bamboo `8.1` has a limitation where partial setups are broken. For the purposes of the first Bamboo helm chart `1.0.0` release, unattended setups are non optional and must be used. A bug has been raised against Bamboo itself to address this limitation: 

https://jira.atlassian.com/browse/BAM-21542

Until this issue is fixed unattended setups will be required when deploying Bamboo server via Helm.

### DIFF
--- a/src/main/charts/bamboo/README.md
+++ b/src/main/charts/bamboo/README.md
@@ -1,6 +1,6 @@
 # bamboo
 
-![Version: 0.0.1](https://img.shields.io/badge/Version-0.0.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 8.0.4-jdk11](https://img.shields.io/badge/AppVersion-8.0.4--jdk11-informational?style=flat-square)
+![Version: 0.0.1](https://img.shields.io/badge/Version-0.0.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 8.1.1](https://img.shields.io/badge/AppVersion-8.1.1-informational?style=flat-square)
 
 A chart for installing Bamboo Data Center on Kubernetes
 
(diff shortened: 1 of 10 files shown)


src/main/charts/bamboo/values.yaml
```diff
@@ -506,7 +522,7 @@ bamboo:
     
     # -- How often (in seconds) the Bamboo container readiness probe will run
     #
-    periodSeconds: 5
+    periodSeconds: 10
```
errcode1202, line 525: Bamboo does not like to have its `/status` endpoint hit on a regular basis otherwise issues can occur during setup. Increase this threshold to mitigate this issue being encountered.

src/main/charts/bamboo/values.yaml
```diff
@@ -2,6 +2,14 @@
 # 
 # HEADS UP! 
 #
+# Release 1.0.0 of this chart only support FULL unattended setups.
```
nanux, line 5: So it is not possible to have either an unattended setup or a fully manual setup? I thought we can give the choice to the customers.

nghazali: This will affect terraform project as well. What we are going to do in terraform part is make sure the user has provided all info for unattended setup otherwise we stop them to proceed. 

Do we need to give users choice to do manual setup?

errcode1202: During testing full manual setups at times [exhibited issues](https://hello.atlassian.net/wiki/spaces/DCD/pages/1450841428/Fix+verification+-+testing#Nothing-pre-seeded---working-but-flakey)

Because of this Im not comfortable suggesting this as an approach right now.

src/main/charts/bamboo/README.md
```diff
@@ -63,12 +61,12 @@ Kubernetes: `>=1.19.x-0`
...
 | bamboo.sysadminCredentials.displayNameSecretKey | string | `"displayName"` | The key in the Kubernetes Secret that contains the sysadmin display name |
 | bamboo.sysadminCredentials.emailAddressSecretKey | string | `"emailAddress"` | The key in the Kubernetes Secret that contains the sysadmin email address |
 | bamboo.sysadminCredentials.passwordSecretKey | string | `"password"` | The key in the Kubernetes Secret that contains the sysadmin password |
 | bamboo.sysadminCredentials.secretName | string | `nil` | The secret that contains the admin user information |
 | bamboo.sysadminCredentials.usernameSecretKey | string | `"username"` | The key in the Kubernetes Secret that contains the sysadmin username |
+| bamboo.unattendedSetup | bool | `true` | To skip the setup wizard post deployment set this property to 'true' and ensure values for all 'REQUIRED' and 'UNATTENDED' stanzas  (see banner of this file) have been supplied. For release 1.0.0 this value is by default set to 'true' |
```
nanux, line 69:
```suggestion
| bamboo.unattendedSetup | bool | `true` | To skip the setup wizard post deployment set this property to 'true' and ensure values for all 'REQUIRED' and 'UNATTENDED' stanzas  (see banner of this file) have been supplied. For release 1.0.0 this value is by default set to 'true' and shouldn't be changed. |
```

src/main/charts/bamboo/values.yaml
```diff
@@ -341,9 +350,13 @@ ingress:
...
 
-  # -- Supply the base URL for the installation.
+  # -- To skip the setup wizard post deployment set this property to 'true' and
+  # ensure values for all 'REQUIRED' and 'UNATTENDED' stanzas 
+  # (see banner of this file) have been supplied.
+  #
+  # For release 1.0.0 this value is by default set to 'true'
   #
-  baseUrl:
+  unattendedSetup: true
```
nanux, line 359: What is the play after we will support a partial unattended setup? Will this be `true` and then just customer provides less properties?

What will happen if this is set to `false`? Just from curioisity.

src/main/charts/bamboo/values.yaml
```diff
@@ -376,24 +389,27 @@ bamboo:
   #
   disableAgentAuth: false
 
-
-  # -- The license to provide to the Bamboo nodes (optional). If you
-  # have an existing license it can be provided to the server up-front
-  # to skip the configuration screen on first run.
+  # UNATTENDED
```
nanux, line 392:
```suggestion
  # UNATTENDED SETUP
```

nanux (approved):

nghazali (approved):

src/main/charts/bamboo/values.yaml
errcode1202: So:

- `Full` unattended setup this should be `true`
- `Partial` unattended setup this should be `true`
- Attended setup this should be `false`

Im not a fan of this, and I'd like the fix being addressed by the Bamboo team to negate the need for this.

PR #364, opened 2021-12-22
add new feature podLabels

janfuhrer:
**What this PR does:**
New feature which allows to define additional `podLabels` for all products.
This is necessary because `additionalLabels` is only used for statefulset/deployment, but not for the labels of the pods.


yzha645:
Thank you @janfuhrer for raising this PR! 

While we're working on reviewing it, do you mind adding some unit tests for it? Cheers.

janfuhrer:
I have added some unit tests, thanks for the review.

src/main/charts/jira/templates/statefulset.yaml
```diff
@@ -22,6 +22,9 @@ spec:
         {{- end }}
       labels:
         {{- include "jira.selectorLabels" . | nindent 8 }}
+        {{- with .Values.podLabels }}
```
yzha645, line 25:
```suggestion
        {{- with .Values.additionalLabels }}
```
Will it fit your use case if we re-use `additionalLabels` here instead of creating new `podLabels`? Introducing another field for label seems a bit duplicated to me.

yzha645: And do you mind moving it into named templates? A named template will give more flexibility to customize/override certain values. Something similar to this PR https://github.com/atlassian/data-center-helm-charts/pull/341

janfuhrer: change into named templates: 5bd5bbfe97b0679f9a3236413e22ba86701ab60a

nghazali (approved):

nghazali:
@janfuhrer , Thanks for the contribution. :)

src/main/charts/jira/templates/statefulset.yaml
yzha645: yup, fair enough! Thanks, Jan~

yzha645 (approved):

PR #369, opened 2022-01-11
Changed deprecated annotation for ingress class name

reinventit:
The api for (for default) ingress class is changed since Kubernetes 1.18

See:
https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation

This is an issue when using a Kubernetes setup with multiple ingress controllers. While most ingress controllers were supporting the deprecated style using the annotation 'kubernetes.io/ingress.class' for some time; recent implementation for at least Nginx stopped supporting it.

I therefore propose a change to the new notation in favour of the old annotation. The old annotation is not needed anymore for the Helm chart already only supports Kubernetes 1.19+

### DIFF
--- a/src/main/charts/jira/templates/ingress.yaml
+++ b/src/main/charts/jira/templates/ingress.yaml
@@ -6,7 +6,6 @@ metadata:
   labels:
     {{- include "jira.labels" . | nindent 4 }}
   annotations:
-    "kubernetes.io/ingress.class": {{ default "nginx" .Values.ingress.className }}
     {{ if .Values.ingress.nginx }}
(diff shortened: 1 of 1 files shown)


yzha645:
Hi @reinventit , thanks for your contribution! 

To get all checks green, can you please fix the unit test? 

And also do you mind adding the change to other products as well to keep it consistent?

tarka:
Looking more closely at this PR it is currently incomplete. The change only applies to Jira, but would also be required for the other products, and currently lacks testing. Additionally, given the complexity of K8S ingress we should probably do more testing against test clusters. I'm closing this for now, but have raised DCD-1524 internally to make this change.

PR #371, opened 2022-01-17
[bitbucket]: fix templating of PLUGIN_SSH_BASEURL with custom port

janfuhrer:
**What this PR does:**
Fix templating the environment variable `PLUGIN_SSH_BASEURL` with a custom port other than `22`.

**Current behaviour**
If you use following values, the value of the environment variable `PLUGIN_SSH_BASEURL` is still `ssh://testhost-ssh/` without the custom port number.

_values.yml_
```yaml
ingress:
  host: test
bitbucket:
  sshService:
    enabled: true
    port: 7999
    host: testhost-ssh
```

**Exepted behaviour**
With the updated function, the result is the expected string `ssh://testhost-ssh:7999/`.
I removed the "default" templating section, because a default is already defined in the values file (https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bitbucket/values.yaml#L462)

### DIFF
--- a/src/main/charts/bitbucket/templates/statefulset.yaml
+++ b/src/main/charts/bitbucket/templates/statefulset.yaml
@@ -132,7 +132,7 @@ spec:
             {{- end }}
             {{- if and .Values.bitbucket.sshService.enabled (not (empty .Values.bitbucket.sshService.host)) }}
             - name: PLUGIN_SSH_BASEURL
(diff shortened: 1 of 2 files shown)


yzha645:
Thanks for the PR, @janfuhrer ! 

Do you mind adding a unit test for this scenario as well?

janfuhrer:
Hi @yzha645
No problem, it's done.

yzha645 (approved):

jjeongatl (approved):

nghazali (approved):

nghazali:
Thank you @janfuhrer for the contribution.

PR #380, opened 2022-02-10
Release/1.2.0 products excluding bamboo

yzha645:
This PR bumps chart version to 1.2.0 for products other than Bamboo.


yzha645:
Will create a unified release for all products.

PR #383, opened 2022-02-10
Add Pull Request template

nanux:
This is a first draft open for discussion what we want to have in the PR template:

Screen Shot 2022-02-10 at 5 02 52 pm

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,8 @@
+## Pull request description
+
+_Provide description for the PR_
+
+## Checklist
+- [ ] I have added unit tests
+- [ ] I have applied the change to all applicable products
+- [ ] I have added the change description to the `changelog.md` and `Chart.yaml` files


.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+## Pull request description
+
+_Provide description for the PR_
+
+## Checklist
+- [ ] I have added unit tests
+- [ ] I have applied the change to all applicable products
+- [ ] I have added the change description to the `changelog.md` and `Chart.yaml` files
```
nghazali, line 8: Shouldn't we do this part ourself consistently instead of asking the contributors for change log notes. Some changes might be non-functional or minor which we don't want to mention in change log.

We currently do it in each release and not in each tasks or stories, and the release manager decides if include the changes and how to do it. We may should keep this practice for contributors too.

nanux: I wanted to have different PR template for forks and our internal one but this can be done only by specifying a parameter when opening PR.

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

I think you are right, we can split the template into two sections (general tasks, internal tasks - skip for fork PRs)

nghazali (approved):

Issue #379, opened 2022-02-09
Shared home emptyDir prevents k8s maintenance, autoscaling, etc.


konrad-garus:
### Suggestion

I want to run a single node of Bitbucket. I don't need shared home, I'm not gonna use it anyway.

The chart only seems to have 2 options for shared-home: `emptyDir` or `persistentVolumeClaim`. Even if the volume is not even going to be mounted, it will still be created.

This is causing some issues later, as k8s autoscaler won't evict a pod that has local storage.

It looks like a better solution could be:
* Support `persistentVolumeClaim` for multi-node deployments.
* Do not create the shared-home volume at all for single-node deployments. What's the point of an unused `emptyDir`?

If I'm missing something and there's still some value in that `emptyDir`, perhaps it would be possible to add a switch that lets us avoid creating this volume?

### Product

Bitbucket

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jjeongatl:
Hey @konrad-garus. Thanks for raising the issue.

For the bitbucket, mounting `persistentVolume` to shared home is mandatory (regardless of number of bitbucket nodes) as it will store key contents. 

`emptyDir` will mount VM storage to pod however this is **ephemeral** volume therefore not recommended for production clusters. 

For this case, you can use custom `persistentVolume` for shared home.

- If you are already using it for local home, you can re-use this for share home by provide it as `customVolume` to shared home in values.yaml file.
- If you are not using `persistentVolume` for local home, manually create a PVC first, and provide it as `customVolume` to both local home and shared home in values.yaml file.

konrad-garus:
I find this a little strange.

When I just generate the template with no variables (`helm template atlassian-data-center/bitbucket`), I end up with `emptyDir` for `shared-home` in my StatefulSet `volumes` section, but it's not actually mounted anywhere. So it looks like this:

```
volumeMounts:
  - name: local-home
    mountPath: "/var/atlassian/application-data/bitbucket"

# ...

volumes:
  - name: local-home
    emptyDir: {}
  - name: shared-home
    emptyDir: {}
```

Looking at statefulset.yaml, it does seem to all be conditional:

```
{{- if or .Values.volumes.sharedHome.persistentVolumeClaim.create .Values.volumes.sharedHome.customVolume }}
- name: BITBUCKET_SHARED_HOME
  value: {{ .Values.volumes.sharedHome.mountPath | quote }}
{{- end }}
```

```
volumeMounts:
[...]

PR #386, opened 2022-02-16
DCD-1461: Remove emptyDir from Bitbucket share home volume options

jjeongatl:
When PVC is not provided, the Bitbucket chart chooses `emptyDir` as Bitbucket share home volume.
 I changed the `helper.tpl` to not to create emptyDir for shared home. This will allow share home to be mounted on subpath (/shared) of local home.

See the test result below.

Screen Shot 2022-02-16 at 2 39 08 pm
PVC for local home. No PVC for shared home

Screen Shot 2022-02-16 at 2 38 49 pm
Confirmed that share home uses local home sub directory 

#379


errcode1202:
@jjeongatl just wondering what the latest is on this PR?

jjeongatl:
CI https://server-syd-bamboo.internal.atlassian.com/browse/DCD-K8SHELMTEST76-BBSAWSDEFAULTS-6/log

jjeongatl:
This change is for single bitbucket node user. It might be worth mentioning this in docs

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -61,6 +61,9 @@ volumes:
 
 As explained above, this default configuration is suitable only for evaluation or testing purposes. Proper volume management needs to be configured.
 
+!!!info "Bitbucket default configuration exception"
+    In case of Bitbucket, `emptyDir` will not be created as a `shared-home` default volume. This is to utilise a Bitbucket's feature which allows `local-home`'s subpath (`<LOCAL_HOME_DIRECTORY>/shared`) to be `share-home` directory when `share-home` volume is not provided. Note that this is for single bitbucket node user only and using persistence volume for `local-home` is highly recommended for a production environment.
```
errcode1202, line 65: I think this could be a bit more succinct, maybe something like:

```suggestion
For single node Bitbucket deployments, if no `shared-home` volume is defined, then a subpath of `local-home` will automatically be used for this purpose, namely: `<LOCAL_HOME_DIRECTORY>/shared`. This is behaviour specific to Bitbucket itself and is not orchestrated via the Helm chart.
```

errcode1202 (changes requested):

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -61,6 +61,9 @@ volumes:
 
 As explained above, this default configuration is suitable only for evaluation or testing purposes. Proper volume management needs to be configured.
 
+!!!info "Bitbucket default configuration exception"
```
errcode1202, line 64: Maybe:

```suggestion
!!!info "Bitbucket default `shared-home` location"
```

errcode1202 (approved):

docs/docs/userguide/CONFIGURATION.md
nghazali: Thanks @jjeongatl , I see you have documented the use case is only for single node. Technically, we can enforce replica count to 1 if user set it to more than one as well as an action. However, I think this is not a high priority as we have the documentation and leave it to to decide.

nghazali (approved):

PR #387, opened 2022-02-22
Update ingress.yaml

rcabrerizo:


src/main/charts/jira/templates/ingress.yaml
```diff
@@ -29,11 +29,11 @@ spec:
     - host: {{ .Values.ingress.host }}
       http:
         paths:
-          - path: {{ include "jira.ingressPath" . }}
+          - path: {{ include "jira.ingress.path" . }}
```
nghazali, line 32: Hey @rcabrerizo , would you please explain the reason for this change?

nanux: We currently have `.Values.ingress.path` and also defined template `jira.ingressPath` in `_helpers.tpl` which is used here.
@rcabrerizo I think the previous value was correct. Did you hit any issues with ingress that lead you to create a PR?

nanux:
I am closing the PR due to inactivity. If you have additional information, please add it to the PR and reopen it.

PR #389, opened 2022-02-23
Make pod securityContext optional.

in-journey:
Currently, helm charts enforce securityContext, most notably fsGroup is set to 2001. For instances with lots of files this is problematic due to bug https://github.com/kubernetes/kubernetes/issues/69699 (issue is fixed, but will reach GA in 1.23).

Furthermore, as you already found out, setting fsGroup doesn't work for NFS, so this step can also be skipped for that kind of storage. From my personal research I also noticed fsGroup doesn't work well with samba shares, but this needs more through testing.

This PR makes setting whole securityContext optional for Jira and main goal here is to workaround the K8s issue. By default it will still apply securityContext.

If you like it then I can apply the change for the rest of applications.

I'll appreciate if you approve it quickly, as we want to migrate Jira & Confluence to K8s, but in current state mounting shared disk can take 40 minutes for us. I can imagine instances with more files taking even more time to mount a disk.

### DIFF
--- a/src/main/charts/jira/values.yaml
+++ b/src/main/charts/jira/values.yaml
@@ -392,7 +392,7 @@ jira:
   securityContext:
 
     # -- The GID used by the Jira docker image
-    # If not supplied, will default to 2001 if securityContextEnables is set to true.
+    # GID will default to 2001 if not supplied and securityContextEnables is set to true.
     # This is intended to ensure that the shared-home volume is group-writeable by the GID used by the Jira container.
     # However, this doesn't appear to work for NFS volumes due to a K8s bug: https://github.com/kubernetes/examples/issues/260
     #
--- a/src/main/charts/jira/values.yaml
+++ b/src/main/charts/jira/values.yaml
@@ -385,14 +385,14 @@ jira:
   # Do not populate when deploying to OpenShift, unless anyuid policy is attached to a service account.
   #
 
-  # -- Wether to apply security context to pod.
+  # -- Whether to apply security context to pod.
   #
   securityContextEnabled: true
 
   securityContext:
 
     # -- The GID used by the Jira docker image
-    # GID will default to 2001 if not supplied and securityContextEnables is set to true.
+    # GID will default to 2001 if not supplied and securityContextEnabled is set to true.
     # This is intended to ensure that the shared-home volume is group-writeable by the GID used by the Jira container.
     # However, this doesn't appear to work for NFS volumes due to a K8s bug: https://github.com/kubernetes/examples/issues/260
     #
--- a/src/main/charts/jira/values.yaml
+++ b/src/main/charts/jira/values.yaml
@@ -384,10 +384,15 @@ jira:
   # https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
   # Do not populate when deploying to OpenShift, unless anyuid policy is attached to a service account.
   #
+
+  # -- Wether to apply security context to pod.
+  #
+  securityContextEnabled: true
+
   securityContext:
 
     # -- The GID used by the Jira docker image
-    # If not supplied, will default to 2001
+    # If not supplied, will default to 2001 if securityContextEnables is set to true.
     # This is intended to ensure that the shared-home volume is group-writeable by the GID used by the Jira container.
     # However, this doesn't appear to work for NFS volumes due to a K8s bug: https://github.com/kubernetes/examples/issues/260
     #
--- a/src/main/charts/bamboo-agent/templates/deployment-agent.yaml
+++ b/src/main/charts/bamboo-agent/templates/deployment-agent.yaml
@@ -20,6 +20,7 @@ spec:
     spec:
       serviceAccountName: {{ include "agent.serviceAccountName" . }}
       terminationGracePeriodSeconds: {{ .Values.agent.shutdown.terminationGracePeriodSeconds }}
+      {{- if .Values.agent.securityContextEnabled }}
       {{- with .Values.agent.securityContext }}
       securityContext:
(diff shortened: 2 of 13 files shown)


bianchi2:
@in-journey thanks for your contribution. Another use case is running Atlassian DC products on OpenShift where you need to allow service account to use any fsGroup.

I did introduce flexibility to the charts to enable or disable security context - https://github.com/atlassian/data-center-helm-charts/pull/56 but it looks like these changes were then overwritten by this PR https://github.com/atlassian/data-center-helm-charts/pull/309 where fsGroup is still unavoidable. 

+1 for being able to skip it entirely

cc @errcode1202 @nanux

in-journey:
I've added option for all other applications. I hope I didn't miss anything.

I didn't test it, I've only checked the output of `helm --dry-run` command and it seems to be correct.

Can we have any estimation on this, when this will be implemented?

errcode1202:
Hi @in-journey,

Apologies for the delay. We'll look into getting this reviewed by the end of this week.

src/main/charts/jira/values.yaml
```diff
@@ -384,10 +384,15 @@ jira:
   # https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
   # Do not populate when deploying to OpenShift, unless anyuid policy is attached to a service account.
   #
+
+  # -- Whether to apply security context to pod.
+  #
+  securityContextEnabled: true
```
errcode1202, line 390: Im thinking, would it make more sense to rename this property to `enabled` and move it into the `securityContext` stanza itself? This would then also follow the current pattern that is used, for instance; `ingress.create` & `nfsPermissionFixer.enabled` etc.

errcode1202 (changes requested):

src/main/charts/jira/values.yaml
in-journey: I wanted to do that in first place, but that already exists and for my purpose I can't use it. The if logic currently implemented will force the default value to show up, see the code:
https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/jira/templates/statefulset.yaml#L29-L37

If I remember correctly, it was the else part (lines 32 and 33) that blocked me from reworking this in order to keep the backward compatibility. Though it may be okay for you to change from `else` to `else if .enabled`. I don't really know what is the use case here, plus it is deprecated, so I basically decided to stay away from it.

If you decide to change the `else` into `else if .enabled`, then also the `if not .fsGroup` have to be handled properly, but that seems easy.

bordenit:
Can confirm securityContext is a real issue. Portworx vendor support has recommended that we do not set securityContext for the sharedHome (Aka. SharedV4 volume). However, during upgrades we seem to need it.

src/main/charts/jira/values.yaml
errcode1202: @in-journey apologies for the late response on this. Let me take a look today.

errcode1202:
Initial acceptance test run:
https://server-syd-bamboo.internal.atlassian.com/browse/DCD-K8SHELMTEST74-1

errcode1202 (approved):

in-journey:
> Hi @in-journey,
> 
> Apologies for the delay. It would be great if you could add a unit test(s) for this new logic to [SecurityContextTest.java](https://github.com/atlassian/data-center-helm-charts/blob/main/src/test/java/test/SecurityContextTest.java). In the meantime we'll look into getting this reviewed by the end of the week.

Woah, sorry, I didn't see this one. I could miss the edit... anyway, I don't write in java so I couldn't do much anyway.

Thanks for working on improving and fixing this one :)

By the way I can confirm that changes proposed here work for 1.3.0 charts (initially I wrote this based on 1.2.0 charts) - I successfully patched Jira & Confluence to newer version.

tarka (approved):

nghazali (approved):

nanux (approved):

errcode1202:
> 

Not a problem @in-journey. Reviewed and merged 🚀 .

errcode1202:
> 



> > Hi @in-journey,
> > Apologies for the delay. It would be great if you could add a unit test(s) for this new logic to [SecurityContextTest.java](https://github.com/atlassian/data-center-helm-charts/blob/main/src/test/java/test/SecurityContextTest.java). In the meantime we'll look into getting this reviewed by the end of the week.
> 
> Woah, sorry, I didn't see this one. I could miss the edit... anyway, I don't write in java so I couldn't do much anyway.
> 
> Thanks for working on improving and fixing this one :)
> 
> By the way I can confirm that changes proposed here work for 1.3.0 charts (initially I wrote this based on 1.2.0 charts) - I successfully patched Jira & Confluence to newer version.


Not a problem @in-journey. Reviewed and merged 🚀 .

Wicaeed:
😕  I'm a bit confused as to how I can go about testing these changes? I have a need to try and disable the SecurityContext setting for Jira, but I don't see any new Helm chart versions available as jira-1.3.0 is the latest, and that was publiushed back in March

jjeongatl:
> 😕 I'm a bit confused as to how I can go about testing these changes? I have a need to try and disable the SecurityContext setting for Jira, but I don't see any new Helm chart versions available as jira-1.3.0 is the latest, and that was publiushed back in March

Hi @Wicaeed, probably it is too late to reply but we aim to release new version every 2-4 weeks. this feature was added in 1.4.0. Also If you want to test the main branch without waiting for next release you can always pull the codes to your local machine and deploy charts manually. Refer to the [official helm chart documentation](https://helm.sh/docs/helm/helm_install/) for how to install chart from local directory

PR #419, opened 2022-06-14
CLIP-1553: Use custom port for product server

nanux:
## Pull request description

* Fix the use of the custom ports for the products
* Add unit tests
* Bitbucket is not using Tomcat, therefore it needs to use a different env. property (`server.port`).

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/test/java/test/CustomPortTest.java
+++ b/src/test/java/test/CustomPortTest.java
@@ -1,0 +1,47 @@
+package test;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+import test.helm.Helm;
+import test.model.Container;
+import test.model.Product;
+
+import java.util.Map;
+
+import static test.jackson.JsonNodeAssert.assertThat;
+
+class CustomPortTest {
+    private Helm helm;
+
+    @BeforeEach
+    void initHelm(TestInfo testInfo) {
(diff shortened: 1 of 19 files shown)


nanux:
End to end test - https://server-syd-bamboo.internal.atlassian.com/browse/DCD-K8SHELMTEST87

src/test/java/test/CustomPortTest.java
```diff
@@ -0,0 +1,47 @@
...
+import static test.jackson.JsonNodeAssert.assertThat;
+
+class CustomPortTest {
+    private Helm helm;
+
+    @BeforeEach
+    void initHelm(TestInfo testInfo) {
+        helm = new Helm(testInfo);
+    }
+
```
nghazali, line 22: Is not crucial but I think maybe good to have unit test for default values for `ATL_TOMCAT_PORT` in different products

nghazali (approved):
Left an optional comment.

src/test/java/test/CustomPortTest.java
nanux: The default values are covered usually by the `HelmOutputComparisonTest` and using the expected `output.yaml` files - this is an example:
https://github.com/atlassian/data-center-helm-charts/pull/419/files#diff-63ea51911cc0f4498740fb7f8ee9bba6524a8c59bf6378ae51489847c9178444R268

nanux:
Passing build with the latest changes - https://server-syd-bamboo.internal.atlassian.com/browse/DCD-K8SHELMTEST87-2

jjeongatl (approved):

PR #420, opened 2022-06-20
synchrony path fix

### LINKED ISSUE #225: Synchrony Ingress Path Incorrect
The ingress path for synchrony is incorrect.           

```
- path: "{{ trimSuffix "/" .Values.ingress.path }}/synchrony"
```
This should not rely on the .Values.ingress.path as it then nests the synchrony path inside the Confluence contextPath and/or puts synchrony at the same base path, which usually users have to authenticate to with SSO. I fixed it manually by just changing this path to /synchrony in the ingress, as I already have a context path for Confluence.

The path in the ingress should be changed to /synchrony if Confluence is using a contextPath already. But, ideally the ingress for Synchrony should be broken down in the same way the ingress is for Confluence where a host and a path are both parameters. There will also be a different requirement for ingress depending on whether the user uses or does not use a contextPath with Confluence.

User Case 1 - User uses a contextPath with Confluence
- The same ingress can be used for both Confluence and Synchrony by putting both Confluence and Synchrony at different contextPaths. Confluence would route to {{ .Values.ingress.host }}/confluence and and Synchrony would route to {{ .Values.ingress.host }}/synchrony for example.

Use Case 2 - User does not use a contextPath with Confluence
- The same ingress can't be used for both Confluence and Synchrony as a different ingress will be used for a second host used by Synchronys ingress to avoid SSO collisions with the baseURL. Synchrony would route to {{ .Values.sychrony.ingress.host }} and Confluence would route to  {{ .Values.ingress.host }} where they don't have any chance of colliding. 

This would create a conditional requirement to create a second ingress specifically for synchrony if a contextPath is not used as well.

I can work on a PR if you think creating a conditional for second ingress and breaking down synchrony into host and path options is feasible.

bordenit:
## Pull request description
Closes #225 
Allows synchrony to work with or without context path that is driven from the ingress.host used for the confluence application. The context path for synchrony would always be /synchrony and that context path would be reserved for synchrony. I think it removes some of the confusion in the values.yaml around the ingress hosts and context paths and also a lot of the cross-walking of the configuration. There doesn't appear to be much benefit to have the ingress host for synchrony to be configurable as it is only for collaborative editing. A user does not necessarily route to this url directly to use collaborative editing. It just happens in the background. In most cases if the network/web server path to ingress.host is open, then the network path to synchrony will also be open. Currently, the synchrony url does not work with a context path due to incorrect value being placed in the ingress for the synchrony context path.

## Checklist
- [ x ] I have added unit tests
- [ x ] I have applied the change to all applicable products
- [ x ] I have added the change description to the `changelog.md` and `Chart.yaml` files

### DIFF
--- a/src/main/charts/confluence/templates/_helpers.tpl
+++ b/src/main/charts/confluence/templates/_helpers.tpl
@@ -128,11 +128,15 @@ Pod labels
 
 {{- define "confluence.sysprop.synchronyServiceUrl" -}}
 {{- if .Values.synchrony.enabled -}}
--Dsynchrony.service.url={{ .Values.synchrony.ingressUrl }}/v1
+    {{- if .Values.ingress.https -}}
+    -Dsynchrony.service.url=https://{{ .Values.ingress.host }}/synchrony/v1
+    {{- else }}
+    -Dsynchrony.service.url=http://{{ .Values.ingress.host }}/synchrony/v1
+    {{- end }}
 {{- else -}}
 -Dsynchrony.btf.disabled=true
 {{- end -}}
-{{- end }}
+{{- end -}}
 
 {{/*
 Create default value for ingress path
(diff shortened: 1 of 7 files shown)


nghazali:
Thanks you @bordenit for the contribution. I merge this to a branch in our repo to run more tests and will create a new PR to merge against the master branch.

PR #422, opened 2022-06-30
CLIP-1413: Fix synchrony ingress path

nghazali:
* synchrony path fix
* This PR is a contribution from community (https://github.com/atlassian/data-center-helm-charts/pull/420)
Merging to CLIP-1413-fix-synchrony-ingress-path branch to run e2e tests

## Pull request description

Allows synchrony to work with or without context path that is driven from the ingress.host used for the confluence application. The context path for synchrony would always be /synchrony and that context path would be reserved for synchrony. I think it removes some of the confusion in the values.yaml around the ingress hosts and context paths and also a lot of the cross-walking of the configuration. There doesn't appear to be much benefit to have the ingress host for synchrony to be configurable as it is only for collaborative editing. A user does not necessarily route to this url directly to use collaborative editing. It just happens in the background. In most cases if the network/web server path to ingress.host is open, then the network path to synchrony will also be open. Currently, the synchrony url does not work with a context path due to incorrect value being placed in the ingress for the synchrony context path.

## Checklist
- [x] I have added unit tests

### DIFF
--- a/.github/workflows/e2e-tf-deployment.yaml
+++ b/.github/workflows/e2e-tf-deployment.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: Pin Kubectl version
         uses: azure/setup-kubectl@v2.0
         with:
-          version: 'v1.23.6'
+          version: 'v1.24.0'
 
       - name: Install Helm
         uses: azure/setup-helm@v1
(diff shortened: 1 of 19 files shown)


nanux:
The changes are looking good. Thanks @bordenit for the contribution and @nghazali for shepherding the change.

@nghazali could you please add the change description to the `changelog.md` and `Chart.yaml` files.

nanux (changes requested):
@nghazali could you please add the change description to the `changelog.md` and `Chart.yaml` files.

nghazali:
Hi @nanux , the required changes are applied now.

nanux (approved):

jjeongatl (approved):

PR #432, opened 2022-07-15
CLIP-1611: Fix the release problem by pin the docker version of quay.io/helmpack…

nghazali:
…/chart-releaser

Release process broken because the latest docker version of the `quay.io/helmpack/chart-releaser` deprecated `--charts-repo` and end up with an error. Pinning to the older version fixes the issue. 

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/src/main/scripts/generate_chart_repo.sh
+++ b/src/main/scripts/generate_chart_repo.sh
@@ -37,7 +37,7 @@ docker run --user "$(id -u):$(id -g)" \
   -v "$(pwd)/$PACKAGE_DIR:/releases" \
   --entrypoint cr \
   --rm \
-  quay.io/helmpack/chart-releaser \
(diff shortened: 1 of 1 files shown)


nanux (approved):
Would be good to upgrade to `chart-releaser:1.4.0`. Seems that the `charts-repo` was dropped and shouldn't be necessary anymore. 

We don't need to do it as part of this PR.

jjeongatl:
> Would be good to upgrade to `chart-releaser:1.4.0`. Seems that the `charts-repo` was dropped and shouldn't be necessary anymore.
> 
> We don't need to do it as part of this PR.

I created follow up ticket [CLIP-1612](https://bulldog.internal.atlassian.com/browse/CLIP-1612)

jjeongatl (approved):

PR #442, opened 2022-08-17
CLIP-1636: Fix the setup-go github action

nghazali:
## Pull request description

Update setup-go action to the latest as it became flaky.

## Checklist
- [ ] I have added unit tests


nghazali:
Adding `e2e label to this PR end up to run e2e test using `main` branch which result a failure because the fix still is not merged to main. 
To prove that the test is fixing the problem I ran e2e test manually through the action workflow.

nanux (approved):

Issue #398, opened 2022-03-23
[Suggestion] - Allow the option to add annotations for Standalone synchrony service


athira-1142:
### Suggestion

We have to add annotations on the services to enable session affinity and create a health check (backend configuration). At present, we can add annotations through values.yaml only on confluence service. For standalone synchrony service, we have to add annotations after the application is deployed. 

It would be great if there is an option to add annotations on Synchrony service in values.yaml as we can manage everything through Helm.

### Product

Confluence

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

errcode1202:
Hi @athira-1142,

Thankyou for the suggestion. For context, and so we can understand better what you need, would you mind creating a PR with the changes and we can use that then as a talking point as a potential addition to `main`.

DarthKoax:
This is a feature that we have discovered would be very helpful for us. We have run into the issue also with our build where Synchrony's health check and stickiness are configured incorrectly due to inheriting from our confluence ingress annotations. 

This is recommended by Atlassian to have stickiness on for confluence and off for synchrony, along with health check pointing to /status for confluence and /sychrony/heartbeat for synchrony. 
https://confluence.atlassian.com/confkb/how-to-configure-amazon-web-service-application-load-balancer-with-confluence-859515417.html

If [athira-1142](https://github.com/athira-1142) has not already started a PR, we may look into doing so.

PR #464, opened 2022-10-19
CLIP-1567: Make synchrony pod and service annotations configurable

bianchi2:
This PR makes it possible to redefine:

* pod annotations for synchrony. If undefined, `.Values.podAnnotations` is used
* service annotations for synchrony. If undefined, `.Values.confluence.service.annotations` is used

Fixes https://github.com/atlassian/data-center-helm-charts/issues/398

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/confluence/Chart.yaml
+++ b/src/main/charts/confluence/Chart.yaml
@@ -25,7 +25,7 @@ annotations:
       links:
       - name: Github PR
         url: https://github.com/atlassian/data-center-helm-charts/pull/451
-    - kind: changed
+    - kind: added
       description: Synchrony pod and service annotations can be redefined or use default confluence annotations (#464)
       links:
       - name: Github PR
--- a/src/main/charts/confluence/Chart.yaml
+++ b/src/main/charts/confluence/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: confluence
 description: A chart for installing Confluence Data Center on Kubernetes
 type: application
-version: 1.7.0
+version: 1.7.1
 appVersion: 7.19.2
 kubeVersion: ">=1.19.x-0"
 keywords:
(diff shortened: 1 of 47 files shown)


src/main/charts/confluence/Chart.yaml
```diff
@@ -25,6 +25,11 @@ annotations:
       links:
       - name: Github PR
         url: https://github.com/atlassian/data-center-helm-charts/pull/451
+    - kind: changed
```
nghazali, line 28: I think for this one we should set the kind as `added`?

bianchi2: Thanks, fixed.

src/main/charts/confluence/values.yaml
```diff
@@ -744,6 +744,11 @@ synchrony:
   # This will result in a separate StatefulSet and Service to be created for Synchrony.
   # If disabled, then collaborative editing will be disabled in Confluence.
   enabled: false
+
+  # -- Custom annotations that will be applied to all Synchrony pods.
```
nanux, line 748: Should we mention that if not defined, `confluence.podAnnotations` will be used?

src/main/charts/confluence/templates/statefulset-synchrony.yaml
```diff
@@ -15,7 +15,11 @@ spec:
     metadata:
       annotations:
         checksum/config-jvm: {{ include (print $.Template.BasePath "/synchrony-start-script.yaml") . | sha256sum }}
+        {{- if .Values.synchrony.podAnnotations }}
```
nanux, line 18: When this is evaluated not true - is this false for `.Values.synchrony.podAnnotations == {}`?

src/main/charts/confluence/values.yaml
```diff
@@ -761,6 +766,10 @@ synchrony:
     #
     loadBalancerIP:
 
+    # -- Additional annotations to apply to Synchrony Service
```
nanux, line 769: I don't think they are additional, it is `either or`. If defined, they are used, otherwise Confluence ones are used.

nanux (approved):

nanux:
Approved but left some comments to consider before merging.

src/main/charts/confluence/templates/statefulset-synchrony.yaml
bianchi2: yes, undefined podAnnotations will make `{{- if .Values.synchrony.podAnnotations }}` appear as false

nghazali (dismissed):

nanux (approved):

src/main/charts/confluence/values.yaml
bianchi2: Yes, it's there in the next line

nghazali (dismissed):

jjeongatl (approved):

nanux (approved):

nghazali (approved):

Issue #393, opened 2022-03-18
Standalone Synchrony Pod run as root and securityContext does not exist for Synchrony


athira-1142:
### Suggestion

The standalone Synchrony pods for Confluence run as root, while the Confluence pods can run with a dedicated user.  While ```exec``` into the pod: 
```
root@confluence-synchrony-0:~# ps axu
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   6972  3304 ?        Ss   Mar14   0:00 bash /scripts/start-synchrony.sh
root           6  0.9  3.3 8172900 1116324 ?     Sl   Mar14  59:30 java -Xms1g -Xmx2g -Xss2048k -XX:ActiveProcessorCount=2 -classpath /opt/atlassian/confluence/confluence/WEB-
root       13133  0.0  0.0   7236  4004 pts/0    Ss   20:11   0:00 /bin/bash
root       13178  0.0  0.0   8896  3308 pts/0    R+   20:24   0:00 ps axu
```

There is no securityContext available for standalone Synchrony in the values.yaml. Could you please suggest here?

### Product

Confluence

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #465, opened 2022-10-19
CLIP-1500: Add security context for synchrony

bianchi2:
This PR adds security context for Synchrony (both in pod and container spec).

Fixes https://github.com/atlassian/data-center-helm-charts/issues/393

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/confluence/templates/statefulset-synchrony.yaml
+++ b/src/main/charts/confluence/templates/statefulset-synchrony.yaml
@@ -22,7 +22,6 @@ spec:
     spec:
       serviceAccountName: {{ include "confluence.serviceAccountName" . }}
       terminationGracePeriodSeconds: {{ .Values.synchrony.shutdown.terminationGracePeriodSeconds }}
-      {{- if .Values.synchrony.securityContextEnabled }}
       {{- with .Values.synchrony.securityContext }}
       securityContext:
         {{/* this condition will be removed in v2.0.0 */}}
@@ -34,7 +33,6 @@ spec:
         fsGroup: 2002
         {{- end }}
         {{- end }}
-      {{- end }}
       {{- end }}
       hostAliases:
         {{- include "confluence.additionalHosts" . | nindent 8 }}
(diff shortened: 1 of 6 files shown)


src/main/charts/confluence/templates/statefulset-synchrony.yaml
```diff
@@ -22,6 +22,20 @@ spec:
     spec:
       serviceAccountName: {{ include "confluence.serviceAccountName" . }}
       terminationGracePeriodSeconds: {{ .Values.synchrony.shutdown.terminationGracePeriodSeconds }}
+      {{- if .Values.confluence.securityContextEnabled }}
+      {{- with .Values.confluence.securityContext }}
```
nghazali, line 26: I am wonder if we should add a separated security context for Synchrony or we should re-use the confluence security context for Synchrony

bianchi2: I am not sure using different security contexts for Synchrony and Confluence is the use case we'll see much. That's why I decided to unify them not to bloat values.yaml.

src/main/charts/confluence/templates/statefulset-synchrony.yaml
```diff
@@ -22,6 +22,20 @@ spec:
     spec:
       serviceAccountName: {{ include "confluence.serviceAccountName" . }}
       terminationGracePeriodSeconds: {{ .Values.synchrony.shutdown.terminationGracePeriodSeconds }}
+      {{- if .Values.synchrony.securityContextEnabled }}
+      {{- with .Values.synchrony.securityContext }}
+      securityContext:
+        {{/* this condition will be removed in v2.0.0 */}}
```
nanux, line 28: I don't think we need the condition here. We never had this defined for Synchrony statefulset so providing backward compatibility is not necessary.

src/main/charts/confluence/templates/statefulset-synchrony.yaml
nanux: Thanks, @bianchi2. Based on our previous experience it is better to provide complete customizability for the user as it is hard to predict how they will use the helm charts. I agree that adding this to the helm chart is a good idea. In general, the guiding principle is to provide a high degree of flexibility and don't worry about the size of `values.yaml` file.

nanux (changes requested):

src/main/charts/confluence/templates/statefulset-synchrony.yaml
bianchi2: Thanks @nanux I wasn't quite sure why this condition was there, so decided not to delete it. Fixed now.

nanux (approved):

tarka (approved):

nghazali (approved):

PR #468, opened 2022-10-21
Pin Python version to 3.9.14

bianchi2:
See: https://github.com/atlassian-labs/data-center-terraform/pull/270

Pinning python version to make sure pythonpath is correct (required to execute aws_clan.py script)


bianchi2:
Merging with no approvals to avoid failing builds

PR #470, opened 2022-11-02
CLIP-1702: Create Role instead of ClusterRole for Hazelcast Kube client in Confluence and Bitbucket Helm charts

bianchi2:
This PR sets new defaults when it comes to creating RBAC for Hazelcast in Confluence and Bitbucket Helm charts. ClusterRole and ClusterRole binding is only necessary to **list and get nodes** which is https://github.com/hazelcast/hazelcast-kubernetes#zone-aware scenario. Currently, in logs we have:

```
WARN [Catalina-utility-1] [spi.discovery.integration.DiscoveryService] log [10.0.2.138]:5701 [confluence] [3.12.11] Cannot fetch the current zone, ZONE_AWARE feature is disabled
```
i.e. this feature is disabled, and the chart does allow for any Hazelcast config customization (also node aware scenarios aren't being tested)

The PR attempts to introduce as few changes as possible. The new behavior is:

* by default Role and Role binding are created ( with privileges to list and get pods and endpoints)
* name for the Role and RoleBinding reuses what we already have in helpers
* it is possible to avoid creating Role and Role binding and do it manually outside helm chart
* ClusterRole and ClusterRoleBinding are created as before, the only difference is that the default value is false, i.e. customers who really need it should explicitly set it to true in their custom yamls

When upgrading, customers will see the following behavior:

* Role and RoleBinding are created unless `serviceAccount.role.create` is explicitly set to false
* ClusterRole and ClusterRoleBinding will be deleted unless `serviceAccount.clusterRole.create` is explicitly set to true

In fact, the upgrade will silently create Role and ClusterRole (with default settings) which is sufficient for Hazelcast Kube client. To keep changes at minimum, existing ClusterRoleName and ClusterRoleBindingName envs are reused to create namespaced objects

## Checklist
- [x] I have added unit tests


src/main/charts/bitbucket/templates/rbac.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: {{ template "bitbucket.clusterRoleName" . }}
+  labels:
+  {{- include "common.labels.commonLabels" . | nindent 4 }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - endpoints
+      - pods
+{{- if .Values.serviceAccount.clusterRole.create }}
```
nanux, line 18: Can we refer to the hazelcast documentation, why this is needed?

src/main/charts/bitbucket/templates/rbac.yaml
```diff
@@ -0,0 +1,49 @@
+{{- if and (.Values.bitbucket.clustering.enabled) (or .Values.serviceAccount.clusterRole.create .Values.serviceAccount.role.create) -}}
+apiVersion: rbac.authorization.k8s.io/v1
+{{- if .Values.serviceAccount.clusterRole.create }}
+kind: ClusterRole
+{{- else }}
+kind: Role
+{{- end }}
+metadata:
+  name: {{ template "bitbucket.clusterRoleName" . }}
```
nanux, line 9: I know we want to keep the changes limited but what about renaming this to `bitbucket.roleName`?

src/main/charts/confluence/values.yaml
```diff
@@ -81,14 +87,22 @@ serviceAccount:
     # -- Set to 'true' if a ClusterRole should be created, or 'false' if it
     # already exists.
     #
-    create: true
+    create: false
```
nanux, line 90: What will happen to existing installations, if they upgrade to the new version and they didn't override the default, it will delete the ClusterRole and create Role?

nanux (dismissed):

jjeongatl (dismissed):

src/main/charts/confluence/values.yaml
bianchi2: Yes, correct. ClusterRole and ClusterRoleBinding will be deleted, instead Role and RoleBinding created.

src/main/charts/bitbucket/templates/rbac.yaml
bianchi2: Honestly, I don't even understand why we have `name` in there and what's the benefit of letting the user to override it. And yes, my motivation was to reuse as much as possible.

src/main/charts/bitbucket/templates/rbac.yaml
bianchi2: I have dropped the link in the PR description https://github.com/hazelcast/hazelcast-kubernetes#zone-aware or do you mean it'd be nice to have it in the comment too?

src/main/charts/bitbucket/templates/rbac.yaml
bianchi2: @nanux ok, I'm now using `{{ include "common.names.fullname" . }}` for role and rolebinding names to avoid confusion. Like I said before, I do not see much value in introducing name override here.

src/main/charts/bitbucket/templates/rbac.yaml
bianchi2: Updated PR description with a link to Hazelcast readme and a log snippet.

src/main/charts/confluence/values.yaml
bianchi2: Updated PR description

nanux (approved):

jjeongatl (approved):

PR #471, opened 2022-11-03
Clip 1698: Release scripting

tarka:
## Pull request description

## Checklist
- [ ] I have added unit tests
- [/] I have applied the change to all applicable products


src/main/scripts/prepare-release.py
```diff
@@ -0,0 +1,143 @@
...
+        foundfirst = False
+        with open(clfile, 'r') as clfd:
+            for line in clfd:
+                if not foundfirst and re.match(r'^## [0-9]\.[0-9]\.[0-9]', line) != None:
+                    # First version line, inject ours before it
+                    foundfirst = True
+                    now = datetime.now()
+                    os.write(tmpfd, ("## %s\n\n" % version).encode())
+                    os.write(tmpfd, ("**Release date:** %s-%s-%s\n\n" % (now.year, now.month, now.day)).encode())
+                    os.write(tmpfd, '![AppVersion: 9.0.0](https://img.shields.io/static/v1?label=AppVersion&message=9.0.0&color=success&logo=)\n'.encode())
```
bianchi2, line 64: Can we grab actual values from Chart.yaml instead? Both appVersion and Kubernetes version.

bianchi2 (changes requested):

bianchi2 (approved):

src/main/scripts/prepare-release.py
```diff
@@ -0,0 +1,165 @@
...
+products = ["bamboo", "bamboo-agent",
+            "bitbucket", "confluence", "crowd", "jira"]
+prodbase = "src/main/charts"
+
+
+# TODO: As a first pass this currently only filters anything tagged
+# with 'CLIP-nnn'. However there are other options; see
+# https://hello.atlassian.net/wiki/spaces/DCD/pages/2108707663/DACI+Automating+the+Helm+release+process
+def changelog_filter(log_entry):
+    return re.match(r'^\*\s+CLIP-[0-9]+', log_entry) != None
```
nghazali, line 22: Maybe it's good to document in our process that to mention ticket number and use `CLIP-` as port of log as a mandatory step.

bianchi2: I think merging with squash solve the problem of multiple commits in one PR, so merge commit message is smth to take care of (and they can be authored when a PR is merged). As to multiple PRs to the same repo - I don't see an issue her as long as it's a separate PR with its own title and changes.

PR #477, opened 2022-11-13
CLIP-1709: Fix 1 exit code when fetching PRs

bianchi2:
Just a small fixup to handle 1 exit code from grep (if there are no open PRs with a pre-defined title). The workflow was failing when grep returned 1 (i.e. there are no open PRs)

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/.github/workflows/update-lts-tags.yaml
+++ b/.github/workflows/update-lts-tags.yaml
@@ -5,7 +5,7 @@ on:
     - cron: '0 0 * * SUN' # At 00:00 on every Sunday
 
 jobs:
-  tls_check:
+  lts_check:
     runs-on: ubuntu-latest
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/update-lts-tags.yaml
```diff
@@ -39,12 +39,12 @@ jobs:
...
           if [ ${CHANGES} -ne 0 ]; then
-            gh pr list --state open | grep "Update appVersions for DC apps"
-            if [ $? -ne 0 ]; then
+            OPEN_PRS=$(gh pr list --state open | grep "Update appVersions for DC apps" || true)
+            if [[ ${OPEN_PRS} ]]; then
               echo "Not raising PR because there are open PRs to upgrade versions. Merge or close them to let the workflow raise new PRs"
             else
               git checkout -b lts-tag-update-${GITHUB_RUN_ID}
-              git add -A
+              git add src
```
nghazali, line 47: Would you please explain why we need this change?

bianchi2: Must be leftovers from some experiments. This works too since the update script edits files in `src/main/charts` and `src/test`

.github/workflows/update-lts-tags.yaml
```diff
@@ -39,12 +39,12 @@ jobs:
         run: |
           CHANGES=$(git status --porcelain=v1 2>/dev/null | wc -l)
           if [ ${CHANGES} -ne 0 ]; then
-            gh pr list --state open | grep "Update appVersions for DC apps"
-            if [ $? -ne 0 ]; then
+            OPEN_PRS=$(gh pr list --state open | grep "Update appVersions for DC apps" || true)
+            if [[ ${OPEN_PRS} ]]; then
```
nghazali, line 43: and does this change introduce any new functionality? I see this as a refectory for the existing code, right?

.github/workflows/update-lts-tags.yaml
```diff
@@ -5,7 +5,7 @@ on:
     - cron: '0 0 * * SUN' # At 00:00 on every Sunday
 
 jobs:
-  tls_check:
+  lts_check:
```
nghazali, line 8: good catch 👍

.github/workflows/update-lts-tags.yaml
bianchi2: No new functionality. It's just not to fail the script if no existing update appVersion PRs are found. Previously it was relying on exit code of the command, which did not work well.

https://github.com/atlassian/data-center-helm-charts/pull/478 is raised by this workflow

nghazali (approved):

errcode1202 (approved):

Issue #447, opened 2022-09-05
[Suggestion] - Set Validationquery for Oracle


Zarrewitch:
### Suggestion

Currently the validationQuery for database connections is set to _select 1_ by default. But when configuring Oracle this leads to database Log-Errors because Oracle needs _select 1 from dual_ as validation Query. https://confluence.atlassian.com/conf76/surviving-database-connection-closures-1018769693.html

By extending the helm Charts e.g. _helpers.tpl this could be avoided/auto-detected.
`{{ if eq .Values.database.type "oracle" }}
- name: ATL_DB_VALIDATIONQUERY
  value: select 1 from dual
{{ end }}`

### Product

Confluence

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bianchi2:
hi @AlexanderOeser it's a [known issue](https://jira.atlassian.com/browse/SCALE-72) and currently the workaround is to set additional env var in value.yaml (which you know already). I think it'll be addressed soon, or else you're welcome to contribute as well.

jjeongatl:
@AlexanderOeser We created an internal ticket to handle this suggestion

PR #481, opened 2022-11-17
CLIP-1554: Automatically set ATL_DB_VALIDATIONQUERY for oracle db type

bianchi2:
Just a nice little improvement - setting `ATL_DB_VALIDATIONQUERY="select 1 from dual"` if `database.type` is oracle (Confluence) or `oracle10g` (Jira). 

Fixes https://github.com/atlassian/data-center-helm-charts/issues/447

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/confluence/templates/_helpers.tpl
+++ b/src/main/charts/confluence/templates/_helpers.tpl
@@ -477,9 +477,11 @@ volumeClaimTemplates:
       name: {{ . }}
       key: {{ $.Values.database.credentials.passwordSecretKey }}
 {{ end }}
-{{- if and (.Values.database.type) (contains "oracle" .Values.database.type ) }}
+{{- if .Values.database.type }}
+{{- if contains "oracle" .Values.database.type }}
 - name: ATL_DB_VALIDATIONQUERY
   value: "select 1 from dual"
+{{- end }}
 {{- end }}
 {{ end }}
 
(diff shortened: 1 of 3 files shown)


src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -477,6 +477,12 @@ volumeClaimTemplates:
       name: {{ . }}
       key: {{ $.Values.database.credentials.passwordSecretKey }}
 {{ end }}
+{{- if .Values.database.type }}
```
bianchi2, line 480: Interesting finding - with helm 3.7 (configured in GH actions) `{{- if and (.Values.database.type) (contains "oracle" .Values.database.type ) }}` does not work when `database.type` isn't set:

```
Error: template: confluence/templates/statefulset.yaml:129:16: executing "confluence/templates/statefulset.yaml" at <include "confluence.databaseEnvVars" .>: error calling include: template: confluence/templates/_helpers.tpl:480:61: executing "confluence.databaseEnvVars" at <.Values.database.type>: wrong type for value; expected string; got interface {}
```
That's why I has to go with nested condition

yzha645:
Changes made here is good. But in my opinion, the ultimate solution would be unsetting [this value in docker image](https://bitbucket.org/atlassian-docker/docker-atlassian-confluence-server/src/f8a888f1625a22f9062928c6841b62b0807e7c20/config/confluence.cfg.xml.j2#lines-42), and leave it to the DB engine to decide what's the best validation query. See https://hello.atlassian.net/wiki/spaces/~jponting/pages/903617392/Database+Validation+Queries.

PR #482, opened 2022-11-17
Fix additionalJvmArgs parameter for Synchrony

Lac00s:
## Pull request description

Currently a dict is used in the default values for additionalJvmArgs, but the helm charts expects a list. Essentially this is a typo of {} instead of [].

## Checklist
- [x] I have added unit tests


bianchi2 (approved):

errcode1202 (approved):
🙏🏼

PR #487, opened 2022-11-29
Use ingress.https to enable https in ingress

fredrikand:
## Pull request description

ingress.tlsSecretName was used to enable https in the ingress. This forces to use a TLS secret, but in some cases the default tls secret for the
 default ingress can be used.
In the case the default TLS secret for the default ingress controller should be used
 it is not possible to leave ingress.tlsSecretName enpty and simply set ingress.https to true.

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/src/main/charts/bamboo/templates/ingress.yaml
+++ b/src/main/charts/bamboo/templates/ingress.yaml
@@ -18,11 +18,13 @@ metadata:
   {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-{{ if and (.Values.ingress.tlsSecretName) (.Values.ingress.host) }}
+{{ if and (.Values.ingress.https) (.Values.ingress.host) }}
   tls:
     - hosts:
         - {{ .Values.ingress.host }}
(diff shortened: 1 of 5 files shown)


bianchi2:
@fredrikand thanks for your contribution. Indeed, the ingress controller can be configured with the default certificate/key and secretName isn't necessary. I have ran unit tests and all is good.

Would you mind signing the CLA please?

fredrikand:
@bianchi2 I have signed the CLA for my private account fredrikand, using my private email fredrik.and@gmail.com
Unfortunately I had another email configured when I first comitted and the automation seems to have picked up that email as well.
If it is not possible to work around, I can open a new PR with only the correct email.

bianchi2 (approved):

bianchi2:
@fredrikand all good, CLA signed. Thanks

nanux (approved):

fredrikand:
@bianchi2 are we all good for merge here?

bianchi2:
Merged. Thanks @fredrikand

PR #488, opened 2022-11-29
Set variable SEARCH_ENABLED to control Bitbucket starting bundled ela…

fredrikand:
Add variable SEARCH_ENABLED that is recognized by Bitbucket start scripts.

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/src/main/charts/bitbucket/templates/_helpers.tpl
+++ b/src/main/charts/bitbucket/templates/_helpers.tpl
@@ -353,7 +353,7 @@ volumeClaimTemplates:
 
 {{- define "bitbucket.elasticSearchEnvVars" -}}
 {{- if or .Values.bitbucket.elasticSearch.baseUrl .Values.bitbucket.clustering.enabled }}
-- name: ELASTICSEARCH_ENABLED
+- name: SEARCH_ENABLED
   value: "false"
 {{- end }}
 {{ with .Values.bitbucket.elasticSearch.baseUrl }}
--- a/src/test/java/test/ElasticSearchTest.java
+++ b/src/test/java/test/ElasticSearchTest.java
@@ -26,7 +26,7 @@ class ElasticSearchTest {
         resources.getStatefulSet(product.getHelmReleaseName())
                 .getContainer()
                 .getEnv()
-                .assertHasValue("ELASTICSEARCH_ENABLED", "false")
+                .assertHasValue("SEARCH_ENABLED", "false")
                 .assertHasValue("PLUGIN_SEARCH_ELASTICSEARCH_BASEURL", "https://foo/");
     }
 


bianchi2:
@fredrikand thanks for your contribution. In fact, [both env vars are recognized](https://bitbucket.org/atlassian-docker/docker-atlassian-bitbucket-server/src/044add055c98f56ac5a11a610ef815788096c807/entrypoint.py#lines-10), so having 2 envs vars does not really make any difference. Do you have any use case in mind where `ELASTICSEARCH_ENABLED` is ignored?

fredrikand:
@bianchi2 
For some reason the entrypoint scripts does not recognize ELASTICSEARCH_ENABLED properly.
This might be a bug in the entrypoint scripts in the image, but since SEARCH_ENABLED is the new variable, I think this should be included.
I guess removing ELASTICSEARCH_ENABLED is also an option, but might break backwards compatibility?

```
# docker run --name bbsrv --rm -e ELASTICSEARCH_ENABLED=false atlassian/bitbucket-server:7.21.6-jdk11
WARNING:root:######################################################################
WARNING:root:Starting Bitbucket with embedded search service. Note that this is
not a recommended configuration and is known to have issues with
clean shutdown. Ideally OpenSearch (previously Elasticsearch) should
be started in a separate container/pod.
WARNING:root:######################################################################
INFO:root:User is currently root. Will downgrade run user to bitbucket
INFO:root:Running Bitbucket with command '/opt/atlassian/bitbucket/bin/start-bitbucket.sh', arguments ['/opt/atlassian/bitbucket/bin/start-bitbucket.sh', '-fg']
Starting Atlassian Bitbucket as the current user
20:30:55.413 [main] INFO com.atlassian.bitbucket.searchconfig.DefaultSearchConfigService - No existing bundled search configuration exists
```

```
# docker run --name bbsrv --rm -e SEARCH_ENABLED=false atlassian/bitbucket-server:7.21.6-jdk11
INFO:root:User is currently root. Will downgrade run user to bitbucket
INFO:root:Running Bitbucket with command '/opt/atlassian/bitbucket/bin/exec-bitbucket-node.sh', arguments ['/opt/atlassian/bitbucket/bin/exec-bitbucket-node.sh']
Starting Atlassian Bitbucket as the current user
Starting Atlassian Bitbucket as the current user
/opt/atlassian/bitbucket/bin/_exec-webapp.sh: line 62: \nStarting Bitbucket webapp at http://localhost:${bitbucket.http.port}${bitbucket.context}: bad substitution
2022-11-29 20:32:01,914 INFO  [main]  c.a.b.i.b.BitbucketServerApplication Starting BitbucketServerApplication v7.21.6 using Java 11.0.17 on c86347ebb45b with PID 7 (/opt/atlassian/bitbucket/app/WEB-INF/classes started by bitbucket in /var/atlassian/application-data/bitbucket)
2022-11-29 20:32:01,917 INFO  [main]  c.a.b.i.b.BitbucketServerApplication No active profile set, falling back to 1 default profile: "default"
2022-11-29 20:32:04,026 INFO  [main]  c.a.b.i.boot.log.BuildInfoLogger Starting Bitbucket 7.21.6 (d8e4f04 built on Wed Oct 26 07:01:16 CEST 2022)
2022-11-29 20:32:04,027 INFO  [main]  c.a.b.i.boot.log.BuildInfoLogger JVM: Eclipse Adoptium OpenJDK 64-Bit Server VM 11.0.17+8
2022-11-29 20:32:07,044 INFO  [main]  c.a.b.i.b.BitbucketServerApplication Started BitbucketServerApplication in 8.715 seconds (JVM running for 10.591)

```

bianchi2:
Indeed, [SEARCH_ENABLED](https://bitbucket.org/atlassian-docker/docker-atlassian-bitbucket-server/src/044add055c98f56ac5a11a610ef815788096c807/Dockerfile#lines-18) is set to `true` in a Dockerfile and it's [the first to be picked up in the entrypoint](https://bitbucket.org/atlassian-docker/docker-atlassian-bitbucket-server/src/044add055c98f56ac5a11a610ef815788096c807/entrypoint.py#lines-10).

Removing `ELASTICSEARCH_ENABLED` entirely should be safe, since changes to Dockerfile and/or scripts are propagated to all tags (i.e. they are updated). I'll have to confirm that though.

bianchi2:
@fredrikand it should be ok just to replace `ELASTICSEARCH_ENABLED` with `SEARCH_ENABLED` in env.

fredrikand:
@bianchi2 Followed your recommendation and replaced variable instead of adding.

bianchi2:
@fredrikand thanks for that. One minor thing - failed unit test. The fix is replace the env var here https://github.com/atlassian/data-center-helm-charts/blob/main/src/test/java/test/ElasticSearchTest.java#L29. Can you please do that and I'll re-run unit tests?

fredrikand:
@bianchi2 Thank you. I forgot about the tests after removing the old variable.

bianchi2 (approved):

errcode1202 (approved):

bianchi2:
Thanks @fredrikand . Merged 🥳

PR #489, opened 2022-11-30
CLIP-1721: Introduce Kubernetes in Docker (KinD) tests

bianchi2:
This PR adds a new reusable workflow (see checks in this PR below) that:

* creates a [KinD cluster](https://github.com/kubernetes-sigs/kind). Both K8S API and KinD versions are configurable.
* deploys Nginx ingress controller, NFS server and NFS client provisioner that automatically creates RWX PVs when a PVC is created
* deploys Postgres DB (Bitnami Helm chart) for each app
* deploys Bitbucket, Confluence, Jira and Bamboo (each in its own job, i.e. in its own cluster) with persistence and ingress enabled in values. The values have a few modifications related to lack of CPU, so requests had to be cut a bit. Apps supporting unattended setup use the corresponding values. Database, admin user and license env vars are added to all Helm releases. When Bamboo is deployed, Bamboo agent chart is also installed, agent is pointed to the server, and auto approval is turned on in server
* waits for the helm chart to be successfully deployed
* verifies endpoint status (sadly, it's not possible to run helm test since apps are available on localhost, and tests are run in a container)
* scales each app to 2 replicas and verifies the second pod is in Ready status (i.e. passes its readiness probe)
* collects k8s logs/events from each job and saves them as artifacts

What this PR is not:

* integration/func tests can be pointed to localhost and kind kube context but this is not in the scope of this PR

Existing limitations:
* for some reason when synchrony is enabled the job just hangs soon after a Synchrony pod is starting, and it's not possible to get any K8s logs/events to troubleshoot. KinD runs in a privileged container in docker, so odds are that either NFS mounts make it sad (when synchrony pod mounts shared home) or there are other issues related to nested virtualization. For now, Synchrony is disabled
* Bundled Elasticsearch for now
* Crowd isn't tested for now

The workflow is triggered when:

* push to main branch
* a PR is opened
* someone starts it manually from a selected branch

It takes <10 mins to run the workflow (Bitbucket is the slowest but we have 2vCPU only, so it's expected)

Next steps/improvements:
* figure out Synchrony issue in KinD
* deploy standalone Elasticsearch and force Bitbucket use it
* use existing test suite in KinD or write a new one with a series of functional tests

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -1,0 +1,92 @@
+name: Test DC App in a KinD Cluster
+
+on:
+  workflow_call:
+    inputs:
+      dc_app:
+        required: true
+        type: string
+    secrets:
+      BITBUCKET_LICENSE:
+        description: 'Bitbucket license'
+      CONFLUENCE_LICENSE:
+        description: 'Confluence license'
+      JIRA_LICENSE:
+        description: 'Jira license'
+      BAMBOO_LICENSE:
+        description: 'Bamboo license'
+
+jobs:
+  kind-testing:
+    runs-on: ubuntu-latest
+    env:
+      # See: https://github.com/kubernetes-sigs/kind/tags
+      KIND_VERSION: "v0.17.0"
+      # See: https://hub.docker.com/r/kindest/node/tags
+      K8S_VERSION: "v1.25.3"
+      DC_APP: ${{inputs.dc_app}}
+      LICENSE: ${{ secrets[format('{0}_LICENSE', inputs.dc_app)] }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@master
+
+      - name: Pin Kubectl version
+        uses: azure/setup-kubectl@v3.0
+        with:
+          version: 'v1.25.3'
+
+      - name: Pin Helm version
+        uses: azure/setup-helm@v3
+        with:
+          version: v3.9.2
+
+      - name: Create KinD cluster
+        run: |
+          src/test/scripts/kind/create_kind.sh
+
+      - name: Install KinD tooling
+        run: |
+          src/test/scripts/kind/configure_kind.sh
+
+      - name: Deploy postgres database
+        run: |
+          source src/test/scripts/kind/deploy_app.sh
+          deploy_postgres
+
+      - name: Create db, admin and license secrets
+        run: |
+          source src/test/scripts/kind/deploy_app.sh
+          create_secrets
+
+      - name: Deploy ${{inputs.dc_app}}
+        run: |
+          source src/test/scripts/kind/deploy_app.sh
+          deploy_app
+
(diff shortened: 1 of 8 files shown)


.github/workflows/kind.yaml
```diff
@@ -0,0 +1,92 @@
...
+        uses: azure/setup-kubectl@v3.0
+        with:
+          version: 'v1.25.3'
+
+      - name: Pin Helm version
+        uses: azure/setup-helm@v3
+        with:
+          version: v3.9.2
+
+      - name: Create KinD cluster
```
nanux, line 44: I like that they are distinct scripts in the workflow, so it should be easier to read the results e.g. when failure happens.

.github/workflows/kind.yaml
```diff
@@ -0,0 +1,92 @@
...
+          sleep 10
+          kubectl wait --for=condition=ready pod/${DC_APP}-1 -n atlassian --timeout=360s
+
+      - name: Get debug info
+        if: always()
+        run: |
+          kubectl get pods -n atlassian
+          kubectl describe pvc -n atlassian
+          kubectl describe nodes
+          curl -s https://raw.githubusercontent.com/atlassian-labs/data-center-terraform/main/scripts/collect_k8s_logs.sh | bash -s -- dummy dummy logs/${{inputs.dc_app}}
```
nanux, line 85: What is the `dummy` for?

src/test/config/kind/common-values.yaml
```diff
@@ -0,0 +1,47 @@
+DC_APP_REPLACEME:
+  # claim little resources as there are limitations in GitHub actions
+  resources:
+    container:
+      requests:
+        cpu: 20m
```
nanux, line 6: Is this enough for apps to spin up? I know it won't limit the app resources, but does it fit the nodes?

src/test/config/kind/common-values.yaml
```diff
@@ -0,0 +1,47 @@
...
+        memory: 1G
+  # we want to test scaling to 2 and making sure nodes have joined the cluster
+  clustering:
+    enabled: true
+
+  sysadminCredentials:
```
nanux, line 12: Bitbucket specific only

nanux (commented):
Awesome improvement @bianchi2, let's get it merged so we can test it properly.

nanux (approved):

src/test/config/kind/common-values.yaml
bianchi2: Yes, pods start slower but eventually start.

.github/workflows/kind.yaml
bianchi2: That script assumes that it's EKS cluster and needs clustername and region to update kubeconfig.

nanux (approved):

nghazali (approved):

PR #493, opened 2022-12-07
DCCLIP-270: Confluence S3 storage support

bianchi2:
This PR add additional envs to support upcoming Confluence S3 storage backend.

## Checklist
- [x] I have added unit tests
- [x] (Atlassian only) I have run the E2E test (if applicable)


src/main/charts/confluence/templates/serviceaccount.yaml
```diff
@@ -9,6 +9,12 @@ metadata:
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- if .Values.confluence.s3AttachmentsStorage.eksIrsa.roleArn }}
+  {{- if not .Values.serviceAccount.annotations }}
```
bianchi2, line 13: It's tricky to squeeze anything in `with` block, so as a workaround checking if service account annotations are defined to avoid `annotations` duplication.

src/main/charts/confluence/templates/config-jvm.yaml
```diff
@@ -16,6 +16,9 @@ data:
     {{ . }}
     {{- end }}
     -XX:ActiveProcessorCount={{ include "flooredCPU" .Values.confluence.resources.container.requests.cpu }}
+    {{ if .Values.confluence.s3AttachmentsStorage.eksIrsa.roleArn }}
+    -Daws.webIdentityTokenFile=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
```
bianchi2, line 20: Since by default `AWS_WEB_IDENTITY_TOKEN_FILE ` which EKS sets will be unset by the [image entrypoint](https://bitbucket.org/atlassian-docker/docker-atlassian-confluence-server/src/4ee36a2c95622f1312ec2d661b1d013132a0300b/entrypoint.py#lines-21),  adding its equivalent as a Java property.

nghazali (approved):

errcode1202 (changes requested):
Hi @bianchi2 there has been ongoing discussions recently about how config should be supplied across the board, see here: https://stash.atlassian.com/projects/CONFSERVER/repos/confluence/pull-requests/19670/overview.

Long story short, supplying config via environment variables is frowned upon from a product standpoint. In terms of the Confluence Docker image, this is fine as we just provide the bucket name and region via `JVM_SUPPORT_RECOMMENDED_ARGS` but for the helm charts particularly this PR we should change things so that the `statefulset` doesnt set `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_NAME` and `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_REGION` as environment variables when supplied via the `s3AttachmentsStorage` stanza in `values.yaml`. Rather these should be set using `config-jvm.yaml` with a helper function for each property like is done for: [confluence.sysprop.clusterNodeName](https://github.com/atlassian/data-center-helm-charts/blob/28c42add78b78918d584637e94b09592382784bb/src/main/charts/confluence/templates/_helpers.tpl#L113)

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +608,31 @@ confluence:
     #
     usePodNameAsClusterNodeName: true
 
+  # Use AWS S3 to store attachments. From Confluence 8.1 onwards
```
nanux, line 611:
```suggestion
  # Use AWS S3 to store attachments. From Confluence 8.1 onwards.
```

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +608,31 @@ confluence:
...
+  #
+  s3AttachmentsStorage:
+    # Bucket name to store attachments. If a bucket name is defined, Confluence will automatically
```
nanux, line 614: Empty line

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +608,31 @@ confluence:
...
+    # use AWS S3 to store attachments. Only bucket name is required, not the full arn. This can also be defined in
+    # `confluence.additionalEnvironmentVariables` as `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_NAME`
+    # or `confluence.additionalJvmArgs` as `-Dconfluence.filestore.attachments.s3.bucket.name=<bucket-name>`
+    #
+    bucketName:
```
nanux, line 619: Empty line

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +608,31 @@ confluence:
     #
     usePodNameAsClusterNodeName: true
 
+  # Use AWS S3 to store attachments. From Confluence 8.1 onwards
+  #
+  s3AttachmentsStorage:
+    # Bucket name to store attachments. If a bucket name is defined, Confluence will automatically
+    # use AWS S3 to store attachments. Only bucket name is required, not the full arn. This can also be defined in
```
nanux, line 615: I am not sure we need to include

```This can also be defined in
    # `confluence.additionalEnvironmentVariables` as `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_NAME`
    # or `confluence.additionalJvmArgs` as `-Dconfluence.filestore.attachments.s3.bucket.name=<bucket-name>
```

this here. Maybe in docs? But if they are setting the value here, why customer need to understand the internals?

src/main/charts/confluence/values.yaml
nanux, line 614:
```suggestion
    # Bucket name to store attachments. If a bucket name and region (see below) are defined, Confluence will automatically
```

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +608,31 @@ confluence:
...
+    # `confluence.additionalEnvironmentVariables` as `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_NAME`
+    # or `confluence.additionalJvmArgs` as `-Dconfluence.filestore.attachments.s3.bucket.name=<bucket-name>`
+    #
+    bucketName:
+    # AWS region where the S3 bucket is located. This can also be defined in
+    # `confluence.additionalEnvironmentVariables` as `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_REGION`
```
nanux, line 621: The same applies as above. I am not sure this is necessary.

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +608,31 @@ confluence:
...
+    # or `confluence.additionalJvmArgs` as `-Dconfluence.filestore.attachments.s3.bucket.region=<aws-region>`
+    #
+    awsRegion:
+
+    # Use EKS IRSA. See: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
+    #
+    eksIrsa:
```
nanux, line 628: Is this useful only for S3? Can we extract that to higher level?

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +608,31 @@ confluence:
...
+    # use AWS S3 to store attachments. Only bucket name is required, not the full arn. This can also be defined in
+    # `confluence.additionalEnvironmentVariables` as `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_NAME`
+    # or `confluence.additionalJvmArgs` as `-Dconfluence.filestore.attachments.s3.bucket.name=<bucket-name>`
+    #
+    bucketName:
+    # AWS region where the S3 bucket is located. This can also be defined in
+    # `confluence.additionalEnvironmentVariables` as `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_REGION`
+    # or `confluence.additionalJvmArgs` as `-Dconfluence.filestore.attachments.s3.bucket.region=<aws-region>`
+    #
+    awsRegion:
```
nanux, line 624:
```suggestion
    region:
```

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +608,31 @@ confluence:
...
+
+    # Use EKS IRSA. See: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
+    #
+    eksIrsa:
+      # Defining a full arn of AWS role to assume will automatically annotate confluence service account this role.
+      # Make sure the policy attached to the role supports the following actions: "s3:PutObject", "s3:ListBucket",
+      # "s3:GetObject", "s3:DeleteObject" on resources "arn:aws:s3:::${BUCKET_NAME}/*" and "arn:aws:s3:::${BUCKET_NAME}"
```
nanux, line 631: This should link to official Confluence documentation, when it is ready.

src/test/java/test/ConfluenceS3EnabledTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @BeforeEach
+    void initHelm(TestInfo testInfo) {
+        helm = new Helm(testInfo);
+    }
+
+
+    @ParameterizedTest
+    @EnumSource(value = Product.class, names = {"confluence"})
+    void confluence_s3_storage_env_vars(Product product) throws Exception {
```
nanux, line 25: We can also test a case when only one of the values is defined => and the sys props are not included.

src/test/java/test/model/KubeResources.java
```diff
@@ -44,6 +47,11 @@ public Deployment getDeployment(final String name) {
                 .getOrElseThrow(() -> new AssertionError("No Deployment found with name " + name));
     }
 
+    public ConfigMap getConfigMap(final String name) {
```
nanux, line 50: Interesting we didn't need it until now :)

nanux: There are few places where we could rewrite `resources.get(ConfigMap, product.getHelmReleaseName() + "xyz");`. But that is only for bonus points ;)

nanux (changes requested):

src/main/charts/confluence/templates/_helpers.tpl
```diff
@@ -110,6 +110,13 @@ Pod labels
 -Dconfluence.cluster.hazelcast.listenPort={{ .Values.confluence.ports.hazelcast }}
 {{- end }}
 
+{{- define "confluence.sysprop.s3Config" -}}
+{{- if and .Values.confluence.s3AttachmentsStorage.bucketName .Values.confluence.s3AttachmentsStorage.awsRegion }}
+-Dconfluence.filestore.attachments.s3.bucket.name={{ .Values.confluence.s3AttachmentsStorage.bucketName }}
+-Dconfluence.filestore.attachments.s3.bucket.region={{ .Values.confluence.s3AttachmentsStorage.awsRegion }}
```
nanux, line 116: We should probably also include the ability to use `endpointOverride` as in Kubernetes environment I can imagine using minio will be a nice "experimental" feature.

nanux (changes requested):

src/main/charts/confluence/values.yaml
bianchi2: perhaps, it should be in serviceAccount section

src/test/java/test/model/KubeResources.java
bianchi2: In fact, I failed to find a common way to get a configmap, hence added it to be available for all tests. Yes, we can refactor existing tests afterwards.

src/main/charts/confluence/values.yaml
```diff
@@ -608,6 +617,24 @@ confluence:
...
+    # `confluence.additionalEnvironmentVariables` as `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_NAME`
+    # or `confluence.additionalJvmArgs` as `-Dconfluence.filestore.attachments.s3.bucket.name=<bucket-name>`
+    #
+    bucketName:
+
+    # AWS region where the S3 bucket is located. This can also be defined in
+    # `confluence.additionalEnvironmentVariables` as `CONFLUENCE_FILESTORE_ATTACHMENTS_S3_BUCKET_REGION`
+    # or `confluence.additionalJvmArgs` as `-Dconfluence.filestore.attachments.s3.bucket.region=<aws-region>`
+    #
+    region:
```
errcode1202, line 636: for consistency can this be renamed to `bucketRegion`?

errcode1202 (approved):

src/main/charts/confluence/values.yaml
bianchi2: Moved to serviceAccount

src/main/charts/confluence/values.yaml
bianchi2: Fixed

src/main/charts/confluence/values.yaml
bianchi2: Since IRSA is a generic thing, I removed any policy/role instructions, as this should be in official docs

src/test/java/test/ConfluenceS3EnabledTest.java
bianchi2: Test case added

src/main/charts/confluence/templates/_helpers.tpl
bianchi2: Added

src/main/charts/confluence/values.yaml
bianchi2: Done

errcode1202 (approved):

nanux (approved):

PR #496, opened 2022-12-14
DCCLIP-350: Bitbucket Mesh

bianchi2:
This PR adds support of Bitbucket Mesh

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/src/main/charts/bitbucket/values.yaml
+++ b/src/main/charts/bitbucket/values.yaml
@@ -752,6 +752,110 @@ bitbucket:
     #
     upstreamUrl:
 
+  # Mesh configuration
+  mesh:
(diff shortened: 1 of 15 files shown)


src/main/charts/bitbucket/values.yaml
```diff
@@ -752,6 +752,110 @@ bitbucket:
     #
     upstreamUrl:
 
+  # Mesh configuration
+  mesh:
+    # -- Enable Bitbucket Mesh. See: https://confluence.atlassian.com/bitbucketserver/bitbucket-mesh-1128304351.html
+    #
+    enabled: true
```
tarka, line 759: Doesn't matter here, but we need to make a decision at some point about whether to enable this by default.

PR #500, opened 2022-12-20
CLIP-1747: Update Postgres and Elasticsearch helm charts

bianchi2:
Bitnami have removed postgresql 10.16.2 and elasticsearch 17.9.29 from their indexes.

This PR bumps bitnami/postgresql chart version, and replaces bitnami/elasticsearch chart with an official elastic chart.

Had to completely replace postgres and es values to match new helm charts.

As a result, postgres 14 is deployed, while Elasticsearch version remained unchanged.

## Checklist
- [ ] I have added unit tests


src/test/infrastructure/elasticsearch/elasticsearch-values.yaml
```diff
@@ -1,27 +1,15 @@
----
+persistence:
+  enabled: false
 
-global:
-  coordinating:
-    name: coo
+secret:
```
bianchi2, line 4: Auth is enabled by default. If we want to we may keep it enabled and pass secret name and keys to bitbucket.

src/test/infrastructure/elasticsearch/elasticsearch-values.yaml
```diff
@@ -1,27 +1,15 @@
...
+  enabled: false
 
-master:
-  replicas: 1
-  podAnnotations:
-    "atlassian.com/business_unit": "Engineering: Enterprise DC"
-  persistence:
-    enabled: false
+createCert: false
```
bianchi2, line 7: Disable ssl, or else Bitbucket will need the cert it its keystore, which is an overkill for CI.

src/test/scripts/helm_install.sh
```diff
@@ -126,9 +129,9 @@ bootstrap_database() {
      --values $PSQL_CHART_VALUES \
      --set fullnameOverride="$POSTGRES_RELEASE_NAME" \
      --set image.tag="$POSTGRES_APP_VERSION" \
-     --set postgresqlDatabase="$DB_NAME" \
-     --set postgresqlUsername="$PRODUCT_NAME" \
-     --set postgresqlPassword="$PRODUCT_NAME" \
+     --set auth.database="$DB_NAME" \
```
bianchi2, line 132: See: https://github.com/bitnami/charts/tree/main/bitnami/postgresql

nghazali (approved):

nanux (approved):

nanux:
This is green as well - https://server-syd-bamboo.internal.atlassian.com/browse/DCD-K8SHELMTEST144-8

PR #508, opened 2023-02-07
feat: Added support for configuring healthcheckPath to Synchrony readinessProbe

jvasallo:
## Pull request description

When setting up my external Google Cloud Ingress (not using Nginx Ingress), we need the ability to configure the healthcheck path. As documented, Synchrony expects the path to include "/synchrony" to be included in all requests. If not, we were seeing healthchecks returning back a `204 No Content`.

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/confluence/templates/statefulset-synchrony.yaml
+++ b/src/main/charts/confluence/templates/statefulset-synchrony.yaml
@@ -54,7 +54,7 @@ spec:
           readinessProbe:
             httpGet:
               port: {{ .Values.synchrony.ports.http }}
-              path: /heartbeat
+              path: {{ .Values.synchrony.readinessProbe.healthcheckPath }}
             initialDelaySeconds: {{ .Values.synchrony.readinessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.synchrony.readinessProbe.periodSeconds }}
             failureThreshold: {{ .Values.synchrony.readinessProbe.failureThreshold }}
--- a/src/main/charts/confluence/values.yaml
+++ b/src/main/charts/confluence/values.yaml
@@ -829,6 +829,12 @@ synchrony:
   #
   readinessProbe:
 
+    # -- The healthcheck path to check against for the Synchrony container 
+    # useful when configuring behind a reverse-proxy or loadbalancer such as "/synchrony/heartbeat"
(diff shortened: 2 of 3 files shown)


bianchi2:
@jvasallo I am struggling to understand in which scenario you would want to change path for httpGet in the probe config. Would't it fails if it's anything but /heartbeat? Kubelet will use internal IP to run the requests, i.e. ingress isn't involved

jvasallo:
@bianchi2 Google Cloud Ingresses by default look at the readiness probe and use that as the interpreted healthcheck:
https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#interpreted_hc

My solution was to retain the existing default path (`/heartbeat`), whilst allowing me to customize it to the true working path (`/synchrony/heartbeat`). I was hoping this would spawn a conversation on the default HC path as I am not sure how it ever worked...yet folks on the Atlassian forums seem to claim it does. 

I can revert my changes (and hardcode the path) or allow users to specify the path. It could be helpful if folks front Syncronchy or need to tweak it at all in the java settings.

nghazali:
@jvasallo , thank you for the contribution. I think this change is useful and regardless of the k8s documentation about the acceptance result for health check, this is an improvement. WDYT @bianchi2?

bianchi2:
@jvasallo can you please make the default `/synchrony/heartbeat` since this is the path that returns 200. Once done, I am happy to merge :)

jvasallo:
@bianchi2 In terms of implementation, would you rather:

1. Retain the ability to override HC Path in Helm Value
2. Fix the path in the template, revert Helm Value adds.

I am leaning towards option 2 as it's more inline with the rest of the project but happy to keep changes I originally proposed :)

bianchi2:
In fact, I'm ok with keeping it as is - having default in values and making it possible to override.

Also, could you please fix unit tests by changing path here too? https://github.com/atlassian/data-center-helm-charts/blob/main/src/test/resources/expected_helm_output/confluence/output.yaml#L167

jvasallo:
Awesome and thank you. Hours didn't align for me to fix it, but was about to this morning. Let me know if you need anything else from me and thank you!

bianchi2 (approved):

nghazali (approved):

PR #509, opened 2023-02-08
CLIP-1767: Replica count for Synchrony; defaultMode made configurable in volumes

bianchi2:
This PR makes it possible to scale Synchrony sts independently from Confluence.

Also, defaultMode becomes configurable which is handy when mounting secrets and configmaps as executable files. It defaults to 484 (decimal) which is 0744 octal.

## Checklist
- [x] I have added unit tests
- [x] (Atlassian only) I have run the E2E test (if applicable)


src/main/charts/confluence/values.yaml
```diff
@@ -375,6 +375,11 @@ volumes:
   # those needs to be defined in synchrony.additionalVolumeMounts
   #
   additionalSynchrony: []
+  
+  # -- Mode bits used to set permissions on created files by default. Must be an octal value between 0000 and 0777 or a decimal value between 0 and 511
+  # Typically overriden in volumes from Secrets and ConfigMaps to make mounted files executable
+  #
+  defaultMode: 484
```
errcode1202, line 382: Minor - would it be clearer if this was renamed to `defaultPermissionsMode`

errcode1202 (approved):

src/main/charts/confluence/values.yaml
bianchi2: I used the name as it appears in K8S API assuming that users who are going to override it know exactly what they are doing. Renaming it should not be a big deal, I just wanted to stick to kube api names.

bianchi2: Renamed

src/main/charts/confluence/values.yaml
```diff
@@ -375,6 +375,11 @@ volumes:
   # those needs to be defined in synchrony.additionalVolumeMounts
   #
   additionalSynchrony: []
+  
+  # -- Mode bits used to set permissions on created files by default. Must be an octal value between 0000 and 0777 or a decimal value between 0 and 511
+  # Typically overridden in volumes from Secrets and ConfigMaps to make mounted files executable
+  #
+  defaultPermissionsMode: 484
```
nanux, line 382: Might be worth leaving comment # decimal 484 == octal 0744

nanux (approved):

errcode1202 (approved):

Issue #510, opened 2023-02-08
[Suggestion] - Restart or relaunch pod when OOM


vernonlingleysonos:
### Suggestion

With the current setup, pods that hit an OOM error or the like are not restarted or rebuilt, they are only removed from the service's load balancer. This is because the JIRA chart only uses a `readinessProbe`, not a `livenessProbe` as suggested by the Kubernetes documentation https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/. By including a `livenessProbe` with a TCP check on the port used by the application alongside the existing readiness check, the k8s cluster can monitor the health of the application running on the pod while also supporting other intended actions that may take a pod offline, such as doing a full reindexing.

I have tested this out using a modified chart built from a fork of this repo and it works as intended. I will submit a PR once I clear the CLA.

### Product

Jira

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

vernonlingleysonos:
https://github.com/atlassian/data-center-helm-charts/pull/511

PR #511, opened 2023-02-08
JIRA - Add livenessProbe check to monitor the health of applications for each pod

vernonlingleysonos:
## Pull request description

As outlined in [ [Suggestion] - Restart or relaunch pod when OOM #510 ](https://github.com/atlassian/data-center-helm-charts/issues/510) JIRA pods that hit an OOM error are not automatically restarted or rebuilt, they are just taking offline. This means that manual intervention is required to clear these errors. By using a `livenessProbe` check you can automate this process, while continuing to use the `readinessProbe` check to take nodes out of the service's load balancer when reindexing the service.

## Checklist
- [x] I have added unit tests


bianchi2:
@vernonlingleysonos thanks for your contribution. Indeed, liveness probes are useful (but can be disruptive as well in some cases).  A couple of questions/suggestions:

* do you think it makes sense to use tcp check for a startup probe, i.e. start checking /status only when the server is listening on a socket. I bet polling `/` will be a bunch of 302 (which are fine)
* perhaps it makes sense to make liveness probe confitional? On by default but make it possible to turn it off if an admin is ok with having a struggling pod hanging for a while (for example to investigate some issues)?

vernonlingleysonos:
>  @vernonlingleysonos thanks for your contribution. Indeed, liveness probes are useful (but can be disruptive as well in some cases). A couple of questions/suggestions:
> 
>   * do you think it makes sense to use tcp check for a startup probe, i.e. start checking /status only when the server is listening on a socket. I bet polling `/` will be a bunch of 302 (which are fine)

No harm here. I was using http because I wasn't sure when tomcat opened the port up to any TCP traffic. It's kinda moot though, since the pod won't be doing anything until the `readinessProbe` passes as well.

>
>   * perhaps it makes sense to make liveness probe confitional? On by default but make it possible to turn it off if an admin is ok with having a struggling pod hanging for a while (for example to investigate some issues)?

I added in `Values.livenessProbe.enabled`, defaulted to true, and make it conditional inside of `statefulset.yaml`. 

>
>   * I understand you're targeting Jira but could you add liveness probes for all the products please?

Maybe, but I don't think I would be the right person for this. I don't have any way to test them all out as  only have been using Jira inside of k8s. In addition I am not really familiar with hosting bamboo or crucible and wouldn't know what to change.

>
>   * Unit tests are failing. You need to revers changes to Cchart.yaml. Also, there's a test to compare processed template to expected output.

Done.

src/main/charts/jira/templates/statefulset.yaml
```diff
@@ -122,6 +122,18 @@ spec:
             initialDelaySeconds: {{ .Values.jira.readinessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.jira.readinessProbe.periodSeconds }}
             failureThreshold: {{ .Values.jira.readinessProbe.failureThreshold }}
+          {{- if .Values.livenessProbe.enabled }}
```
bianchi2, line 125: .Values.jira.livenessProbe.enabled

bianchi2:
@vernonlingleysonos one last thing please :) Even though it's in the PR template, can you please revert changes to Changelog? We're generating it automatically now. Thanks.

vernonlingleysonos:
> @vernonlingleysonos one last thing please :) Even though it's in the PR template, can you please revert changes to Changelog? We're generating it automatically now. Also, can you please sign CLA? Thanks.

Okay! I reverted the changelog. As for the CLA I am waiting for my legal team to review it before signing.

bianchi2:
@vernonlingleysonos thanks for signing the CLA! I have got a few comments.

You decreased [failureThreshold to 1](https://github.com/atlassian/data-center-helm-charts/pull/511/files#diff-84a7b6888b640f239b64d5d8168c9069fcaeb5edb85961ed674a35a20a5b1b8bR470). I makes sense when a startup probe is there (but there can be situations (potentially) when Tomcat is up and listening in tcp socket but status does not respond with the status code between 200 and 400). However, if you disable liveness probe, which will disable startup probe too, then Jira pods will have 10 seconds to pass their readiness probes.

Also I think it makes sense to add initialDelaySeconds to startupProbe to give container some time before kubelet starts checking? Also, I am now thinking that perhaps using httpGet is a better idea, as listening on portdoes not always mean the app is ready to serve requests. WDYT?

vernonlingleysonos:
@bianchi2 The reason I set `failureThreshold` to 1 rather than 30 is so that if the application is not up and available (with returning a good status on `/status`) that it won't be added to the service's load balancer. In what situation would you want the system on the load balancer but unable to serve any traffic? If you are thinking about having a delay during the startup, that's better handled with the `initialDelaySeconds` setting. Even if `livenessProbe` is disabled this setting still works, as an unhealthy instance won't serve up any traffic. The worst that happens is that the probe starts checking starts early, which shouldn't hurt the service in any way as it won't remove any pods.

I do agree with your second point though. Do you think the `readinessProbe`'s `initialDelaySeconds` setting should match the total `failureThreshold` and checks setting for the `startupProbe`? Right now it's set to 600 seconds/10 minutes, since it took 5+ minutes for my cluster to do that when it had to build out new home directories. I can lower that though if you think it is too high, or move it back to the old setting of 150 seconds.

IMO adding a `initialDelaySeconds` setting to `readinessProbe` could be confusing, since you need to add that time to figure out the total time before the startup probe fails. ```initialDelaySeconds * readinessProbe``` is easier to wrap your head around than ```(initialDelaySeconds * readinessProbe) + initialDelaySeconds```

bianchi2:
@vernonlingleysonos thanks for the explanation. Indeed, disabling liveness probe did not do any harm, so we're all good here.

As to initialDelaySeconds, having thought about it again, I think it's ok too. Readiness probe will start only after startup probe is done. And since everything is configurable in values.yaml, it's up to admins to configure whatever makes sense to them. It can take 30+ mins for a new Jira node to start with a big dataset, so I don't think we can come up with a reasonable defaults.

bianchi2 (approved):

vernonlingleysonos:
@nanux @tarka @nghazali @errcode1202 @yzha645 @jjeongatl I would appreciate a review when you have time, I think this would be a useful feature for helm deployments.

nanux (approved):
Thank you for the improvement @vernonlingleysonos!

mattj-relativity:
I believe this introduces a regression in an edge case.  there are scenarios during an initalization of a new helm environment for confluence ( at least ) where status will return 500 but useful information will be displayed on port 8090.  Readiness and Liveness probe yaml sections in value yaml seem to have no impact on overriding the readiness check on 8090/status.

There should be a way to disable that in the values yaml or more likely, status should not be used for this.  The runtime pid lockfile on filesystem seems like a much better choice.

bianchi2:
@mattj-relativity good point on being able to disable startup and readiness probes. As to replacing them with pid lockfile or smth similar, it will not guarantee the application is really ready, so `/status` is a much more reliable way to mark the pod as ready.

I'll be happy to approve a PR with a new `enabled` field for readiness/startup probes :)

mattj-relativity:
Hey thanks for the reply... since posting this I have been further down the rabbit hole.

So this PR seems to be JIRA specific... which is interesting cause the values yaml for confluence has some readinessprobe overide segments:
https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/confluence/values.yaml#L594C1-L608C24 
But looking at the actual statefulset... and from functional testing...
https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/confluence/templates/statefulset.yaml#L76
They aren't actually wired to the readiness probe for the initial startup.

So on one hand I'd love to see it fixed on both sides to be configurable... on the other... I probably can't push a merge request to ya'll cause atm I probably can't convince our org to do CLA work.  We're kinda in a heads down major sprint on something and we need to avoid distractions...  so roping in legal for conversations would likely be a non starter.

Glad to assist however in that limited capacity though.

bianchi2:
@mattj-relativity yeah, I hear you :) 

I am thinking of adding the following:
* turn off readiness and startup probes entirely (on by default)
* redefine readiness probe in values.yaml (say, you want tcpSocket or exec)

WDYT?

bianchi2:
Raised https://github.com/atlassian/data-center-helm-charts/pull/626

PR #522, opened 2023-02-20
CLIP-1774: Automate Helm charts release in GitHub Actions

bianchi2:
This PR adds 2 new workflows and introduces a few changes to the release scripts, namely:

* there's no filtering by Jira issue key, so all commits are picked up
* changelog is generated for each product - adding path to chart root to git log command. If there are no changes, the default "Update Helm chart version" is added to changelog
* fix getting the latest version in Changelog.md, handle 2 digits after `1`
* remove duplicate git log messages (like `Update appVersions for DC apps` which is committed every week by a bot)
* add GitHub action that will generate changelog and commit changes back (triggered in `release/*` branches only)
* renames the script to be able to all a function to get changelog (if there's a hyphen in the script name, it'll fail):

```
python3 -c 'import prepare_release; print (prepare_release.gen_changelog("bamboo", "/Users/user/projects/forks/atlassian/data-center-helm-charts"))'

['* Fix typo in readme as per jira ticket', '* Update appVersions for DC apps', '* Update Bamboo Readme file']
```

Ideally, the changelog should not require any editing or deep review after the release script generates it.

Prepare release workflow will prepare chart files for a release and raise a PR with changes. Release charts workflow will cut a new release and raise a PR with changes to index.yaml

## Checklist
- [ ] I have added unit tests


.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,82 @@
...
+      PYTHONPATH: /opt/hostedtoolcache/Python/3.9.14/x64/lib/python3.9/site-packages
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Checkout repo
```
bianchi2, line 24: Even fetch-depth = 0 does not help and git log says tags are bad revision. As a workaround, clone the repo with git, checkout to a branch, run prepare_release script and then copy src back to the repo checkout in Checkout code step.

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,180 @@
...
+    repo = git.Repo(path)
+    cli = git.Git(path)
+    tags = sorted(repo.tags, key=lambda t: t.commit.committed_datetime)
+
+    lasttag = tags[-1]
+    tagver = re.sub(r'^[^-]+-', '', lasttag.name)
+    log.info(f'Generating {product} changelog since {tagver}')
+
+    # git log will pick up commits in src/main/charts/$product directory only
+    changelog = cli.log(f'{lasttag}..main', "--", prodbase + '/' + product, graph=True, pretty='format:%s',
```
bianchi2, line 46: Adding `-- path/to/chart/root` to pick up commits relevant for a product

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,180 @@
...
+                        abbrev_commit=True, date='relative', )
+
+    if not changelog:
+        # It is possible that there are no commits to the Helm chart,
+        # in this case we just write a generic git log
+        log.info(f'No commits to {product} Helm chart found')
+        default_git_log = '* Update Helm chart version'
```
bianchi2, line 53: We can change this message, but in any case having empty release notes isn't good

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,180 @@
...
+        changelog = default_git_log
+
+    changelog = changelog.split('\n')
+    filtered_changelog = list(changelog)
+
+    # remove Jira keys from commit messages. This substitution assumes the commit message may have the following format:
+    # CLIP-1234: Here is my message
+    # DCCLIP-1234: Here is my message
+    sanitized_changelog = map(lambda c: re.sub(r'(CLIP|DCCLIP)-[0-9]{1,5}(?![0-9]): ', '', c), filtered_changelog)
```
bianchi2, line 62: We don't need Jira keys in changelog

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,180 @@
...
+def update_changelog_file(product, version, changelog, chartversions):
+    log.info(f"Updating {product} Changelog.md to {version}")
+    proddir = f"{prodbase}/{product}"
+    clfile = f"{proddir}/Changelog.md"
+    (tmpfd, tmpname) = mkstemp(dir=proddir, text=True)
+
+    foundfirst = False
+    with open(clfile, 'r') as clfd:
+        for line in clfd:
+            if not foundfirst and re.match(r'^## [0-9]\.[0-9]{1,2}\.[0-9]{1,2}', line) != None:
```
bianchi2, line 87: Expect double digits in the second and third number, i.e. `1.10.1` or `1.10.11`

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,180 @@
...
+    sanitized_changelog = map(lambda c: re.sub(r'(CLIP|DCCLIP)-[0-9]{1,5}(?![0-9]): ', '', c), filtered_changelog)
+    return list(dict.fromkeys(sanitized_changelog))
+
+
+def format_changelog_yaml(changelog):
+    # The artifacthub annotations are a single string, but formatted
+    # like YAML. Replacing the leading '*' with '-'  and wrapping
+    # strings in double quotes should be sufficient.
+    sanitized_changelog = []
+    for string in changelog:
```
bianchi2, line 71: Wrap each string in double quotes

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,180 @@
...
+        string = string + '"'
+        sanitized_changelog.append(string)
+    c2 = map(lambda c: re.sub(r'^\* ', '- "', c), sanitized_changelog)
```
bianchi2, line 74: Replace `* ` with `- "` which will end up in:

```
- "Line 1"
- "Line 2"
```

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,180 @@
...
+        changelog = default_git_log
+
+    changelog = changelog.split('\n')
+    filtered_changelog = list(changelog)
+
+    # remove Jira keys from commit messages. This substitution assumes the commit message may have the following format:
+    # CLIP-1234: Here is my message
+    # DCCLIP-1234: Here is my message
+    sanitized_changelog = map(lambda c: re.sub(r'(CLIP|DCCLIP)-[0-9]{1,5}(?![0-9]): ', '', c), filtered_changelog)
+    return list(dict.fromkeys(sanitized_changelog))
```
bianchi2, line 63: Remove duplicate commits, e.g. Upate appVersion for DC apps

.github/pull_request_template.md
```diff
@@ -5,6 +5,5 @@ _Provide description for the PR_
 ## Checklist
 - [ ] I have added unit tests
 - [ ] I have applied the change to all applicable products
-- [ ] I have added the change description to the `changelog.md` and `Chart.yaml` files
-- [ ] (Atlassian only) I have run the E2E test (if applicable)
+- [ ] The E2E test is passing (`e2e` label)
```
nghazali, line 8:
```suggestion
- [ ] The E2E test is passed (use `e2e` label)
```

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,183 @@
...
+    log.info(f'Generating {product} changelog since {tagver}')
+
+    # git log will pick up commits in src/main/charts/$product directory only
+    changelog = cli.log(f'{lasttag}..main', "--", prodbase + '/' + product, graph=True, pretty='format:%s',
+                        abbrev_commit=True, date='relative', )
+
+    changelog = changelog.split('\n')
+    # we don't need pre-release commits that update chart.yamls and changelogs
+    # this pattern is based on the commit message from a GitHub action that prepares Helm release
+    pattern = re.compile(r'^\* Prepare release [0-9].{1,4}')
```
bianchi2, line 52: Filter out commit messages from GH actions bot

src/main/scripts/generate_chart_repo.sh
```diff
@@ -27,20 +27,19 @@ rm -rf "$PACKAGE_DIR"
...
-  -v "$(pwd)/$PACKAGE_DIR:/releases" \
-  --entrypoint cr \
+docker run -v "$(pwd)/$PACKAGE_DIR:/releases" \
   --rm \
-  quay.io/helmpack/chart-releaser:v1.2.1 \
+  quay.io/helmpack/chart-releaser:v1.5.0 \
   upload \
   --skip-existing \
   --package-path /releases \
+  --release-notes-file RELEASE_NOTES.md \
```
bianchi2, line 42: This makes it possible to include release notes in GitHub releases page in UI

src/main/scripts/generate_chart_repo.sh
```diff
@@ -27,20 +27,19 @@ rm -rf "$PACKAGE_DIR"
...
 echo "Uploading chart packages as Github releases"
 # This will scan $PACKAGE_DIR for the tgz files that 'helm package' just generated, and upload them to the GitHub
 # repo as Release artifacts. GitHub will create corresponding git tags for each chart.
-docker run --user "$(id -u):$(id -g)" \
-  -v "$(pwd)/$PACKAGE_DIR:/releases" \
-  --entrypoint cr \
+docker run -v "$(pwd)/$PACKAGE_DIR:/releases" \
   --rm \
-  quay.io/helmpack/chart-releaser:v1.2.1 \
+  quay.io/helmpack/chart-releaser:v1.5.0 \
```
bianchi2, line 38: 1.5.0 is the latest as of today

src/main/scripts/generate_chart_repo.sh
```diff
@@ -51,16 +50,15 @@ echo "Regenerating chart repo index.yaml"
 # $PUBLISH_DIR for committing to git.
 docker run \
   --user "$(id -u):$(id -g)" \
-  -v "$(pwd)/$PUBLISH_DIR:/index" \
+  -v "$(pwd):/index" \
   -v "$(pwd)/$PACKAGE_DIR:/packages" \
-  --entrypoint cr \
```
bianchi2, line 56: No need to set entrypoint here

src/main/scripts/generate_chart_repo.sh
```diff
@@ -51,16 +50,15 @@ echo "Regenerating chart repo index.yaml"
 # $PUBLISH_DIR for committing to git.
 docker run \
   --user "$(id -u):$(id -g)" \
-  -v "$(pwd)/$PUBLISH_DIR:/index" \
```
bianchi2, line 54: We need the entire repo since 1.5.0 helm chart releaser expects a git repository.

src/main/scripts/generate_chart_repo.sh
```diff
@@ -51,16 +50,15 @@ echo "Regenerating chart repo index.yaml"
...
+  --workdir="/index" \
   --rm \
-  quay.io/helmpack/chart-releaser:v1.2.1 \
+  quay.io/helmpack/chart-releaser:v1.5.0 \
   index \
   --owner atlassian \
   --git-repo data-center-helm-charts \
-  --charts-repo https://atlassian.github.io/data-center-helm-charts \
-  --index-path /index/index.yaml \
+  --index-path /index/docs/docs/index.yaml \
```
bianchi2, line 61: Change path for index.yaml since we're mounting an entire repo

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          python -m pip install --upgrade pip gitpython ruamel_yaml
+
+      - name: Generate Release Notes for GitHub Releases
+        run: |
+          cd src/main/scripts
+          products=(bamboo bamboo-agent bitbucket confluence crowd jira)
+          export REPO_PATH="${GITHUB_WORKSPACE}"
+          export CHARTS_LOCATION="src/main/charts"
+          for product in "${products[@]}"; do
```
bianchi2, line 50: Upload another md file to each chart tgz so that chart releaser can upload release notes to GitHub releases page.

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          git checkout -b update-index-yaml-${GITHUB_RUN_ID}
+          git commit -m "Update index.yaml post release"
+          git push origin update-index-yaml-${GITHUB_RUN_ID}
+          gh pr create --title "Update index.yaml" --body "Update index.yaml after release" --base main --head update-index-yaml-${GITHUB_RUN_ID}
+      
+      - name: Upload Public Key To Release Assets
+        run: |
+          cd src/main/scripts
+          export RELEASE_VERSION=$(python3 -W ignore -c "import prepare_release; print (prepare_release.get_chart_versions(\"${GITHUB_WORKSPACE}/src/main/charts\"))" | sed "s/'/\"/g" | python -c 'import sys, json; print (json.load(sys.stdin)["bamboo"]["version"])')
```
bianchi2, line 99: Here we parse  Chart.yaml of any of the products to get the version which is then used in the uploader script.

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          echo "[INFO]: Preparing release ${GITHUB_REF_NAME##*/}"
+          python src/main/scripts/prepare_release.py ${GITHUB_REF_NAME##*/}
+
+      - name: Configure Git
+        run: |
+          git config --global user.name "$GITHUB_ACTOR"
+          git config --global user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Lint yamls with ah cli
```
bianchi2, line 51: We need to make sure ArtifactHub.io will like the annotations and the new version will be successfully parsed.

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,71 @@
...
+        uses: envoy/install-helm-docs@v1.0.0
+        with:
+          version: 1.11.0
+
+      - name: Generate changelog and Chart annotations
+        run: |
+          echo "[INFO]: Preparing release ${GITHUB_REF_NAME##*/}"
+          python src/main/scripts/prepare_release.py ${GITHUB_REF_NAME##*/}
+
+      - name: Update README files with helm-docs
```
bianchi2, line 51: Since we update versions in Chart.yaml it is a good idea to refresh docs too

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,71 @@
...
+on:
+  push:
+    branches:
+      - release/*
+
+jobs:
+  prepare-release:
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      PYTHONPATH: /opt/hostedtoolcache/Python/3.9.14/x64/lib/python3.9/site-packages
```
nanux, line 12: Why this need to be defined explicitly?

.github/workflows/prepare-release.yaml
nanux: Nice!

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,106 @@
+name: Release Helm Charts
+
+on:
+  workflow_dispatch:
```
nanux, line 4: The expectation is that after merging the PR that was generated by the action above, the release manager will go and trigger this action manually. I think that is completely fine for now, I am just confirming.

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,106 @@
...
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
+      pull-requests: write
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      HELM_SIGNING_KEY: ${{ secrets.TEST_SIGN_HELM }}
```
nanux, line 15: Is the signing working without issues?

src/main/scripts/github_asset_uploader.sh
```diff
@@ -0,0 +1,42 @@
...
+set -e
+
+if [[ -z $GH_TOKEN ]]; then
+  echo "Must supply a github token. Export GH_TOKEN env variable in the shell where the script is executed"
+  exit 1
+fi
+
+RELEASE_VERSION=$1
+if [[ -z $RELEASE_VERSION ]]; then
+  echo "Must specify release version, e.g. 1.10.2"
```
nanux, line 11:
```suggestion
  echo "Must pass release version as an argument, e.g. 1.10.2"
```

src/main/scripts/github_asset_uploader.sh
```diff
@@ -0,0 +1,42 @@
...
+  exit 1
+fi
+
+if [[ -z $GITHUB_REPOSITORY ]]; then
+  echo "GITHUB_REPOSITORY env var is not set. Using the default atlassian/data/center=helm-charts"
```
nanux, line 16:
```suggestion
  echo "GITHUB_REPOSITORY env var is not set. Using the default atlassian/data/center-helm-charts"
```

src/main/scripts/github_asset_uploader.sh
```diff
@@ -0,0 +1,42 @@
...
+  export GITHUB_REPOSITORY="atlassian/data/center=helm-charts"
```
nanux, line 17:
```suggestion
  export GITHUB_REPOSITORY="atlassian/data/center-helm-charts"
```

src/main/scripts/prepare_release.py
nanux: This might become a list in future.

src/main/scripts/prepare_release.py
nanux: I think this can be hoisted up to a "constant" so it is clear where to change it in case we are changing or adding Jira ticket key.

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,183 @@
...
+    return list(dict.fromkeys(sanitized_changelog))
+
+
+def format_changelog_yaml(changelog):
+    # The artifacthub annotations are a single string, but formatted
+    # like YAML. Replacing the leading '*' with '-'  and wrapping
+    # strings in double quotes should be sufficient.
+    sanitized_changelog = []
+    for string in changelog:
+        string = string + '"'
```
nanux, line 75: Should it be from both sides so it works for the first one as well?

src/main/scripts/prepare_release.py
```diff
@@ -0,0 +1,183 @@
+import git
```
nanux, line 1: Reading this script is not the easiest. I think it is ok as it is but maybe some readability finessing would be useful. For example, the short variable naming (without `_`) makes it harder to follow the code flow and it is not fully self-documenting. It would be also good to add a comment on the top describing the general functionality and processing flow.

Nevertheless, great job mate! This will save a lot of time and avoid errors.

nanux (approved):

src/main/scripts/prepare_release.py
bianchi2: I agree about lack of docs and some readability. I'll make it a bit nicer.

src/main/scripts/prepare_release.py
bianchi2: @nanux Do you mean the regex itself deserves to be in a variable?

src/main/scripts/prepare_release.py
bianchi2: @nanux so the original string is `* I am a string`, and first it is transformed into `* I am a string"` and then to `- "I am a string"` to fix the yaml annotations.

.github/workflows/release-charts.yaml
bianchi2: Yes, correct. Even though we can run it on every PR to main (or if it's merge common from release branch) having more control is preferable for now. So, manual run by a release engineer.

.github/workflows/release-charts.yaml
bianchi2: Leftovers from my test repo. Removed

src/main/scripts/github_asset_uploader.sh
bianchi2: Replaced with export GITHUB_REPOSITORY="${GITHUB_REPOSITORY}"

src/main/scripts/prepare_release.py
nanux: Yes :)

nanux:
Thanks @bianchi2 for the readability improvements 🥇 

🟢

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,104 @@
...
+        with:
+          aws-region: ${{ secrets.HELM_SIGNING_KEY_AWS_REGION }}
+          role-to-assume: ${{ secrets.HELM_SIGNING_KEY_AWS_IAM_ROLE }}
+          role-session-name: HelmSignSession
+
+      - name: Get Helm Signing Key
+        run: |
+          aws secretsmanager get-secret-value --secret-id ${{ secrets.HELM_SIGNING_KEY_AWS_SECRET_ID }} --query 'SecretBinary' --output text --region ${{ secrets.HELM_SIGNING_KEY_AWS_REGION }} | base64 --decode > helm_key
+          gpg --import helm_key
+          gpg --export > ~/.gnupg/pubring.gpg
```
yzha645, line 79: We might not need to export pub key ring to the legacy format here. As it's already exported on the second next line below.

yzha645 (approved):

.github/workflows/release-charts.yaml
bianchi2: Right, this export is needed if we need to verify the package. I think I left it for a possible future test case where we may want to run helm verify after signing (not implemented yet)

PR #523, opened 2023-02-20
Skip mesh deployment to KITT

bianchi2:
Just minor cleanup to fix CI. KITT values were hardcoded for testing purposes. Mesh tests run in a different Bamboo plan

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/src/test/config/bitbucket/values-KITT.yaml
+++ b/src/test/config/bitbucket/values-KITT.yaml
@@ -30,10 +30,3 @@ bitbucket:
   mesh:
     podAnnotations:
       "atlassian.com/business_unit": "server_engineering"
-    enabled: true
-    nodeAutoRegistration: true
-    setByDefault: true
-
-image:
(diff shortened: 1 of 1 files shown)


src/test/config/bitbucket/values-KITT.yaml
```diff
@@ -30,10 +30,3 @@ bitbucket:
   mesh:
```
nanux, line 30: I think we could remove the whole mesh section, but this is good as well.

nanux (changes requested):

nanux (approved):

src/test/config/bitbucket/values-KITT.yaml
bianchi2: We need pod annotations since the plan that tests mesh needs it when Bitbucket is deployed to KITT.

PR #526, opened 2023-02-28
CLIP-1748: Update README.md when bumping appVersions

bianchi2:
Adding an additional step to update README.md since this action changes appVersion, and helm-docs uses data from Chart.yaml to update README.md

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/.github/workflows/update-lts-tags.yaml
+++ b/.github/workflows/update-lts-tags.yaml
@@ -25,12 +25,22 @@ jobs:
           python -m pip install --upgrade pip
           pip install pyyaml requests
 
+      - name: Install Helm Docs
+        uses: envoy/install-helm-docs@v1.0.0
+        with:
+          version: 1.11.0
+
       - name: Configure Git
         run: |
           git config user.name "$GITHUB_ACTOR"
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
       - name: Update LTS tags in Helm charts
+        run: |
+          cd src/test/scripts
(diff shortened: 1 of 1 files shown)


.github/workflows/update-lts-tags.yaml
```diff
@@ -35,6 +40,11 @@ jobs:
           cd src/test/scripts
           python update_versions.py
 
+      - name: Update README files with helm-docs
+        run: |
+          cd src/main/charts
+          helm-docs
```
nanux, line 46: I think we need to add the template file argument here. We define it for the pre-commit hook as well: https://github.com/atlassian/data-center-helm-charts/blob/main/.pre-commit-config.yaml#LL11C24-L11C24

```
          - --template-files=./README.md.gotmpl
```

nanux (approved):

.github/workflows/update-lts-tags.yaml
bianchi2: @nanux yes, but we use the default tpl name, and:

```
  -t, --template-files strings       gotemplate file paths relative to each chart directory from which documentation will be generated (default [README.md.gotmpl])
```

nanux: So we can remove it from the `pre-commit-config.yaml`?

bianchi2: As long as our tpl file matches the default that helm-docs cli has

PR #538, opened 2023-03-22
CLIP-1782: Improve release script and add unit tests

bianchi2:
This PR makes git log filtering in the release script a bit smarter. Once a week, [update-lts-tags workflow](https://github.com/atlassian/data-center-helm-charts/blob/main/.github/workflows/update-lts-tags.yaml) raises PRs to bump LTS appVersions. The bot writes the same commit message - `Update appVersions for DC apps`. However, when a PR is merged, by default PR number enclosed in brackets is appended to the commit message, go in git log we see `Update appVersions for DC apps (#533)`. These commits make it to Git log, and we need just one in the release notes - the latest.

Also, adding a few simple unit tests to make sure the script is doing the right job by:

* removing duplicate commit messages
* removing Update appVersions for DC apps with lower PR number, i.e. only one is left
* removing ira keys from commit messages
* applying the default changelog if git log is empty
* formatting changelog according to artifacthub.io requirements (enclosing in double quotes, replacing * with - etc)

In addition, adding a GH action to run the tests on every push to any branch if any of the release script files are changed. Here's the action: https://github.com/atlassian/data-center-helm-charts/actions/runs/4494167510/jobs/7906310139?pr=538#step:5:1

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/scripts/prepare_release.py
+++ b/src/main/scripts/prepare_release.py
@@ -29,7 +29,7 @@ prod_base = "src/main/charts"
 
 jira_keys_pattern = r'(CLIP|DCCLIP)-[0-9]{1,5}(?![0-9]): '
 drop_commits_pattern = r'^\* Prepare release [0-9].{1,4}'
-
+update_app_versions_commit_msg = r'(\* Update appVersions for DC apps|.*#(\d+))'
 
 # parse Chart.yaml to get K8s version, app version and Helm chart version
 # to use those when generating changelog/release notes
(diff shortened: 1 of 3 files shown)


src/main/scripts/prepare_release.py
```diff
@@ -53,7 +53,7 @@ def get_chart_versions(path=prod_base):
     return versions
 
 
-def gen_changelog(product, path):
+def gen_changelog(product, path, changelog=None, test=False):
```
bianchi2, line 56: Making it possible for the func to get a dummy gitlog from tests

jjeongatl (approved):

PR #548, opened 2023-03-28
CLIP-1784: Multiple fixes to GitHub Actions

bianchi2:
Current problems:

* KinD tests fail when trigged by PRs created from forks because secrets are not available for actions triggered by `pull_request`.
* e2e tests triggered by labeled PRs checkout code from main branch in this repo rather than a fork

With this PR merged, to run KinD tests, `kind` label needs to be added to PR (or one can manually start this workflow). Also, e2e tests will checkout code from the fork and actually test the changes

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/.github/workflows/e2e-tf-deployment.yaml
+++ b/.github/workflows/e2e-tf-deployment.yaml
@@ -40,9 +40,11 @@ jobs:
 
     steps:
       - name: Checkout Helm charts
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.head_ref }}
 
       - name: Pin Kubectl version
         uses: azure/setup-kubectl@v3.0
--- a/src/test/config/kind/common-values.yaml
+++ b/src/test/config/kind/common-values.yaml
@@ -21,23 +21,6 @@ DC_APP_REPLACEME:
   # this applies to Bamboo only and will be ignored in other Helm charts
   disableAgentAuth: true
 
-  # applicable to Bitbucket only
-  mesh:
-    enabled: true
-    nodeAutoRegistration: true
-    setByDefault: true
-    resources:
-      jvm:
-        maxHeap: "1g"
-        minHeap: "512m"
-      container:
-        requests:
-          cpu: "20m"
-          memory: "1G"
-        limits:
-          cpu: "2"
-          memory: "2G"
-
 database:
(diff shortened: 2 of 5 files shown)


.github/workflows/e2e-tf-deployment.yaml
```diff
@@ -40,9 +40,11 @@ jobs:
 
     steps:
       - name: Checkout Helm charts
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.head_ref }}
```
nanux, line 47: Did we really miss this? At some point, I believe we have verified that we are checking out the branch and testing the changes. Maybe it got lost in the ether.

Good catch.

src/test/config/kind/common-values.yaml
```diff
@@ -21,23 +21,6 @@ DC_APP_REPLACEME:
   # this applies to Bamboo only and will be ignored in other Helm charts
   disableAgentAuth: true
 
-  # applicable to Bitbucket only
```
nanux, line 24: Why we don't need this anymore?

src/test/scripts/kind/deploy_app.sh
```diff
@@ -43,10 +43,7 @@ deploy_app() {
   fi
   sed -i "s/DC_APP_REPLACEME/${DC_APP}/g" ../../../test/config/kind/common-values.yaml
   sed -i "s/DB_TYPE_REPLACEME/${DB_TYPE}/g" ../../../test/config/kind/common-values.yaml
-  if [ ${DC_APP} == "bitbucket" ]; then
-    # this is to test mesh
-    IMAGE_OVERRIDE="--set image.tag=8.6.1"
```
nanux, line 48: Was this due to the mesh?

nanux (approved):

src/test/scripts/kind/deploy_app.sh
bianchi2: Yes, leftovers. We test mesh in a separate Bamboo plan now.

.github/workflows/e2e-tf-deployment.yaml
bianchi2: It was ok for branches in the repo, but not forks.

src/test/config/kind/common-values.yaml
bianchi2: That's mesh testing leftovers

PR #549, opened 2023-03-28
Fix upgrade docs

bianchi2:
This PR fixes upgrade instructions that suggest the use of `--reuse-values` when upgrading Helm **chart version**. When `--reuse-values` is passed, the default values.yaml packaged in the Helm chart is ignored. As a result, if it contains some new default values that are used in templates, upgrade to a new version will fail with a templating error. We're not always adding defaults in the templates themselves.

It's ok to use `--reuse-values` when upgrading appVersion only.

## Checklist
- [ ] I have added unit tests


nanux (approved):

bianchi2:
@nanux yes you got it right, `--reuse-values` breaks update from 1.5.0 to, say, 1.8.1 because of that.

PR #550, opened 2023-03-30
Add a warning to Crowd's values.yaml ingress.path field

bianchi2:
Since it is currently not possible to change context path in Crowd's Tomcat, it makes sense to add a warning in values.yaml

## Checklist
- [ ] I have added unit tests


nanux (approved):

nanux:
Is this an internal Crowd limitation? How come it does still support two values? `/` and `/crowd`?

bianchi2:
https://confluence.atlassian.com/crowd/removing-the-crowd-context-from-the-application-url-223217080.html this is the official doc on how to change context, and we need to figure out what exactly the script changes and how to do that in the image entrypoint with jinja.

Both / and /crowd work, since when you go to / for the first time you'll see the page:

image

and Configure Crowd button takes you to /crowd anyway. Once configured, your context will be /crowd anyways.

If you go directly to /crowd during the first installation, you'll see configuration page right away. As a result, in this case it does not matter if you have / or /crowd in the ingress path since the Tomcat's context is still /crowd.

PR #554, opened 2023-04-12
Crowd service type edge case documentation

jjeongatl:
## Pull request description

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/docs/docs/troubleshooting/LIMITATIONS.md
+++ b/docs/docs/troubleshooting/LIMITATIONS.md
@@ -61,6 +61,10 @@ import:
 
 Using this approach will restore the full dataset as part of the Helm install process.
 
+## Crowd
+
+### LoadBalancer service type
+If a kubernetes environment has multiple cluster nodes and `loadBalancer` is used for Crowd `service.type`, Crowd pods may receive client requests from different cluster node ip address. This will cause a session expiry in the client browser. You can resolve it by unticking "Require consistent client IP address" in Session configuration.
 ## Platform limitations
 These configurations are explicitly not supported, and the Helm charts don’t work without modifications in these environments:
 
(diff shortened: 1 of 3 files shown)


docs/docs/troubleshooting/LIMITATIONS.md
```diff
@@ -61,6 +61,10 @@ import:
 
 Using this approach will restore the full dataset as part of the Helm install process.
 
+## Crowd
+
+### LoadBalancer service type
+If a kubernetes environment has multiple cluster nodes and `loadBalancer` is used for Crowd `service.type`, Crowd pods may receive client requests from different cluster node ip address. This will cause a session expiry in the client browser. You can resolve it by unticking "Require consistent client IP address" in Session configuration.
```
bianchi2, line 67: Perhaps we want to say that an alternative way to fix it would be enabling session affinity for the LB, which will result in requests going to the same node:port. Sure thing, when a node is gone the request will go to a different node and this may cause session invalidation

bianchi2, line 67:
```suggestion
If a Kubernetes cluster has multiple cluster nodes and Crowd `service.type` is `LoadBalancer`, Crowd pods may receive client requests from different cluster node IPs. This will cause session invalidation in the client browser, which is a security feature of Crowd. You can resolve it by either:

* Unticking "Require consistent client IP address" in Session configuration or
* Configuring AWS LoadBalancer with session affinity by annotating a service. You can find more details in [AWS LoadBalancer documentation](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/).
```

bianchi2 (approved):

PR #559, opened 2023-04-18
DCCLIP-481: Add common jmx yaml snippets to common chart

bianchi2:
This PR adds jmx related yamls snippets that will be used by all DC charts in the next version.

This is the proposed monitoring in values.yaml:

```
# Monitoring
#
monitoring:
  
  # -- Expose JMX metrics with jmx_exporter https://github.com/prometheus/jmx_exporter
  #
  exposeJmxMetrics: false
  
  # -- Fetch jmx_exporter jar from the image. If set to false make sure to manually copy the jar
  # to shared home and provide an absolute path in jmxExporterCustomJarLocation 
  #
  fetchJmxExporterJar: true
  
  # -- Image repository with jmx_exporter jar
  #
  jmxExporterImageRepo: bitnami/jmx-exporter
  
  # -- Image tag to be used to pull jmxExporterImageRepo
  #
  jmxExporterImageTag: 0.18.0
  
  # -- Port number on which metrics will be available
  #
  jmxExporterPort: 8099
      
  # -- Location of jmx_exporter jar file if mounted from a secret or manually copied to shared home
  #
  jmxExporterCustomJarLocation:
  
  # -- Custom jmx config with the rules. Make sure to keep jmx-config key
  #
  jmxExporterCustomConfig: {}
  # jmx-config: |
  #   rules:
  #     - pattern: ".*"
      
  # -- Create ServiceMonitor. See: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md
  createServiceMonitor: false
```

## Checklist
- [ ] I have added unit tests


src/main/charts/common/templates/_jmx.tpl
```diff
@@ -0,0 +1,57 @@
...
+{{- end }}
+
+{{/*
+Jmx config volume mount
+*/}}
+
+{{- define "common.jmx.config.volumeMounts" -}}
+{{- if .Values.monitoring.exposeJmxMetrics }}
+- name: jmx-config
+  mountPath: /opt/atlassian/confluence/jmx
```
jjeongatl, line 19: can we use confluence in the path if this is for all DC products?

bianchi2: Good catch! Even though it's not a big deal and the mount will happen anyways, it needs to be smth generic. I'll fix that!

bianchi2: Fixed. It's just `/opt/atlassian/jmx` now

jjeongatl (approved):

PR #570, opened 2023-05-02
Enable Bamboo JMS traffic via Kubernetes Service

mjdazz:
## Pull request description

This enables forwarding traffic to the pod's JMS port via Kubernetes service. There is a new variable `bamboo.service.jmsPort` to configure it. It enables traffic from remote agents outside of the Kubernetes cluster Bamboo is running in the same way SSH git connections are possible as explained here: https://atlassian.github.io/data-center-helm-charts/examples/bitbucket/BITBUCKET_SSH/

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/bamboo/README.md
+++ b/src/main/charts/bamboo/README.md
@@ -44,6 +44,11 @@ Kubernetes: `>=1.21.x-0`
 | bamboo.import | object | `{"path":null,"type":"clean"}` | Bamboo can optionally import an existing exported dataset on first-run. These optional values can configure the import file or skip this stage entirely. For more details on importing and exporting see the documentation:  https://confluence.atlassian.com/bamboo/exporting-data-for-backup-289277255.html https://confluence.atlassian.com/bamboo/importing-data-from-backup-289277260.html  |
 | bamboo.import.path | string | `nil` | Path to the existing export to import to the new installation. This should be accessible by the cluster node; e.g. via the shared-home or `additionalVolumeMounts` below.  |
 | bamboo.import.type | string | `"clean"` | Import type. Valid values are `clean` (for a new install) or `import`, in which case you should provide the file path. The default is `clean`.  |
+| bamboo.jmsService.annotations | object | `{}` | Additional annotations to apply to the JMS Service  |
+| bamboo.jmsService.enabled | bool | `false` | Whether to create a separate Service for JMS Agent traffic  |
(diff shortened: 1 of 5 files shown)


bianchi2:
@mjdazz thanks for your PR. Looks reasonable, there's one thing though. Do you think it will make more sense to create a separate service for jms to be able to control svc type and annotations separately for http and jms traffic?

mjdazz:
That sounds very reasonable. I'm unsure about the naming, though. Would you be fine with `bamboo.jmsService` set up the same way as `bamboo.service` ?

If so, I'll go ahead and change it accordingly then.

bianchi2:
I think adding the following under bamboo will make sense:

```
bamboo:
  jmsService:

    # -- The port on which the Bamboo JMS Service will listen
    #
    port: 54663

    # -- The type of K8s service to use for Bamboo JMS traffic
    #
    type: ClusterIP

    # -- Use specific loadBalancerIP. Only applies to service type LoadBalancer.
    #
    loadBalancerIP:
```
And then create a new template, say, service-jms.yaml:

```
apiVersion: v1
kind: Service
metadata:
  name: {{ include "common.names.fullname" . }}-jms
  labels:
  {{- include "common.labels.commonLabels" . | nindent 4 }}
  annotations:
  {{- with .Values.bamboo.jmsService.annotations }}
  {{- toYaml . | nindent 4 }}
  {{- end }}
spec:
  type: {{ .Values.bamboo.jmsService.type }}
  {{- if and (eq .Values.bamboo.jmsService.type "LoadBalancer") (not (empty .Values.bamboo.jmsService.loadBalancerIP)) }}
  loadBalancerIP: {{ .Values.bamboo.jmsService.loadBalancerIP }}
  {{- end }}
  ports:
    - port: {{ .Values.bamboo.jmsService.port }}
      targetPort: jms
      protocol: TCP
      name: jms
  selector:
  {{- include "common.labels.selectorLabels" . | nindent 4 }}
```

src/main/charts/bamboo/README.md
```diff
@@ -1,4 +1,4 @@
-# bamboo
+g bamboo
```
bianchi2, line 1: Typo?

bianchi2:
thanks @mjdazz can you sign CLA please?

bianchi2:
thanks @mjdazz good to merge for me. We'll have to get rid of this warning in docs once merged https://github.com/atlassian/data-center-helm-charts/blob/main/docs/docs/troubleshooting/LIMITATIONS.md#server-and-agent-affinity

bianchi2 (approved):

PR #573, opened 2023-05-10
DCCLIP-501: Add optional ServiceMonitors to DC Helm Charts

bianchi2:
This PR adds [ServiceMonitors](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.ServiceMonitor) to DC Helm charts templates.

By default `monitoring.serviceMonitor.create=false` so it needs to be explicitly set to `true`.
Helm charts do not create `ServiceMonitor` CRDs, so it is expected that Kube API knows what a ServiceMonitor is (otherwise helm install/upgrade will fail).

Deploying [Kube prometheus stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) is one of the ways to deploy this CRD (but not the only one).

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/bamboo/templates/service-monitor.yaml
+++ b/src/main/charts/bamboo/templates/service-monitor.yaml
@@ -9,7 +9,7 @@ metadata:
   {{- end }}
 spec:
   endpoints:
-  - interval: {{ printf "%ds" (default 30 .Values.monitoring.serviceMonitor.scrapeIntervalSeconds)}}
+  - interval: {{ printf "%ds" .Values.monitoring.serviceMonitor.scrapeIntervalSeconds | quote }}
     path: /metrics
     port: jmx
     scheme: http
--- a/src/main/charts/bamboo/templates/service-monitor.yaml
+++ b/src/main/charts/bamboo/templates/service-monitor.yaml
@@ -1,0 +1,19 @@
+{{- if .Values.monitoring.serviceMonitor.create }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "common.names.fullname" . }}-service-monitor
+  labels:
+  {{- with .Values.monitoring.serviceMonitor.prometheusLabelSelector }}
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  endpoints:
+  - interval: {{ printf "%ds" (default 30 .Values.monitoring.serviceMonitor.scrapeIntervalSeconds)}}
+    path: /metrics
+    port: jmx
+    scheme: http
+  selector:
+    matchLabels:
+    {{- include "common.labels.selectorLabels" . | nindent 6 }}
+{{- end }}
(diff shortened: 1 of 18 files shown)


src/main/charts/bamboo/templates/service-monitor.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: ServiceMonitor
+metadata:
+  name: {{ include "common.names.fullname" . }}-service-monitor
+  labels:
+  {{- with .Values.monitoring.serviceMonitor.prometheusLabelSelector }}
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  endpoints:
+  - interval: {{ printf "%ds" (default 30 .Values.monitoring.serviceMonitor.scrapeIntervalSeconds)}}
```
nanux, line 12: Can we move the default into the `values.yaml` file to make it more visible?

src/main/charts/bamboo/values.yaml
```diff
@@ -784,6 +784,21 @@ monitoring:
...
+    create: false
+
+    # -- ServiceMonitorSelector of the prometheus instance.
+    #
+    prometheusLabelSelector: {}
+      # release: prometheus
+
+    # -- Scrape interval for the JMX service. Defaults to 30 seconds
+    #
+    scrapeIntervalSeconds:
```
nanux, line 800:
```suggestion
    scrapeIntervalSeconds: 30
```

Then we can drop the second sentence in the comment.

src/test/java/test/JmxMetricsTest.java
```diff
@@ -177,4 +179,50 @@ void expose_jmx_metrics_bitbucket_mesh(Product product) throws Exception {
...
+
+    @ParameterizedTest
+    @EnumSource(value = Product.class, names = {"bitbucket"}, mode = EnumSource.Mode.INCLUDE)
+    void service_monitor_bitbucket_mesh(Product product) throws Exception {
+        final var resources = helm.captureKubeResourcesFromHelmChart(product, Map.of(
+                "monitoring.serviceMonitor.create", "true",
+                product + ".mesh.enabled", "true"
+        ));
+
+        for (int i = 0; i < 3; i++) {
```
nanux, line 191: /nit This relies on the default `3`. To be safe, I would add `.mesh.replicaCount = 3` to the values. It makes it easier to understand why we check 3.

nanux (approved):

src/test/java/test/JmxMetricsTest.java
bianchi2: I don't think we'll change the default but yes, setting replicaCount is the right thing to do.

src/main/charts/bamboo/values.yaml
bianchi2: Added defaults in values.yaml

src/main/charts/bamboo/templates/service-monitor.yaml
bianchi2: I'd leave it here just in case someone runs helm upgrade with --reuse-values and the default values.yaml will be ignored.

PR #581, opened 2023-05-26
DCCLIP-511: Grafana dashboard conversion script

bianchi2:
The script adds the following to an existing dashboard:

* namespace and service to every query
* namespace and service as env vars to templating.list in the json, along with prometheus datasource

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/scripts/process_dashboard.py
+++ b/src/main/scripts/process_dashboard.py
@@ -24,17 +24,17 @@ parser.add_argument("--product", help="Product")
 parser.add_argument("--mesh", help="sidecar")
 args = parser.parse_args()
 
-if args.source is None:
-    print("Error: --source argument is required. This must be URL of a raw JSON file.")
-    sys.exit(1)
-
-if args.dest is None:
-    print("Error: --dest argument is required. This must be an absolute path to the destination json file.")
-    sys.exit(1)
-
-if args.product is None:
-    print("Error: --product argument is required. This must be the product of the dashboard to be converted.")
-    sys.exit(1)
+# if args.source is None:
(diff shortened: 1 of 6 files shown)


src/main/scripts/process_dashboard.py
```diff
@@ -0,0 +1,99 @@
+import json
```
nanux, line 1: Can we have some general comment/documentation what is the purpose of the script and what dashboards it can process (e.g. as Github links).

Also can we have a test for it? Just defining one URL and defining an expected output?

nanux (approved):

nanux:
I think the script looks fine.

Can we change the existing dashboards to make them kubernetes friendly? Are they used by anybody?

src/main/scripts/process_dashboard.py
```diff
@@ -0,0 +1,99 @@
...
+import urllib.request
+import argparse
+
+parser = argparse.ArgumentParser(description="Read remote JSON file")
+parser.add_argument("--source", help="URL of the target remote JSON file")
+parser.add_argument("--dest", help="Local path of the destination processed json")
+parser.add_argument("--product", help="Product")
+args = parser.parse_args()
+
+if args.source is None:
```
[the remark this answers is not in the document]
bianchi2: Agreed but this script is just for CI

src/main/scripts/process_dashboard.py
bianchi2: Yes, we can test it for sure with a small dashboard json.

bianchi2: Added a test which calls the script and then asserts a few things in the resulted json

PR #586, opened 2023-06-02
Update Grafana dashboards for Bitbucket

bianchi2:
This PR adds endpoint="jmx" filter to all dashboards, except for ticket-hosting.json where we're interested in metrics from mesh sidecar that runs in a different JVM and is exposed on 9998.

Endpoints are defined in serviceMonitors (and in case of Bitbucket we have 2 endpoints - one for the server and one for the mesh)

Without this change all metrics are duplicated:

image

## Checklist
- [ ] I have added unit tests


errcode1202 (approved):

src/main/charts/bitbucket/templates/configmap-mesh-grafana-dashboards.yaml
```diff
@@ -6,14 +6,14 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
 {{- $fileName := split "/" $index }}
-  name: {{ include "common.names.fullname" . }}-{{ $fileName._1 | trimSuffix ".json" }}-dashboard
+  name: {{ include "common.names.fullname" . }}-{{ $fileName._2 | trimSuffix ".json" }}-dashboard
```
bianchi2, line 9: mesh dashboards are in a subdirectory

PR #592, opened 2023-06-06
CLIP-1806: Generate k8s support bundle

errcode1202:
## Pull request description

A utility script that can be used for generating a "support" bundle to facilitate investigations into problematic deployments. The details captured include:

- Pod status
- Pod logs
- Ingress controller (NGINX) logs `(optional)`
- Events logs
- resource descriptions, for
  - svc
  - ingress
  - pvc
  - pv
- Node descriptions.
- Application logs `(optional)`

Usage:
```bash
generate_k8s_support_bundle.sh -n <namespace>
```

The result of running this is the archive `k8s-support.tar.gz` located in the directory in which the script was run.

The script also allows for additional detail to be optionally added, namely:

```bash
options:
-a     Include application logs.
-i     Include ingress-nginx logs. Supply ingess controller namespace with this flag.
-o     Include node logs.
-h     Print help.
```


scripts/generate_k8s_support_bundle.sh
```diff
@@ -0,0 +1,179 @@
...
+get_resource_data() {
+  echo "[INFO]: Getting resource descriptions..."
+  RESOURCES=(svc ingress pvc pv)
+  for RESOURCE in "${RESOURCES[@]}"; do
+    echo "[INFO]: Description obtained for ${RESOURCE}"
+    kubectl describe "${RESOURCE}" -n "${NAMESPACE}" > "${ENV_STATS}"/${RESOURCE_LOG}/"${RESOURCE}"_describe.log 2>&1
+  done
+}
+
+get_node_data() {
```
bianchi2, line 85: This assumes that the script is run by someone who has cluster wide privileges. Like with ingress logs, perhaps we should make it optional? IMHO, out of the box we should assume only namespace scoped privileges.

bianchi2 (approved):

scripts/generate_k8s_support_bundle.sh
errcode1202: Made this an optional using `-o`

PR #603, opened 2023-06-20
Add s3 avatars storage support for Jira

asezavar:
## Pull request description

Jira S3 storage support for avatars

## Checklist
- [x] I have added unit tests


bianchi2 (approved):

src/main/charts/jira/values.yaml
```diff
@@ -533,6 +542,24 @@ jira:
     #
     localHomeSubPath: "log"
 
+  # Use AWS S3 to store attachments. From Confluence 8.1 onwards.
+  #
+
+  s3AvatarsStorage:
```
errcode1202, line 548: Just wondering what happens when attachments for s3 needs to be added? Do we create a new stanza like this i.e.

```
s3AttachmentStorage
  bucketName
  bucketRegion
  endpointOverride
```

Or is it better to just rename this stanza to `s3Storage` and its used when storing avatars or attachments or both?

src/main/charts/jira/values.yaml
```diff
@@ -533,6 +542,24 @@ jira:
     #
     localHomeSubPath: "log"
 
+  # Use AWS S3 to store attachments. From Confluence 8.1 onwards.
```
errcode1202, line 545: Comment is Confluence specific

errcode1202 (changes requested):

src/main/charts/jira/values.yaml
bianchi2: I think it is possible that different buckets may be used for avatars and attachments. Maybe:

```
s3Storage:
  avatars:
    bucketName
    bucketRegion
    endpointOverride
  attachments:
    bucketName
    bucketRegion
    endpointOverride
```

src/test/java/test/JiraS3EnabledTest.java
```diff
@@ -0,0 +1,85 @@
...
+                product + ".s3AvatarsStorage.bucketName", "my-bucket",
+                product + ".s3AvatarsStorage.bucketRegion", "my-region"
+        ));
+
+        resources.getStatefulSet(product.getHelmReleaseName())
+                .getContainer()
+                .getEnv()
+                .assertHasValue("ATL_S3AVATARS_BUCKET_NAME", "my-bucket")
+                .assertHasValue("ATL_S3AVATARS_REGION", "my-region")
+                .assertDoesNotHaveAnyOf("ATL_S3AVATARS_ENDPOINT_OVERRIDE");
```
errcode1202, line 68: Nit: maybe this test can be removed (kinda a duplicate of `jira_s3_avatars_storage_env_vars`) and to cover the last assertion add `.assertDoesNotHaveAnyOf("ATL_S3AVATARS_ENDPOINT_OVERRIDE");` to `jira_s3_avatars_storage_env_vars`?

src/main/charts/jira/values.yaml
errcode1202: Oh interesting. Is this an option @asezavar?

bianchi2: Of course, if we expect attachments to arrive soon :)

asezavar: @errcode1202 yes that's the plan

errcode1202: Should the stanza be updated now then to accomadate attachments?

asezavar: I just changed the s3 storage values structure based on your suggestion, but I haven't added the attachment part, and there will be another ticket for that.

bianchi2 (approved):

errcode1202 (approved):
🚀

PR #606, opened 2023-06-28
Run jmx fetcher init container as root

bianchi2:
If shared home isn't writable to other users, the init container will fail to copy jmx jar. The default bitnami image runs as 1001. This init container needs to run as root

## Checklist
- [ ] I have added unit tests


nanux (approved):

src/main/charts/common/templates/_jmx.tpl
```diff
@@ -30,6 +30,8 @@ Jmx init container
   image: {{ .Values.monitoring.jmxExporterImageRepo}}:{{ .Values.monitoring.jmxExporterImageTag}}
   command: ["cp"]
   args: ["/opt/bitnami/jmx-exporter/jmx_prometheus_javaagent.jar", "{{ .Values.volumes.sharedHome.mountPath }}"]
+  securityContext:
```
nanux, line 33: That is the nice part of having this in the common helm chart. Only single change.

nanux (approved):

PR #615, opened 2023-07-10
Synchrony missing container limits

LaszloCsorbaBearingPoint:
Only container requests are supported in the config, but it is also very critical to specify the limits as well so the container doesn't crash other pods as when limits are not specified, the container can use all the resource available on the node.

As we want to make sure other pods are not affected in case of cpu or memory spikes, we need to be able to set the container limits as well.

## Pull request description

Tested this out locally (uncommented):

kubectl get pod synchrony -n confluence -o=jsonpath='{.spec.containers[0].name}'
synchrony%                                                                                                                                                                           
kubectl get pod synchrony -n confluence -o=jsonpath='{.spec.containers[0].resources}'
{"limits":{"cpu":"2","memory":"4G"},"requests":{"cpu":"2","memory":"4G"}}%                                                                                                           
kubectl get pod synchrony -n confluence
NAME                     READY   STATUS    RESTARTS   AGE
synchrony                    2/2   Running    0                   16m

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/confluence/values.yaml
+++ b/src/main/charts/confluence/values.yaml
@@ -1005,9 +1005,9 @@ synchrony:
         # -- Initial Memory request Synchrony pod
         #
         memory: "2.5G"
-    #  limits:
-    #    cpu: "2"
-    #    memory: "2.5G"      
+      #  limits:
+      #    cpu: "2"
+      #    memory: "2.5G"      
 
   # -- Specifies a list of additional arguments that can be passed to the Synchrony JVM, e.g.
   # system properties.
--- a/src/main/charts/confluence/values.yaml
+++ b/src/main/charts/confluence/values.yaml
@@ -1007,7 +1007,7 @@ synchrony:
         memory: "2.5G"
     #  limits:
     #    cpu: "2"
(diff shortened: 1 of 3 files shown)


bianchi2:
@LaszloCsorbaBearingPoint thanks for your contribution. Yes, it absolutely makes sense to be able to define limits for the container, especially to mitigate noisy neighbors problem when a pod can use all allocatable memory and cpu.

src/main/charts/confluence/values.yaml
```diff
@@ -1005,6 +1005,9 @@ synchrony:
         # -- Initial Memory request Synchrony pod
         #
         memory: "2.5G"
+      #  limits:
```
bianchi2, line 1008: @LaszloCsorbaBearingPoint can you please fix indentation here so that when you uncomment limits are on the same level with requests?

LaszloCsorbaBearingPoint: Sure, I just did that, and I also fixed the same for the Confluence container as well as it was the same (from where I copied over the same syntax)

bianchi2 (approved):

PR #623, opened 2023-07-20
Support running Jira and Confluence containers as non root (incl running in OpenShift)

bianchi2:
# Motivation

Currently, when running containers as nonRoot, there are issues with config files generation because existing entrypoint in Docker images assumes that the generation is executed by a root user. As a result, files like server.xml can't be generated, and they will miss important props like proxyName and proxyPort. Also, it is not possible to override any props like maxThreads etc.

In addition to that, when running on OpenShift, as an unprivileged user, there are permission issues with writing to Tomcat's temp directory, which results in critical errors (observed in Confluence but Jira is likely the victim too).

# What This PR Does

* optional `tomcatConfig` stanza is introduced
* server.xml and seraph-config.xml are ConfigMap in the templates
* a runtime volume is added for tomcat's temp directory (to let an unprivileged user write to it. We assume if server.xml is to be mounted as a configmap, it means the container isn't run as root)

Users will be able to set:

```
confluence:
  tomcatConfig:
    generateByHelm: true
```
and server.xml will be mounted to Tomcat's conf directory. All important properties are configurable and can be overridden in values.yaml.

This is done for seraph-config.xml as well (though there's just one value that can be overridden there - max cookie age). This will not break any existing security best practices like keeping Tomcat's conf directory root owned (and mounted ConfigMaps are read only)

Mounting server.xml as ConfigMap is optional and false by default, so nothing changes for existing users. Changes to server.xml in the Docker repos are very rare, so it's unlikely to be a burden to keep helm templates up to date with what's in Docker repos.

# Who Will Benefit From It

**OpenShift** first and foremost. ➕ any users who override `securityContext`, and set, say, `runAsUser: 2001`, or `runAsNonRoot: true `

## Checklist
- [x] I have added unit tests


src/main/charts/confluence/values.yaml
```diff
@@ -736,6 +736,58 @@ confluence:
   #
   additionalJvmArgs: []
 
+  # -- By default Tomcat's server.xml is generated in the container entrypoint from a template
+  # shipped with an official Confluence image. However, server.xml generation may fail if container
+  # is not run as root, which is a common case if Confluence is deployed to OpenShift.
+  #
+  tomcatConfig:
+
+    # -- Mount server.xml as a ConfigMap. Uncomment configuration elements if necessary to
```
yzha645, line 745: > Uncomment configuration elements if necessary 

Should the below elements be commented out then?

bianchi2: @yzha645 they were, but then I decided to keep them uncommented. I'll change the comment to Override configuration elements if necessary.

src/main/charts/confluence/values.yaml
```diff
@@ -736,6 +736,57 @@ confluence:
...
+  # shipped with an official Confluence image. However, server.xml generation may fail if container
+  # is not run as root, which is a common case if Confluence is deployed to OpenShift.
+  #
+  tomcatConfig:
+
+    # -- Mount server.xml as a ConfigMap. Override configuration elements if necessary
+    #
+    generateByHelm: false
+
+    mgmtPort: "8005"
```
yzha645, line 749: nit: Is it intended that `mgmtPort` and `maxThreads` have different value from default ones?

bianchi2: @yzha645 thanks for catching that! Pocy-paste from Jira. Fixed.

yzha645:
Question, will `filestore-config.xml` be failing to be created as well?

bianchi2:
>Question, will [filestore-config.xml](https://bitbucket.org/atlassian-docker/docker-atlassian-jira/src/master/config/filestore-config.xml.j2) be failing to be created as well?

@yzha645 This file is saved to [local-home](https://bitbucket.org/atlassian-docker/docker-atlassian-jira/src/9cc3476745203c7610288dc0cd1f4487d6d11976/entrypoint.py#lines-25), which we expect to be writable (kubelet or CSI driver will make sure it's writable as per `securityContext.fsGroup`)

yzha645 (approved):

Issue #579, opened 2023-05-15
[Suggestion] - Add additionalFiles to the Bamboo-Agent Helm Chart to mount configMaps for config files like settings.xml etc.


steeu:
### Suggestion

Please add the additionalFiles property to the values.yaml as it ist implemented in the Bamboo chart for the Bamboo-Agent chart as well. This way, config files can be mounted and edited easily with configMaps.

### Example from the Bamboo chart:
https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bamboo/values.yaml
```
# -- Additional existing ConfigMaps and Secrets not managed by Helm that should be
# mounted into service container. Configuration details below (camelCase is important!):
  # 'name'      - References existing ConfigMap or secret name.
  # 'type'      - 'configMap' or 'secret'
  # 'key'       - The file name.
  # 'mountPath' - The destination directory in a container.
# VolumeMount and Volumes are added with this name and index position, for example;
# custom-config-0, keystore-2
#
additionalFiles: []
#  - name: custom-config
#    type: configMap
#    key: log4j.properties
#    mountPath:  /var/atlassian
```

### Product

Other

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bianchi2:
@steeu better late than never 😃 I have raised a PR to add additionalFiles to bamboo-agent, and it'll be in the next release that will see the world pretty soon.

PR #625, opened 2023-07-21
Add additionalFiles to bamboo-agent Helm chart

bianchi2:
Resolves https://github.com/atlassian/data-center-helm-charts/issues/579

Mounting files like maven's settings.xml is way easier with additionalFiles rather than with init containers and additional volume(-mounts).

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/bamboo-agent/values.yaml
+++ b/src/main/charts/bamboo-agent/values.yaml
@@ -308,6 +308,29 @@ additionalInitContainers: []
 additionalLabels: {}
 #  name: <value>
 
+# -- Additional existing ConfigMaps and Secrets not managed by Helm that should be
+# mounted into service container. Configuration details below (camelCase is important!):
+  # 'name'      - References existing ConfigMap or secret name.
+  # 'type'      - 'configMap' or 'secret'
+  # 'key'       - The file name.
+  # 'mountPath' - The destination directory in a container.
+# VolumeMount and Volumes are added with this name and index position, for example;
+# custom-config-0, keystore-2
+#
+additionalFiles: []
+#  - name: custom-config
+#    type: configMap
+#    key: log4j.properties
+#    mountPath:  /var/atlassian
+#  - name: custom-config
+#    type: configMap
+#    key: web.xml
+#    mountPath: /var/atlassian
+#  - name: keystore
+#    type: secret
+#    key: keystore.jks
+#    mountPath: /var/ssl
+
 # -- Additional host aliases for each pod, equivalent to adding them to the /etc/hosts file.
 # https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/
 additionalHosts: []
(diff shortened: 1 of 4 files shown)


src/main/charts/bamboo-agent/values.yaml
```diff
@@ -308,6 +308,29 @@ additionalInitContainers: []
 additionalLabels: {}
 #  name: <value>
 
+# -- Additional existing ConfigMaps and Secrets not managed by Helm that should be
+# mounted into service container. Configuration details below (camelCase is important!):
+  # 'name'      - References existing ConfigMap or secret name.
```
nanux, line 313:
```suggestion
  # 'name'      - References existing ConfigMap or Secret name.
```

src/test/java/test/AdditionalFilesTest.java
```diff
@@ -34,8 +35,14 @@ void additional_files_config_map_creates_volumes(Product product) throws Excepti
                 "additionalFiles[0].mountPath", "/var/atlassian"
         ));
 
-        StatefulSet statefulSet = resources.getStatefulSet(product.getHelmReleaseName());
-        JsonNode configMap = getVolume(statefulSet, "configMap");
+        JsonNode configMap;
+        if (product.name().equals("bamboo_agent")) {
```
nanux, line 39: I think this could be a different test - one for deployments (including only Bamboo agent) and the other for StatefulSets. The preparation part of setting up the `resources` could be a private function to remove duplication. I think it is a matter of personal preference, but I don't like `if` statements in the test functions.

Having said that I think this can stay as it is, it is conceptually testing the same thing so it is just a different tradeoff.

src/test/java/test/AdditionalFilesTest.java
```diff
@@ -58,23 +65,30 @@ void additional_files_secret_creates_new_secrets(Product product) throws Excepti
...
-        StatefulSet statefulSet = resources.getStatefulSet(product.getHelmReleaseName());
-        StatefulSet[] stsToCheck = {statefulSet};
-        if (product.name().contains("bitbucket")) {
-            StatefulSet meshStatefulSet = resources.getStatefulSet(product.getHelmReleaseName() + "-mesh");
-            stsToCheck = new StatefulSet[]{statefulSet, meshStatefulSet};
-        }
-
-        for (var sts : stsToCheck) {
-            JsonNode configMap = getVolume(sts, "secret");
+        if (product.name().equals("bamboo_agent")) {
```
nanux, line 68: This one get even more unwieldy with the long `if` and `else`.

nanux (approved):

nanux:
These are just minor comments regarding the test structure and having conditionals in them but I am happy for us to merge this.

src/test/java/test/AdditionalFilesTest.java
bianchi2: Yeah, I was thinking about a dedicated test, though it's the same amount of duplicated code. Perhaps, once we switch to yaml based tests, things will become simpler.

PR #626, opened 2023-07-22
CLIP-1786: Add liveness probes, make readinessProbes configurable

bianchi2:
**What This PR Does**

* add liveness probes to all DC StatefulSets (optional, and can be configured)
* make readinessProbe optional (sometimes it is necessary to get an ingress even though /status reports 500, which can be caused by reindexing)
* make it possible to override readinessProbe (*for example with tcpSocket or exec)
* add tests

## Checklist
- [x] I have added unit tests


src/main/charts/bamboo/values.yaml
```diff
@@ -609,6 +613,46 @@ bamboo:
...
+    failureThreshold: 120
+
+  # Ensure that the server responds with a LivenessProbe
+  # https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request
+  #
+  livenessProbe:
+
+    # -- Whether to apply the livenessProbe check to pod.
+    #
+    enabled: true
```
nanux, line 645: What led us to introduce the Liveness Probe? When we were working on the helm charts, I remember we didn't introduce it because it can end up in a situation when the product is in a restart loop when it becomes unresponsive or slow. When we will add it `enabled: true`  as default, can it potentially affect existing customers that will upgrade and start experiencing new behaviour?

src/main/charts/bitbucket/README.md
```diff
@@ -49,12 +49,15 @@ Kubernetes: `>=1.21.x-0`
...
 | bitbucket.mesh.additionalEnvironmentVariables | object | `{}` | Defines any additional environment variables to be passed to the Bitbucket mesh containers.  |
 | bitbucket.mesh.additionalFiles | string | `nil` | Additional existing ConfigMaps and Secrets not managed by Helm that should be mounted into service container  |
 | bitbucket.mesh.additionalInitContainers | object | `{}` | Additional initContainer definitions that will be added to all Bitbucket pods  |
 | bitbucket.mesh.additionalJvmArgs | list | `[]` | Specifies a list of additional arguments that can be passed to the Bitbucket Mesh JVM, e.g. system properties.  |
 | bitbucket.mesh.affinity | object | `{}` | Standard Kubernetes affinities that will be applied to all Bitbucket mesh pods  |
-| bitbucket.mesh.enabled | bool | `false` | Enable Bitbucket Mesh. See: https://confluence.atlassian.com/bitbucketserver/bitbucket-mesh-1128304351.html  |
+| bitbucket.mesh.enabled | bool | `false` | Enable Bitbucket Mesh. See: https://Bitbucket.atlassian.com/bitbucketserver/bitbucket-mesh-1128304351.html  |
```
nanux, line 60:
```suggestion
| bitbucket.mesh.enabled | bool | `false` | Enable Bitbucket Mesh. See: https://confluence.atlassian.com/bitbucketserver/bitbucket-mesh-1128304351.html  |
```

src/main/charts/bitbucket/values.yaml
```diff
@@ -192,7 +192,7 @@ volumes:
 
   # Each pod requires its own volume for 'local-home'. This is needed for key data
   # that help define how Bitbucket works.
-  # https://confluence.atlassian.com/bitbucketserver/set-the-home-directory-776640890.html
+  # https://Bitbucket.atlassian.com/bitbucketserver/set-the-home-directory-776640890.html
```
nanux, line 195:
```suggestion
  # https://confluence.atlassian.com/bitbucketserver/set-the-home-directory-776640890.html
```

src/main/charts/bitbucket/values.yaml
```diff
@@ -245,7 +245,7 @@ volumes:
   # environment.
   #
   # Details on how an NFS should be stood up for Bitbucket can be found here:
-  # https://confluence.atlassian.com/bitbucketserver/install-bitbucket-data-center-872139817.html#InstallBitbucketDataCenter-nfs
+  # https://Bitbucket.atlassian.com/bitbucketserver/install-bitbucket-data-center-872139817.html#InstallBitbucketDataCenter-nfs
```
nanux, line 248:
```suggestion
  # https://confluence.atlassian.com/bitbucketserver/install-bitbucket-data-center-872139817.html#InstallBitbucketDataCenter-nfs
```

src/main/charts/bitbucket/values.yaml
```diff
@@ -772,7 +830,7 @@ bitbucket:
   # Mesh configuration
   mesh:
 
-    # -- Enable Bitbucket Mesh. See: https://confluence.atlassian.com/bitbucketserver/bitbucket-mesh-1128304351.html
+    # -- Enable Bitbucket Mesh. See: https://Bitbucket.atlassian.com/bitbucketserver/bitbucket-mesh-1128304351.html
```
nanux, line 833:
```suggestion
    # -- Enable Bitbucket Mesh. See: https://confluence.atlassian.com/bitbucketserver/bitbucket-mesh-1128304351.html
```

src/main/charts/bamboo/templates/statefulset.yaml
```diff
@@ -159,13 +159,33 @@ spec:
...
+            {{- else }}
             httpGet:
               port: {{ .Values.bamboo.ports.http }}
               path: {{ .Values.bamboo.service.contextPath }}/rest/api/latest/status
             initialDelaySeconds: {{ .Values.bamboo.readinessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.bamboo.readinessProbe.periodSeconds }}
             failureThreshold: {{ .Values.bamboo.readinessProbe.failureThreshold }}
+            {{- end }}
+          {{- end }}
+          {{- if .Values.bamboo.livenessProbe.enabled }}
```
yzha645, line 177: Do we want to do `startupProbe` only when `livenessProbe` is enabled? Correct me if I'm wrong, it feels to be useful even when defining `startupProbe` alone without `livenessProbe`, given that `starupProbe` blocks `readinessProbe` from executing as well.

src/main/charts/jira/templates/statefulset.yaml
```diff
@@ -120,13 +120,21 @@ spec:
               protocol: TCP
             {{- include "jira.additionalPorts" . | nindent 12 }}
             {{- include "common.jmx.port" . | nindent 12 }}
+          {{- if .Values.jira.readinessProbe.enabled }}
```
nanux, line 123: Jira doesn't support liveness probe on purpose? I think we can have it here as well.

src/main/charts/bamboo/templates/statefulset.yaml
bianchi2: thanks @yzha645! That's a good point. Yes, startupProbe is executed first, before the other two.  I think I've copied what was done for Jira. I'll make startup probes unconditional.

src/test/java/test/ReadinessLivenessProbesTest.java
```diff
@@ -0,0 +1,102 @@
...
+                product.getHelmReleaseName()).getContainer().get("livenessProbe").get("periodSeconds").asText());
+        assertEquals("1", resources.getStatefulSet(
+                product.getHelmReleaseName()).getContainer().get("livenessProbe").get("failureThreshold").asText());
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = Product.class, names = {"bamboo_agent"}, mode = EnumSource.Mode.EXCLUDE)
+    void test_liveness_probe_overrides(Product product) throws Exception {
+        final var resources = helm.captureKubeResourcesFromHelmChart(product, Map.of(
+                product + ".livenessProbe.periodSeconds", "10",
```
nanux, line 59: I would probably use some more random number (let's say we change to `10` in future, then this test won't really test the overrides).

nanux:
Great improvement @bianchi2.

I just wonder what is the implication of adding the Liveness probe as default (for new and existing customers doing upgrade)?

nanux (approved):

src/main/charts/bamboo/templates/statefulset.yaml
bianchi2: Fixed. startupProbe is always there now (which is tcpSocker check)

src/test/java/test/ReadinessLivenessProbesTest.java
bianchi2: Fixed

src/main/charts/jira/templates/statefulset.yaml
bianchi2: It does https://github.com/atlassian/data-center-helm-charts/pull/626/files#diff-9ed02b7507528872403332c2e1d9880d35cf7e6e95faf7446617c633767a12c7R143

src/main/charts/bamboo/values.yaml
bianchi2: LivenessProbe for Jira was introduced in https://github.com/atlassian/data-center-helm-charts/pull/511

Yes, if readinessProbe (tcpSocket) fails, the pod will be restarted, and yes, kubelet will restart it every time it fails. The difference with the readinessProbe is that it's tcpSocket check, not httpGet /status. I think it's fair for kubelet to restart a pod if it cannot open a socket connection to the server port in the pod (which means the application is unresponsive).

Yes, with the default value of true, statefulsets will be updated and pods will restart after helm upgrade. Do you think false is a safer and more conservative/non-disruptive option?

nanux: Is it possible that the server won't open or being responsive on the port at all for some time?

To be frank, I am not sure if we need to stick to the `false` to the default for Liveness Probe. I am just worried about potential unknown implications for large enterprise instances that we don't have testing or exposure of.

bianchi2: > Is it possible that the server won't open or being responsive on the port at all for some time?
that's a possibility, yes. Perhaps, we can make the default `failureThreshold` a bit higher, say, 12? This way, the pod will be restarted if kubelet is unable to open a socket on the server port for 1 minute.

nanux: I think that does make sense - we should have some more lenient default values to potentially don't have regressive impact on existing customer instances. After a bit of :chatting: I am convinced that it might take some time to open the Tomcat connector to make the socket accessible.

> Socket is accessible in Tomcat after it starts. When Tomcat starts up, it initializes the connectors for the specified ports and starts listening for incoming connections on those ports. Once Tomcat has completed its startup process, including loading the web applications and initializing the servlet contexts, the socket is accessible and ready to accept incoming connections.

> Keep in mind that the actual time for the socket to become accessible may vary depending on factors such as the size of your application, the number of web applications being deployed, and the resources available on your system. To ensure that the liveness probes in Kubernetes do not start checking the health of your application before the socket is accessible, you can configure the `initialDelaySeconds` parameter in the liveness probe configuration. This will give your Tomcat Java application enough time to start and initialize the socket before Kubernetes starts performing the health checks.

src/main/charts/jira/templates/statefulset.yaml
nanux: Sorry, when I expanded the diff in the source code review, I didn't see it.

nanux:
Thinking out loud, I wondered whether we should add documentation to https://atlassian.github.io/data-center-helm-charts/userguide/CONFIGURATION/ ? We currently don't document the `readinessProbe` either but we are adding the custom definition, and we have 3 different probes now (liveness, readiness, startup). I am leaning towards not documenting it explicitly in our documentation but keeping it references in the `values.yaml` file and referring to official Kubernetes documentation. I think this is quite an advanced topic that the admin should have some idea of how to use generally before changing it in our helm charts.

bianchi2:
@nanux so, what I have changed/added is:

* default failureTreshold for a liveness probe is 12, with a 5 sec periodSeconds, it's 60 seconds. Should be good enough to make the probe less sensitive
* added initialDelaySeconds (should start after the startupProbe + delay)
* links to k8s docs for all 3 types of probes in values.yaml comments

And yes, I agree that this is an advanced operational part of the container spec, and those who'll fine tune it or provide a custom check should know what they are doing :) It's good to have liveness probe and be able to turn it off too. Also, there are cases when a readinessProbe needs to be turned off (for example, I may still want to keep Jira node in the list of endpoints when the node performs re-index which will set the node status to maintenance).

yzha645 (approved):

vernonlingleysonos:
I know I am super late to this, but the `initialDelaySeconds` field is redundant if you are going to be using a `startupProbe` since that probe stops any liveness or readiness probes from running until you succeed.

From https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes
> initialDelaySeconds: Number of seconds after the container has started before startup, liveness or readiness probes are initiated. If a startup probe is defined, liveness and readiness probe delays do not begin until the startup probe has succeeded. Defaults to 0 seconds. Minimum value is 0.

bianchi2:
@vernonlingleysonos it's never too late ;) 

Yes, both readiness and liveness probes kick in only after startupProbe succeeds, however, they both respect `initialDelaySeconds` after `startupProbe` is successful, so it's startupProbe time + `initialDelaySeconds` before readiness/livenes probes start doing their job.

In our case, `startupProbe` just checks if the server listens on the socket (which does not check if the app is ready), and you may give Jira/Confluence more time to be ready (/status reports 200) and not make kubelet do those http get calls for no reason (say, if you know that it can take longer).

PR #630, opened 2023-07-25
Disable liveness probes by default, make timeoutSeconds configurable

bianchi2:
In certain situations livenessProbes can be disruptive. This PR makes them off by default. Also, adding timeoutSeconds for both readiness and liveness probes.

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/bamboo/values.yaml
+++ b/src/main/charts/bamboo/values.yaml
@@ -608,6 +608,10 @@ bamboo:
     #
     periodSeconds: 10
 
+    # -- Number of seconds after which the probe times out
+    #
+    timeoutSeconds: 1
+
     # -- The number of consecutive failures of the Bamboo container readiness probe
     # before the pod fails readiness checks.
     #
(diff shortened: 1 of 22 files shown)


src/main/charts/bamboo/values.yaml
```diff
@@ -608,6 +608,10 @@ bamboo:
     #
     periodSeconds: 10
 
+    # -- Number of seconds after which the probe times out
+    #
+    timeoutSeconds: 1
```
errcode1202, line 613: So if the liveness probe does not get a response after `1` second, thats its cue to trigger a failure and reboot the pod?

errcode1202 (approved):

src/main/charts/bamboo/values.yaml
bianchi2: It will consider this attempt as failed and then try again after `periodSeconds` and until either `successTreshold` or `failureTreshold` is reached.

PR #635, opened 2023-08-05
Add additional ConfigMaps to Helm Charts

bianchi2:
While the helm charts have additionalFiles which make it possible to mount external ConfigMaps to container, there are a couple of issues with this approach:

* ConfigMaps need to be maintained outside Helm charts
* When a ConfigMap data changes the StatefulSet is not restarted, which requires additional automation to make it happen

This PR adds additinalConfigMaps stanza which makes it possible to define multiple ConfigMaps with multiple keys (files) along with the content. One obvious use case is to mount log4j.properties to change logging settings. There can be other config files too.

## Checklist
- [x] I have added unit tests


errcode1202 (commented):
🥇

errcode1202 (approved):

PR #636, opened 2023-08-05
Add PodDisruptionBudget to Atlassian DC Helm Charts

bianchi2:
[PodDisruptionBudget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) is a critical resource when it comes to high availability and (un)planned disruptions like worker node upgrades, evictions etc.

This PR adds a pdb template and a pbd stanza in values.yaml. Note that when both `minAvailable` and `maxUnavailable` are set, `maxUnavailable` takes precedence. By default PDB is disabled.

Just in case, custom pdb labels and annotations can be defined in values. Covered with unit tests.

## Checklist
- [x] I have added unit tests


src/main/charts/bamboo/README.md
```diff
@@ -139,6 +139,7 @@ Kubernetes: `>=1.21.x-0`
 | monitoring.serviceMonitor.prometheusLabelSelector | object | `{}` | ServiceMonitorSelector of the prometheus instance.  |
 | monitoring.serviceMonitor.scrapeIntervalSeconds | int | `30` | Scrape interval for the JMX service.  |
 | nodeSelector | object | `{}` | Standard K8s node-selectors that will be applied to all Bamboo pods  |
+| pdb | object | `{"annotations":{},"enabled":false,"labels":{},"maxUnavailable":null,"minAvailable":null}` | PodDisruptionBudget: https://kubernetes.io/docs/tasks/run-application/configure-pdb/ You can specify only one of maxUnavailable and minAvailable in a single PodDisruptionBudget  |
```
nanux, line 142: I think we can also include `Note that when both minAvailable and maxUnavailable are set, maxUnavailable takes precedence.`

nanux:
Can you describe a scenario in which pod disruption budget is used with our products?

src/main/charts/bamboo/README.md
bianchi2: Yes, makes sense

bianchi2:
@nanux thanks for your comments! In more than 50% of helm charts that I have recently installed, pdb is pdb :) Though, there are ones where it's PodDisruptionBudget.  I have changed the name.

As to how pdb can be used with Atlassian DC products, it's not different from any deployment or statefulset. PDB is a guard rail to prevent a node from draining if pdb says there must be, say, at least 2 pods at all times. DPB is ofted used in combo with anti affinity rules to make sure your app always has some quorum. In case of DC products, you may want to make sure that there are at least 3 running and ready Jira pods at all times.

I can't explain it better than [Kube docs do](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) 😃

errcode1202 (approved):

PR #647, opened 2023-08-16
Add defaultMode to additionalConfigMaps

bianchi2:
We may want to make mounted files executable, so control over defaultMode will help. Other uses cases - security requirements that may enforce certain default modes.

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/bamboo/values.yaml
+++ b/src/main/charts/bamboo/values.yaml
@@ -1063,6 +1063,7 @@ additionalConfigMaps: []
 #    keys:
 #      - fileName: hello.properties
 #        mountPath: /opt/atlassian/jira/atlassian-jira/WEB-INF/classes
+#        defaultMode:
 #        content: |
 #          foo=bar
 #          hello=world
--- a/src/main/charts/bamboo/templates/statefulset.yaml
+++ b/src/main/charts/bamboo/templates/statefulset.yaml
@@ -261,6 +261,9 @@ spec:
         - name: {{ .fileName | replace "_" "-" | replace "." "-" }}
           configMap:
             name: {{ include "common.names.fullname" $ }}-{{ $key.name }}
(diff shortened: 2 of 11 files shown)


src/main/charts/bamboo/values.yaml
```diff
@@ -1063,6 +1063,7 @@ additionalConfigMaps: []
 #    keys:
 #      - fileName: hello.properties
 #        mountPath: /opt/atlassian/jira/atlassian-jira/WEB-INF/classes
+#        defaultMode:
```
yzha645, line 1066: I find it a bit confusing to put `defaultMode` under each key. 
`mode` in my opinion might be a better addition, so that we'll have `spec.template.spec.volumes[n].configMap.items.mode` in the statefulset. 
Maybe it makes more sense to put `defaultMode` at configmap level? i.e. 
```
  - name: extra-configmap
    defaultMode:
    keys:
      - fileName: hello.properties
        mountPath: /opt/atlassian/jira/atlassian-jira/WEB-INF/classes
        mode: 
        content: |
          foo=bar
          hello=world
``` 

But I have limited knowledge on how you designed `additionalConfigMaps` at the beginning, happy to have a chat about it.

bianchi2: @yzha645 thanks for your comment. It's possible to define it one level up, however the design allows multiple keys, so multiple volumes will be created, and one may need to make one of the files executable while others remain without defaultMode (kube will add it automatically). If defaultMode is defined on the configmap level, it will be applied to all keys.

yzha645: yea, thanks for the clarification. I guess in our case, each configmapvolumesource will have one and only one item, so whether set `defaultMode` in upper level or `mode` in item level probably have the same effect.

yzha645 (approved):

Issue #644, opened 2023-08-15
[Suggestion] - Allow Confluence Synchrony small CPU requests


Roia5:
### Suggestion

My synchrony instances (3 pods) in my test environment are hardly working yet they still had a request of 1 cpu core.

When I try to set the cpu request to less than 1 core (for example 500m), The pods will not start since the container crashes immediately.
It seems like the 'entrypoint' configmap uses 'ActiveProcessorCount' with a floored value of my cpu request, and the default is 1 core.
I am unsure if this more of a Java issue with ActiveProcessorCount so please let me know if that is the case.
Thank you!

### Product

Confluence

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bianchi2:
hi @Roia5 it does look like a Java issue. Do you have logs from the failed synchrony pod? In the meantime I'll experiment with cpu requests in my lab.

bianchi2:
Released in 1.15.2

PR #650, opened 2023-08-21
Allow small cpu requests for Synchrony container

bianchi2:
Addresses https://github.com/atlassian/data-center-helm-charts/issues/644

This PR makes it possible to set smaller cpu requests for Synchrony container. Since we use `synchrony.resources.container.requests.cpu` in `-XX:ActiveProcessorCount` in synchrony entrypoint script as a JVM flag, setting anything less than 1 is n't possible since `ActiveProcessorCount` accepts int only. See: https://eclipse.dev/openj9/docs/xxactiveprocessorcount/. As a result, setting `synchrony.resources.container.requests.cpu=100m` or `synchrony.resources.container.requests.cpu=0.5` results in Synchrony container crashing.

This PR adds a simple check - if `synchrony.resources.container.requests.cpu` contains `m` character, `-XX:ActiveProcessorCount` is set to 1.

I have also added a few comments to values.yaml docs, with a link to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu

2 simple unit tests check if `-XX:ActiveProcessorCount=1` when `synchrony.resources.container.requests.cpu=20m`, and 
if if `-XX:ActiveProcessorCount=5` when `synchrony.resources.container.requests.cpu=5`

Also, Synchrony is enabled in KinD tests, which wasn't possible because it wasn't possible to have a small cpu claim (GitHub actions have limits)

## Checklist
- [x] I have added unit tests


src/test/config/kind/common-values.yaml
```diff
@@ -53,3 +53,11 @@ monitoring:
     createDashboards: true
     dashboardLabels:
       grafana_dashboard: dc_monitoring
+
+synchrony:
+  enabled: true
+  resources:
+    container:
+      requests:
+        cpu: 20m
```
bianchi2, line 62: This will alow schedule synchrony pod, and ActiveProcessorCount will be set to 1.

yzha645:
🚀

yzha645 (approved):

PR #653, opened 2023-08-24
Disable startup probes by default

bianchi2:
Startup probes can be tricky especially when it comes to migrations and cold starts (and especially for less experienced k8s admins). In some cases Jira may take > 10 mins to even start listening on 8080. If startupProbe times out the container is restarted. If it does not happen gracefully, this results in orphaned database locks which will prevent the node from starting.

This PR:
* disables startup probes (which make sense only in combination with a `readinessProbe` anyway)
* adds  configurable `initialDelaySeconds` to `startupProbe`
* adds some documentation about the probes, with warnings and links to K8s docs.

## Checklist
- [x] I have added unit tests


errcode1202 (approved):
Updated the wording slightly, but otherwise looks good.

PR #661, opened 2023-09-04
SCALE-104: Make container name configurable

bianchi2:
Fixes https://jira.atlassian.com/browse/SCALE-104

From ops perspective it is sometimes difficult to tell which containers belong to which environments. This PR makes it possible to use a unique (in namespace scope) helm release name as main container name. To keep current names unchanged a new boolean variable is introduced - `<product>.useChartNameAsContainerName` with false default value.

Crowd and confluence had hardcoded values.

## Checklist
- [ ] I have added unit tests


LaszloCsorbaBearingPoint:
Just one observation, the default value would mean not to use the containername name as chart name, which would be the opposite of the current default.
Downside of this that all the users who currently depend on this value, will see a change request when updating to the new chart, instead of leaving the default as is, being true.
Suggestion, keep the value true by default, and false could override to use the fullname instead.

LaszloCsorbaBearingPoint:
Will this option be implemented to Confluence as well?

src/main/charts/bamboo/values.yaml
```diff
@@ -380,6 +380,10 @@ ingress:
 #
 bamboo:
 
+  # -- Whether the main container should acquire helm release name. The default, the container name is bamboo (Helm chart name)
```
errcode1202, line 383:
```suggestion
  # -- Whether the main container should acquire Helm release name. By default the container name is `bamboo` which corresponds to the name of the Helm Chart.
```

errcode1202 (approved):

PR #662, opened 2023-09-06
Allow templating of pod annotations

pathob:
## Pull request description

This PR changes the way pod annotations are templated and allows users to use templating in pod annotations.

## Checklist
- [ ] I have added unit tests


pathob:
I guess adding the `e2e` label is something that you guys need to do?

amhibyan:
@pathob 

Does this feature change how to reference podAnnotations in the values.Yaml?

Previously we had the following:
```    
podAnnotations:
      reloader.stakater.com/auto: "true" 
```
      
But after updating to the latest helm chart with these changes we were getting the following error:
`cannot unmarshal bool into Go struct field ObjectMeta.spec.template.metadata.annotations of type string `

Then we tried commenting that line out and the new helm chart worked fine. Are we missing something here?

pathob:
Hi @amhibyan 

sorry, I didn't mean to break your deployments.

I guess adding a `quote` to my change would fix it for such cases:

```
{{- range $key, $value := .Values.podAnnotations }}
{{ $key }}: {{ tpl $value $ | quote }}
{{- end }}
```

Or as @bianchi2 mentioned, you would need to pass a string by wrapping your annotation value with quotes.

@Atlassian: Let me know if I should provide another PR or whether we can expect users to provide their pod annotations with the correct value type.

amhibyan:
I noticed that the Synchrony podAnnotations works fine with our previous method, checked the changes above and noticed that one wasn't changed. Here was ours for Synchrony.
```       
    synchrony:
      enabled: true
      podAnnotations:
        reloader.stakater.com/auto: "true" 
```

It looks like the [Keycloak](https://github.com/codecentric/helm-charts/blob/9ee33665257a23c32da72f54a3d2bebc05ec348c/charts/keycloakx/templates/statefulset.yaml#L34) example they have `| quote` added on.

amhibyan:
Here is the error I get when I remove the quotes from the reloader:
``` 
Helm upgrade failed: template: confluence/templates/tests/test-shared-home-permissions.yaml:8:8: executing "confluence/templates/tests/test-shared-home-permissions.yaml"  │
│ at <include "confluence.podAnnotations" .>: error calling include: template: confluence/templates/_helpers.tpl:280:19: executing "confluence.podAnnotations" at <$value>: wrong type for value; expec │
│ ted string; got bool

amhibyan:
Sorry for the bad question, but what is the best to go about testing something like this before the change is pushed out publicly? Is there a way to reference a specific change/pr? Thanks in advance for getting this pushed so quickly.

PR #663, opened 2023-09-08
CLIP-1826: Add user provided certificates to the default Java truststore

bianchi2:
While it is possible to create a custom truststore and mount it as a secret using `additionalFiles` and add `additionalJvmArguments`, it takes a lot of steps. This PR makes adding custom certs a lot easier.

```
<product>:
  additionalCertificates:
     secretName
```

If secretName is defined (it must be the secret that has been pre-created outside helm), then:

* an additional init container is added
* this container shares a runtime volume with the main container at `/var/ssl`
* the secret is mounted at `/tmp/crt` in the init container
* keytool command imports all files from `/tmp/crt` into the default java truststore at `$JAVA_HOME/lib/security/cacerts`
* `$JAVA_HOME/lib/security/cacerts` is copied to `/var/ssl/cacerts` which is available to the main container
* `-Djavax.net.ssl.trustStore=/var/ssl/cacerts` sysprop is added to JVM args to instruct the server to use a custom truststore

The secret can contain multiple certificates, for example:

```
data:
  stg.crt: base64encodedcrt
  dev.crt: base64encodedcrt
```
Certs will be mounted as ` /tmp/crt/stg.crt`, ` /tmp/crt/dev.srt`. When certs are imported, filenames are used as aliases.

In addition to unit tests, this feature is enabled in KinD tests to make sure that the init containers don't fail and main containers can find the custom trust store. Here's [KinD run](https://github.com/atlassian/data-center-helm-charts/actions/runs/6118262106).

## Checklist
- [x] I have added unit tests

### DIFF
--- a/docs/docs/userguide/CONFIGURATION.md
+++ b/docs/docs/userguide/CONFIGURATION.md
@@ -519,3 +519,53 @@ readinessProbe:
 !!!warning "`startupProbe` and `livenessProbe`"
 
     Both `startupProbe` and `livenessProbe` are disabled by default. Make sure you go through the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/){.external} before enabling such probes. Misconfiguration can result in unwanted container restarts and failed "cold" starts.
+
+## :material-certificate: Self Signed Certificates
+
+Accessing applications or websites that are encrypted with SSL using certificates not signed by a public authority will result in a connection error. The stacktrace will contain the following:
+
+```shell
+caused by: javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed:sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
+```
+
+To fix the issue, self signed certificates need to be added to Java truststore. First off, you need to create a [Kubernetes secret](https://kubernetes.io/docs/concepts/configuration/secret/){.external} containing base64-encoded certificate(s). Here's an example [kubectl command](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#use-source-files){.external} to create a secret from 2 local files:
+
+```shell
+kubectl create secret generic dev-certificates \
+    --from-file=stg.crt=./stg.crt \
+    --from-file=dev.crt=./dev.crt -n $namespace
+```
+
+The resulting secret will have the following data:
+
+```yaml
+data:
+  stg.crt: base64encodedstgcrt
+  dev.crt: base64encodeddevcrt
+```
+
+!!!info "You can have as many keys (certificates) in the secret as required. All keys will be mounted as files to `/tmp/crt` in the container and imported into Java truststore. In the example above, certificates will be mounted as `/tmp/crt/stg.crt` and `/tmp/crt/dev.crt`. File extension in the secret keys does not matter as long as the file is a valid certificate."
+
+Once done, provide the secret name in Helm values: 
+
+```yaml
+jira:
+  additionalCertificates:
+     secretName: dev-certificates
+```
+
+Helm chart will add additional `volumeMounts` and `volumes` to the pod(s), as well as an extra init container that will:
+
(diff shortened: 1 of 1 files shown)


src/main/charts/bamboo/templates/_helpers.tpl
```diff
@@ -307,3 +318,11 @@ Define additional hosts here to allow template overrides when used as a sub char
 {{- toYaml . }}
 {{- end }}
 {{- end }}
+
+{{- define "bamboo.addCrtToKeystoreCmd" }}
+{{- if .Values.bamboo.additionalCertificates.customCmd}}
+{{ .Values.bamboo.additionalCertificates.customCmd}}
+{{- else }}
+set -e; for crt in /tmp/crt/*.*; do echo "Adding $crt to keystore"; keytool -import -cacerts -storepass changeit -noprompt -alias $(echo $(basename $crt)) -file $crt; done; cp $JAVA_HOME/lib/security/cacerts /var/ssl/cacerts
```
nanux, line 326: What happens when customer has or wants to define different passkey than the default `changeit`?

src/main/charts/bamboo/templates/statefulset.yaml
```diff
@@ -55,6 +55,18 @@ spec:
           command: ["sh", "-c", {{ include "bamboo.sharedHome.permissionFix.command" . | quote }}]
         {{- end }}
         {{- include "common.jmx.initContainer" . | nindent 8 }}
+        {{- if .Values.bamboo.additionalCertificates.secretName }}
+        - name: import-certs
+          image: {{ include "bamboo.image" . | quote }}
```
nanux, line 60: I guess we are running this image because we want to run the `keytool` util? I am just wondering whether it is necessary to spin up the large product image to run a simple command.

nanux (commented):
Great improvement @bianchi2! It is great to streamline the experience of setting up the certificates for all users.

Do you think it would be worth adding a paragraph or two to the user facing documentation?
Something along the lines:

**How to use custom certificates**

src/main/charts/bamboo/templates/_helpers.tpl
nanux: Right and if customer wants to do something specific in this command, they can always redefine the `customCmd`.

src/main/charts/bamboo/templates/statefulset.yaml
nanux: Thanks for confirming this. I was thinking that because we pull the image anyway, it is faster on the pull time but for a very short period we take more resources than we need but as the execution is super quick, it is negligible.

Last thing, if this fails, does it stop container execution, and how will customers get to the error logs (are they sufficient to understand why it failed?)

nanux (approved):

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -519,3 +519,53 @@ readinessProbe:
...
+`-Djavax.net.ssl.trustStore=/var/ssl/cacerts` system property will be automatically added to `JVM_SUPPORT_RECOMMENDED_ARGS` environment variable.
+
+If necessary, it is possible to override the default `keytool -import` command:
+
+```yaml
+jira:
+  additionalCertificates:
+     secretName: dev-certificates
+     customCmd: keytool -import ...
+```
```
errcode1202, line 571: @bianchi2 Im just wondering if it makes more sense for this to go under the `TROUBLESHOOTING` section? Maybe not, but as a compromise a section in the troubleshooter page talking to `PKIX` exceptions with a link to the fix might be good?

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -519,3 +519,53 @@ readinessProbe:
...
+
+Once done, provide the secret name in Helm values: 
+
+```yaml
+jira:
+  additionalCertificates:
+     secretName: dev-certificates
+```
+
+Helm chart will add additional `volumeMounts` and `volumes` to the pod(s), as well as an extra init container that will:
```
errcode1202, line 557:
```suggestion
The product Helm chart will add additional `volumeMounts` and `volumes` to the pod(s), as well as an extra init container that will:
```

Issue #666, opened 2023-09-14
[Suggestion] - Make securityContext in fetch-jmx-exporter configurable


ChampiYann:
### Suggestion

When running on Openshift and using the fetch-jmx-exporter initContainer, Openshift won't allow the pod to run if the correct security context constraint is set to allow the fetch-jmx-exporter initContainer to run as user 0.
After removing the securityContext from the initContainer configuration, the container runs fine and metrics are still generated.
I'm not sure why the initContainer should run as user 0 but if this is needed for some usecases, could the securityContext at least be optional?

### Product

Jira, Confluence, Bitbucket

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #667, opened 2023-09-14
Make it possible to run jmx init container as non root in common chart

bianchi2:
Addresses https://github.com/atlassian/data-center-helm-charts/issues/666

When `runAsRoot` is forbidden in the cluster, the pod can't be scheduled. It **does** make sense to be able to disable securityContent for the init container, as well as let users define their own securityContext (can't think of cases when container securityContext will differ from the one set in pod spec though).

This PR brings changes to the common chart which defines jmx init container. Once merged, a new version will be released, and a new PR to helm charts will be raised with the 2nd portion of changes.

## Checklist
- [ ] I have added unit tests

### DIFF
--- a/src/main/charts/bitbucket/Chart.yaml
+++ b/src/main/charts/bitbucket/Chart.yaml
@@ -3,7 +3,7 @@ name: bitbucket
 description: A chart for installing Bitbucket Data Center on Kubernetes
 type: application
 version: 1.15.3
-appVersion: 8.9.4
+appVersion: 8.9.5
 kubeVersion: ">=1.21.x-0"
 keywords:
 - Bitbucket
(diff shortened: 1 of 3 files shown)


src/main/charts/common/templates/_jmx.tpl
```diff
@@ -30,8 +30,15 @@ Jmx init container
   image: {{ .Values.monitoring.jmxExporterImageRepo}}:{{ .Values.monitoring.jmxExporterImageTag}}
   command: ["cp"]
   args: ["/opt/bitnami/jmx-exporter/jmx_prometheus_javaagent.jar", "{{ .Values.volumes.sharedHome.mountPath }}"]
+  {{- if .Values.monitoring.jmxExporterInitContainer.runAsRoot }}
```
nanux, line 33: Maybe I am just blind but I can't see the `monitoring.jmxExporterInitContainer.runAsRoot` and `monitoring.jmxExporterInitContainer.customSecurityContext` stanza in any of the `values.yaml` files. Was this just omitted in the PR?

nanux: I see, that does make sense. Might be just good to indicate in PR description.

nanux (approved):

PR #670, opened 2023-09-18
Make securityContext configurable for jmx exporter init container

bianchi2:
https://github.com/atlassian/data-center-helm-charts/pull/667 is merged. This PR adds values and a few unit tests.

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/bamboo/values.yaml
+++ b/src/main/charts/bamboo/values.yaml
@@ -847,8 +847,8 @@ monitoring:
   #
   jmxExporterInitContainer:
 
-    # -- Whether to run jmx exporter init container as root to be able to copy jmx exporter binary to shared home volume.
-    # Set to false if running containers as root is not allowed in the cluster
+    # -- Whether to run JMX exporter init container as root to copy JMX exporter binary to shared home volume.
+    # Set to false if running containers as root is not allowed in the cluster.
     #
     runAsRoot: true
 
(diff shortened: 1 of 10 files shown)


src/main/charts/bamboo/values.yaml
```diff
@@ -843,6 +843,19 @@ monitoring:
   #
   exposeJmxMetrics: false
 
+  # --  JMX exporter init container configuration
+  #
+  jmxExporterInitContainer:
+
+    # -- Whether to run jmx exporter init container as root to be able to copy jmx exporter binary to shared home volume.
+    # Set to false if running containers as root is not allowed in the cluster
```
nanux, line 851:
```suggestion
    # Set to false if running containers as root is not allowed in the cluster.
```

src/main/charts/bamboo/values.yaml
```diff
@@ -843,6 +843,19 @@ monitoring:
   #
   exposeJmxMetrics: false
 
+  # --  JMX exporter init container configuration
+  #
+  jmxExporterInitContainer:
+
+    # -- Whether to run jmx exporter init container as root to be able to copy jmx exporter binary to shared home volume.
```
nanux, line 850:
```suggestion
    # -- Whether to run JMX exporter init container as root to copy JMX exporter binary to shared home volume.
```

src/main/charts/bamboo/values.yaml
```diff
@@ -843,6 +843,19 @@ monitoring:
...
+    # Set to false if running containers as root is not allowed in the cluster
+    #
+    runAsRoot: true
+
+    # -- Custom SecurityContext for the jmx exporter init container
```
nanux, line 855:
```suggestion
    # -- Custom SecurityContext for the JMX exporter init container
```

nanux (approved):

PR #675, opened 2023-09-20
Make it possible to use one Bitbucket svc for http and ssh without an ingress

bianchi2:
This PR fixes a few bugs in Bitbucket helm chart. Currently, it is [possible to change](https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bitbucket/templates/statefulset.yaml#L92) `bitbucket.ports.ssh` however, changing it will break networking because container port is changed but the port on which Bitbucket server ssh service actually listens on isn't changed. Adding `PLUGIN_SSH_PORT ` fixes it. Note that it can be any port from the unprivileged range (1024-65535). 

The second issue is with the service. Currently, service port for ssh is [bound to a container port](https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bitbucket/templates/service.yaml#L28). However, in a cloud environment is makes sense to set it to 22, so that the cloud provider creates LB with a 22 listener. This PR adds `bitbucket.service.sshPort` with the default value of `7999`, so nothing changes for existing users. Also, it's just a good practice to decouple svc and container ports, especially if we have control over ports the Bitbucket server listens on.

Adding some docs and unit tests too. e2e tests are 🟢 (we do git clone using ssh protocol there)

## Checklist
- [x] I have added unit tests

### DIFF
--- a/docs/docs/examples/bitbucket/BITBUCKET_SSH.md
+++ b/docs/docs/examples/bitbucket/BITBUCKET_SSH.md
@@ -131,4 +131,8 @@ ingress:
 
 The above service annotations are specific to the Classic LoadBalancer, however, you can provide [NLB](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html){.external} specific [annotations](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/){.external} as well.
 
-The default `bitbucket.service.sshPort` is set to `22` so that AWS can create a listener for this port, and as a result your ssh git URL will look like `ssh://bitbucket.example.com/project/repo`. Even though `ingress` is disabled, `ingress.host` needs to be set because it is used in a few conditions in the StatefulSet template.
+The default `bitbucket.service.sshPort` is set to `22` so that AWS can create a listener for this port, and as a result your ssh git URL will look like `ssh://bitbucket.example.com/project/repo`. 
+
+!!!info "Ingress host"
+
+    Even though `ingress` is disabled, `ingress.host` needs to be set because it is used in a few conditions in the StatefulSet template.


docs/docs/examples/bitbucket/BITBUCKET_SSH.md
```diff
@@ -107,3 +108,27 @@ bitbucket:
...
+  additionalEnvironmentVariables:
+    - name: PLUGIN_SSH_BASEURL
+      value: ssh://bitbucket.example.com
+ingress:
+  host: bitbucket.example.com
+```
+
+The above service annotations are specific to the Classic LoadBalancer, however, you can provide [NLB](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html){.external} specific [annotations](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/annotations/){.external} as well.
+
+The default `bitbucket.service.sshPort` is set to `22` so that AWS can create a listener for this port, and as a result your ssh git URL will look like `ssh://bitbucket.example.com/project/repo`. Even though `ingress` is disabled, `ingress.host` needs to be set because it is used in a few conditions in the StatefulSet template.
```
errcode1202, line 134:
```suggestion
The default `bitbucket.service.sshPort` is set to `22` so that AWS can create a listener for this port, and as a result your ssh git URL will look like `ssh://bitbucket.example.com/project/repo`. 

!!!info "Ingress host"

    Even though `ingress` is disabled, `ingress.host` needs to be set because it is used in a few conditions in the StatefulSet template.
```

src/main/charts/bitbucket/values.yaml
```diff
@@ -471,10 +471,14 @@ bitbucket:
   #
   service:
 
-    # -- The port on which the Bitbucket K8s Service will listen
+    # -- The port on which the Bitbucket K8s HTTP Service will listen
     #
     port: 80
 
+    # -- The port on which the Bitbucket K8s SSH Service will listen
```
errcode1202, line 478: This split beteween HTTP and SSH already makes things clearer.

src/main/charts/bitbucket/templates/statefulset.yaml
```diff
@@ -178,6 +178,8 @@ spec:
             {{- include "bitbucket.databaseEnvVars" . | nindent 12 }}
             {{- include "bitbucket.sysadminEnvVars" . | nindent 12 }}
             {{- include "bitbucket.elasticSearchEnvVars" . | nindent 12 }}
+            - name: PLUGIN_SSH_PORT
```
errcode1202, line 181: I see `plugin.ssh.port` is mentioned here: I see `plugin.ssh.port` is a property that can be supplied: https://community.atlassian.com/t5/Bitbucket-questions/Custom-ssh-port-for-bitbucket-smart-mirror/qaq-p/2131340.  Was just wondering if there was any official documentation regarding its use?

errcode1202 (approved):

PR #686, opened 2023-10-08
Make it possible to create a dedicated hazelcast svc

bianchi2:
In certain scenarios having hazelcast port in the main service may cause issues. If the main service is of LoadBalancer type, a cloud provider (for example AWS) will create a LoadBalancer (for example NLB), with listeners per each port. There are 2 issues with it:

* hazelcast port is exposed (which is not necessary since k8s hazelcast discovery happens on Kubernetes svc level)
* TCP health check for the hazelcast target group makes Bitbucket unhappy:

<details>
            <summary>Unexpected bytes from remote node, closing socket</summary>
          
            ```
              2023-10-05 19:53:24,442 WARN  [hz.hazelcast.cached.thread-4]  c.a.s.i.c.DefaultClusterJoinManager ACCEPT(10.0.0.109:5701 <- 10.0.2.3:55593): Node authentication failed: Unexpected bytes from remote node, closing socket
              2023-10-05 19:53:24,442 WARN  [hz.hazelcast.cached.thread-4]  com.hazelcast.nio.tcp.TcpIpAcceptor [10.0.0.109]:5701 [bitbucket] [3.12.13] com.atlassian.stash.internal.cluster.NodeConnectionException: Unexpected bytes from remote node, closing socket
              com.atlassian.stash.internal.cluster.NodeConnectionException: Unexpected bytes from remote node, closing socket
	              at com.atlassian.stash.internal.cluster.DefaultClusterJoinManager.accept(DefaultClusterJoinManager.java:103)
	              at com.atlassian.stash.internal.hazelcast.ClusterJoinSocketInterceptor.onAccept(ClusterJoinSocketInterceptor.java:49)
	              at com.hazelcast.nio.NodeIOService.interceptSocket(NodeIOService.java:300)
	              at com.hazelcast.nio.tcp.TcpIpAcceptor$AcceptorIOThread.configureAndAssignSocket(TcpIpAcceptor.java:316)
	              at com.hazelcast.nio.tcp.TcpIpAcceptor$AcceptorIOThread.access$1400(TcpIpAcceptor.java:138)
	              at com.hazelcast.nio.tcp.TcpIpAcceptor$AcceptorIOThread$1.run(TcpIpAcceptor.java:305)
	              at com.hazelcast.util.executor.CachedExecutorServiceDelegate$Worker.run(CachedExecutorServiceDelegate.java:227)
	              at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #689, opened 2023-10-11
Fix hazelcast svc type

bianchi2:
Fixup to https://github.com/atlassian/data-center-helm-charts/pull/686

A dedicated hazelcast svc was using bitbucket service type which defeats the purpose of having a dedicated hazelcast svc :) 

Added this check to unit tests

## Checklist
- [ ] I have added unit tests


nanux (approved):
Good catch!
How did we discover this?

PR #695, opened 2023-10-20
Fix jmx init container for Bitbucket mirror

bianchi2:
The init container to fetch jmx exporter jar relies on a shared-home volume. Bitbucket Mirror does not have one. This PR extends conditions under which shared-home volumeMount and volume are defined and sets `emptyDir: {}` for the volume. 

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/main/charts/bitbucket/templates/statefulset.yaml
+++ b/src/main/charts/bitbucket/templates/statefulset.yaml
@@ -136,6 +136,7 @@ spec:
           volumeMounts:
             - name: local-home
               mountPath: {{ .Values.volumes.localHome.mountPath | quote }}
+            {{/* jmxExporter is saved to shared home but mirrors don't have one, so we need at least emptyDir */}}
             {{- if or .Values.volumes.sharedHome.persistentVolumeClaim.create .Values.volumes.sharedHome.customVolume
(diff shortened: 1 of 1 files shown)


nanux:
Good catch, was this missed because we never tested observability with Bitbucket mirror? Can we have mirrors as part of our system tests in Github Actions?

nanux (approved):

src/main/charts/bitbucket/templates/statefulset.yaml
```diff
@@ -136,7 +136,8 @@ spec:
           volumeMounts:
             - name: local-home
               mountPath: {{ .Values.volumes.localHome.mountPath | quote }}
-            {{- if or .Values.volumes.sharedHome.persistentVolumeClaim.create .Values.volumes.sharedHome.customVolume }}
+            {{- if or .Values.volumes.sharedHome.persistentVolumeClaim.create .Values.volumes.sharedHome.customVolume
```
nanux, line 139: I think it will be useful to have a comment here explaining why we have this condition. Something like

{{/* jmxExporter is saved to shared home but mirrors don't have one, so we need at least emptyDir */}}

nanux (approved):

errcode1202 (approved):

Issue #680, opened 2023-10-05
[Suggestion] - Add resource requests and limits for init-containers


Pit-Storm:
### Suggestion

It would be helpful to have default resource requests and limits set for the init-containers as well. If a namespace has resource quotas in place, the deployment will fail. This is due to the fact that every containers need to have resource requests and limits need to be defined when resource quotas are specified.

This patch would be quite easy to make for just having default settings.

A second step would be to add the possibility for adjusting them as well in the values.yaml.

### Product

Jira, Confluence

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #704, opened 2023-10-27
Add optional resources requests and limits for init and additional containers

bianchi2:
Fixes https://github.com/atlassian/data-center-helm-charts/issues/680

While resources requests and caps aren't really necessary for some of the init containers, there can be admission controller rules that require all containers to have resources requests and/or limits. Typically, admission controllers mutate containers and add defaults but if it does not happen the container is denied.

More feedback from users:
> Because fluentd container doesn't have default resource requests/limits, as a result I can't create jira pod in a guaranteed QoS status

This means that a container without resources requests is deemed as burstable, i.e. the pod is the number one candidate to be re-homed if the node is under pressure.

This PR adds optional resources requests/limits to:

* fluentd container if enabled
* nfsPermissionFixer init container if enabled
* jmxExporter init container if enabled (requires release of the common chart and an additional PR to bump dep in DC helm charts)

The default behavior stays the same.

## Checklist
- [x] I have added unit tests

### DIFF
--- a/docs/docs/userguide/resource_management/REQUESTS_AND_LIMITS.md
+++ b/docs/docs/userguide/resource_management/REQUESTS_AND_LIMITS.md
@@ -41,3 +41,39 @@ container:
     cpu: "2" 
     memory: "2G"
 ```
+
+## Init and Additional Containers
+
+By default, NFS permission fixer and JMX init containers, as well as Fluentd additional container don't have resources requests and limits. You can set them in `values.yaml`:
+
+```yaml
+volumes:
+  sharedHome:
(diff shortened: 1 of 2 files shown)


nanux:
Nice one @bianchi2.

Would it be worth mentioning the option to set resources on the init containers in our user documentation? We do have a section for resources and limits - https://atlassian.github.io/data-center-helm-charts/userguide/resource_management/REQUESTS_AND_LIMITS/

nanux (approved):

yzha645 (approved):

PR #719, opened 2023-11-20
BAM-25528: Unset pre-stop hook for bamboo-agent

bianchi2:
The shutdown script isn't part of the agent image, so this pre-stop hook was always failing. In fact, a proper pre stop hook for an agent will be the one that checks if an agent is busy or not. However, there's no such feature.

This PR sets `agent.shutdown.command` to nil but still makes it possible to have a pre-stop hook if the command is defined in user provided values.


nanux (approved):
Good catch, how come we didn't see this failure in the tests?

PR #752, opened 2024-02-01
OpenShift Support

bianchi2:
# What This PR Is All About

This PR adds Red Hat OpenShift support to Atlassian DC Helm charts, and namely makes it possible to run DC containers with a restricted SCC, as well as add OpenShift Route templates.

## Restricted SCC

This PR consolidates existing Helm charts features (introduced by https://github.com/atlassian/data-center-helm-charts/pull/623) that allow running DC container as an unprivileged OpenShift user with [restricted SCC](https://docs.openshift.com/container-platform/4.14/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth) which requires that a pod is run as a user in a pre-allocated range of UIDs. It also extends it to Crowd and Bamboo Helm charts.

All the [OpenShift friendly values](https://atlassian.github.io/data-center-helm-charts/platforms/OPENSHIFT/#openshift-run-as-nonroot-friendly-values) were consolidated into a new boolean `openshift.runWithRestrictedSCC` which when set to true:

* **unsets securityContext** in pods (thus letting OpenShift set it dynamically)
* disables **nfs-permission fixer** init container (that runs as root by default)
* enables creation of **ConfigMaps with server.xml and seraph-config.xml**, i.e. stop relying on generation of these files by the image entrypoint scripts (which will fail dues to insufficient permissions - no users other than root can write to products install directories). This also includes init.properties file for Bamboo
* sets **defaultMode** for **synchrony entrypoint script** ConfigMap to 360 (octal 550 - i.e. allowing group to execute the script)
* unsets **securityContext for jmx-exporter** init container (if enabled). See: [#784](https://github.com/atlassian/data-center-helm-charts/pull/748)

All DC statefulsets (including Mesh and Confluence Synchrony) are covered. Bamboo agent deployment has been fixed too. In simple words, `openshift.runWithRestrictedSCC` makes **DC pods runnable with a restricted SCC** (which is the default SCC attached to a new ServiceAccount).

## OpenShift Routes

Existing Ingress stanza is deeply integrated into existing templates and helpers. Thus, it's been decided to keep it and just introduce a new boolean `ingress.openShiftRoute`. When set to true, Helm is instructed to skip creating an ingress in favor of creating a [Route](https://docs.openshift.com/container-platform/4.14/networking/routes/route-configuration.html).  A few gotchas:

* to create OpenShift route both `ingress.create` and `ingress.openShiftRoute` need to be set to true
* ingressClass can be ignored
* `ingress.host` and `ingress.path` are used in routes templates, so now new props here
* props like maxBodySize are ignored
* Route settings should be configured in `ingress.annotations`. See: [Route-specific annotations](https://docs.openshift.com/container-platform/4.14/networking/routes/route-configuration.html#nw-route-specific-annotations_route-configuration).
* for TLS routes it is possible to use the default router certificate or define `ingress.tlsSecretName` with a tls secret name containing tls.crt, tls,key and ca.crt (optional) to force OpenShift router use a specific crt/key for the Route
* TLS terminated at router (edge type) and it's not configurable, i.e. passthrough and re-encrypt routes aren't support (we can do it in future if there are valid use cases)
* `insecureEdgeTerminationPolicy` is **Redirect** and not configurable (we can make it configurable later if there are valid use cases)
* wildcard policy is None and not configurable
* [spec.httpHeaders](https://docs.openshift.com/container-platform/4.14/rest_api/network_apis/route-route-openshift-io-v1.html#spec-httpheaders) are configurable in `ingress.routeHttpHeaders`, so you can pass your own custom object
* `spec.alternateBackends` are not supported (we will add it if there are valid use cases)

## CI Tests

After experimenting with [CodeReady Containers](https://github.com/crc-org/crc), openshift and okd profiles proved to be rather unstable in GitHub Actions. VMs tend to freeze every now and then because GitHub Actions run in shared environments and there's not enough resources to start an OpenShift cluster (even a 1-node cluster). A [MicroShift](https://github.com/openshift/microshift) profile is a solution.  It is possible to start a very minimalistic OpenShift cluster and run tests that verify containers can start with a restricted SCC, and OpenShift Routes respond with 200. A workflow similar to KinD is part of this PR. A pull-secret is required (to pull RH images), and it's been saved in repo secrets. See example run: https://github.com/atlassian/data-center-helm-charts/pull/748

## Docs

It'll be a separate PR since once merged to main, docs are automatically published. 

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/test/resources/expected_helm_output/bamboo/output.yaml
+++ b/src/test/resources/expected_helm_output/bamboo/output.yaml
@@ -352,7 +352,7 @@ data:
       "replicas": 1,
       "isJmxEnabled": true,
       "ingressType": "NONE",
-      "k8sVersion": "1.24",
+      "k8sVersion": "1.28",
       "serviceType": "CLUSTER_IP",
       "dbType": "UNKNOWN",
       "isSharedHomePVCCreated": false,
--- a/src/test/resources/expected_helm_output/bitbucket/output.yaml
+++ b/src/test/resources/expected_helm_output/bitbucket/output.yaml
@@ -403,7 +403,7 @@ data:
       "replicas": 1,
       "isJmxEnabled": true,
       "ingressType": "NONE",
-      "k8sVersion": "1.24",
+      "k8sVersion": "1.28",
       "serviceType": "CLUSTER_IP",
       "dbType": "UNKNOWN",
       "isClusteringEnabled": false,
--- a/src/test/resources/expected_helm_output/confluence/output.yaml
+++ b/src/test/resources/expected_helm_output/confluence/output.yaml
@@ -396,7 +396,7 @@ data:
       "replicas": 1,
       "isJmxEnabled": true,
       "ingressType": "NONE",
-      "k8sVersion": "1.24",
+      "k8sVersion": "1.28",
       "serviceType": "CLUSTER_IP",
       "dbType": "UNKNOWN",
       "isS3AttachmentsStorageEnabled":false,
--- a/src/test/resources/expected_helm_output/crowd/output.yaml
+++ b/src/test/resources/expected_helm_output/crowd/output.yaml
@@ -299,7 +299,7 @@ data:
       "replicas": 1,
       "isJmxEnabled": true,
       "ingressType": "NONE",
-      "k8sVersion": "1.24",
+      "k8sVersion": "1.28",
       "serviceType": "CLUSTER_IP",
       "dbType": "UNKNOWN",
       "isSharedHomePVCCreated": false,
--- a/src/test/resources/expected_helm_output/jira/output.yaml
+++ b/src/test/resources/expected_helm_output/jira/output.yaml
@@ -318,7 +318,7 @@ data:
       "replicas": 1,
       "isJmxEnabled": true,
       "ingressType": "NONE",
-      "k8sVersion": "1.24",
+      "k8sVersion": "1.28",
       "isS3AvatarsEnabled": false,
       "serviceType": "CLUSTER_IP",
       "dbType": "UNKNOWN",


src/main/charts/confluence/values.yaml
```diff
@@ -465,6 +465,14 @@ ingress:
   #
   create: false
 
+  # -- Set to true if you want to create an OpenShift Route instead of an Ingress
+  #
+  openShiftRoute: false
+
+  # -- RouteHTTPHeaders defines policy for HTTP headers. Applicable to OpenShift Routes only
```
yzha645, line 472:
```suggestion
  # -- routeHttpHeaders defines policy for HTTP headers. Applicable to OpenShift Routes only
```

src/main/charts/crowd/templates/configmap-server-config.yaml
```diff
@@ -0,0 +1,83 @@
...
+  labels:
+    {{- include "common.labels.commonLabels" . | nindent 4 }}
+data:
+{{- if or .Values.crowd.tomcatConfig.generateByHelm .Values.openshift.runWithRestrictedSCC }}
+  server.xml: |
+{{- if .Values.crowd.tomcatConfig.customServerXml }}
+{{- .Values.crowd.tomcatConfig.customServerXml | nindent 6 }}
+{{- else }}
+    <?xml version="1.0" encoding="utf-8"?>
+    <Server port="{{ .Values.crowd.tomcatConfig.mgmtPort | default "8005" }}"
```
yzha645, line 15:
```suggestion
    <Server port="{{ .Values.crowd.tomcatConfig.mgmtPort | default "8020" }}"
```

src/main/charts/crowd/templates/configmap-server-config.yaml
```diff
@@ -0,0 +1,83 @@
...
+      <Listener className="org.apache.catalina.core.AprLifecycleListener"
+                SSLEngine="on"/>
+      <Listener className="org.apache.catalina.core.JreMemoryLeakPreventionListener"/>
+      <Listener className="org.apache.catalina.mbeans.GlobalResourcesLifecycleListener"/>
+      <Listener className="org.apache.catalina.core.ThreadLocalLeakPreventionListener"/>
+
+      <Service name="Catalina">
+
+        <Connector port="{{ .Values.crowd.tomcatConfig.port | default "8095" }}"
+                   maxThreads="{{ .Values.crowd.tomcatConfig.maxThreads | default "100" }}"
```
yzha645, line 28:
```suggestion
                   maxThreads="{{ .Values.crowd.tomcatConfig.maxThreads | default "150" }}"
```

src/main/charts/crowd/templates/configmap-server-config.yaml
```diff
@@ -0,0 +1,83 @@
...
+                   minSpareThreads="{{ .Values.crowd.tomcatConfig.minSpareThreads | default "10" }}"
```
yzha645, line 29:
```suggestion
                   minSpareThreads="{{ .Values.crowd.tomcatConfig.minSpareThreads | default "25" }}"
```

src/main/charts/crowd/templates/configmap-server-config.yaml
```diff
@@ -0,0 +1,83 @@
...
+                   connectionTimeout="{{ .Values.crowd.tomcatConfig.connectionTimeout | default "20000" }}"
+                   enableLookups="{{ .Values.crowd.tomcatConfig.enableLookups | default "false" }}"
+                   protocol="{{ .Values.crowd.tomcatConfig.protocol | default "HTTP/1.1" }}"
+                   redirectPort="{{ .Values.crowd.tomcatConfig.redirectPort | default "8443" }}"
+                   acceptCount="{{ .Values.crowd.tomcatConfig.acceptCount | default "10" }}"
```
yzha645, line 34:
```suggestion
                   acceptCount="{{ .Values.crowd.tomcatConfig.acceptCount | default "100" }}"
```

src/main/charts/crowd/values.yaml
```diff
@@ -348,6 +348,33 @@ crowd:
   #
   additionalJvmArgs: []
 
+  tomcatConfig:
+
+    # -- Mount server.xml as a ConfigMap. Override configuration elements if necessary
+    #
+    generateByHelm: false
+
+    mgmtPort: "8005"
```
yzha645, line 357:
```suggestion
    mgmtPort: "8020"
```

src/main/charts/crowd/values.yaml
```diff
@@ -348,6 +348,33 @@ crowd:
...
+    port: "8095"
+    maxThreads: "100"
```
yzha645, line 359:
```suggestion
    maxThreads: "150"
```

src/main/charts/crowd/values.yaml
```diff
@@ -348,6 +348,33 @@ crowd:
...
+    minSpareThreads: "10"
```
yzha645, line 360:
```suggestion
    minSpareThreads: "25"
```

src/main/charts/crowd/values.yaml
```diff
@@ -348,6 +348,33 @@ crowd:
...
+    connectionTimeout: "20000"
+    enableLookups: "false"
+    protocol: "HTTP/1.1"
+    redirectPort: "8443"
+    acceptCount: "10"
```
yzha645, line 365:
```suggestion
    acceptCount: "100"
```

src/main/charts/crowd/values.yaml
```diff
@@ -433,6 +460,14 @@ ingress:
   #
   create: false
 
+  # -- Set to true if you want to create an OpenShift Route instead of an Ingress
+  #
+  openShiftRoute: false
+
+  # -- RouteHTTPHeaders defines policy for HTTP headers. Applicable to OpenShift Routes only
```
yzha645, line 467:
```suggestion
  # -- routeHttpHeaders defines policy for HTTP headers. Applicable to OpenShift Routes only
```

src/main/charts/jira/values.yaml
```diff
@@ -354,6 +354,14 @@ ingress:
   #
   create: false
 
+  # -- Set to true if you want to create an OpenShift Route instead of an Ingress
+  #
+  openShiftRoute: false
+
+  # -- RouteHTTPHeaders defines policy for HTTP headers. Applicable to OpenShift Routes only
```
yzha645, line 361:
```suggestion
  # -- routeHttpHeaders defines policy for HTTP headers. Applicable to OpenShift Routes only
```

yzha645 (changes requested):

yzha645 (approved):

PR #753, opened 2024-02-05
Fix broken links in the documentation

nanux:
## Pull request description

When testing other PR I found out that we have quite a lot of broken links in the docs. I went through them and fixed them (they were reporting 404).

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -32,8 +32,8 @@ plugins:
         remove_comments: true
 markdown_extensions:
   - pymdownx.emoji:
-      emoji_index: !!python/name:materialx.emoji.twemoji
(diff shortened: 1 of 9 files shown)


docs/mkdocs.yml
```diff
@@ -32,8 +32,8 @@ plugins:
         remove_comments: true
 markdown_extensions:
   - pymdownx.emoji:
-      emoji_index: !!python/name:materialx.emoji.twemoji
-      emoji_generator: !!python/name:materialx.emoji.to_svg
+      emoji_index: !!python/name:material.extensions.emoji.twemoji
```
nanux, line 35: Original package is deprecated and we should be using new one.

PR #771, opened 2024-03-11
BSERV-19054 Re-enable plugin upload for test Helm chart

benhumphreys:
## Pull request description

Modern Bitbucket no longer permits plugin upload via UPM by default. This is however useful for testing and in test and development environments we've re-enabled the functionality. The same should be done for the test helm charts.

For further details on the new Bitbucket defaults see: https://jira.atlassian.com/browse/BSERV-19054

## Checklist
- [N/A] I have added unit tests
- [X] The E2E test has passed (use `e2e` label)

### DIFF
--- a/src/test/config/bitbucket/values.yaml
+++ b/src/test/config/bitbucket/values.yaml
@@ -26,6 +26,7 @@ bitbucket:
     secretName: bitbucket-sysadmin-credentials
   additionalJvmArgs:
     - -Dfeature.getting.started.page=false
+    - -Dupm.plugin.upload.enabled=true
     - -XX:ActiveProcessorCount=2
   resources:
     container:


nanux (approved):
LGTM.

I assume this is not something we want to expose to customers as a configuration parameter.

benhumphreys-atlassian:
> I assume this is not something we want to expose to customers as a configuration parameter.

Note I close this pull request and opened another using my (new) Atlassian Github account so I don't have sign the CLA. New PR is here: https://github.com/atlassian/data-center-helm-charts/pull/772

But to answer your question; this is something that is available to customers by adding the same entry to their `bitbucket.properties` file. A few hundred properties are tuneable via bitbucket.properties; many of them frequently used, but I'm not sure what the criteria is for it to be elevated to a first class Helm configuration parameter.

PR #774, opened 2024-03-20
CLIP-1859: Added docs of DC Docker images

yzha645:
## Pull request description

As the README of Docker images grow, it starts to exceed the size limit imposed by Docker Hub. Also, the format is not as great as mkdocs. 

This PR moves docs of DC Docker images to Helm chart docs as a new 'Container' section. Fixed formatting and links as well. 

Some of the changes including: 
1. Added new 'Container' tab to the upper tab
2. Left side navigation bar is consists of different products 
3. Added title to the doc and made logo smaller
4. Added server deprecation warning on top of the doc
5. Now have table of contents at right hand side, which is much easier to navigate through the whole page. 

Screenshot 2024-03-21 at 11 25 58 AM

## Checklist
- [ ] I have added unit tests


nanux:
@yzha645 can you please add screenshot or some test deployment to see how it is rendered?

yzha645:
> @yzha645 can you please add screenshot or some test deployment to see how it is rendered?

@nanux  Updated PR description.

yzha645:
> @yzha645 does it mean we'll maintain 2 copies of each readme, one in GitHub and one in Bitbucket?

@bianchi2 Once this PR is merged, I'll shorten the README in docker repo, and reference this one instead.

yzha645:
> > > @yzha645 does it mean we'll maintain 2 copies of each readme, one in GitHub and one in Bitbucket?
> > 
> > 
> > @bianchi2 Once this PR is merged, I'll shorten the README in docker repo, and reference this one instead.
> 
> So, any PR to the container image repo will require a PR to this repository as well (e.g. adding a new env var)?

yes. 

The biggest driven forces behind this decision is 1. easier and better consumption for readers, + 2. README will keep growing in the future, trimming it will not be a long-term solution. 
[Postgres](https://hub.docker.com/_/postgres) is doing something similar, with the only difference being: as a Docker Official Image, their README lives in [https://github.com/docker-library/docs](https://github.com/docker-library/docs/tree/master/postgres/README.md) instead. 

Is it a big concern for you?

nanux (approved):

PR #785, opened 2024-04-02
BAM-25729: Document VERBOSE_LOGS everywhere as all images use shared-components

eduardoalvarenga:
As every container uses `shared-components` we can spread the good news about `VERBOSE_LOGS`


yzha645 (approved):

errcode1202:
Hi @eduardoalvarenga just wondering if there is a possibility of sensitive info being leaked into the logs when using `VERBOSE_LOGS`?

errcode1202 (approved):

eduardoalvarenga:
> Hi @eduardoalvarenga just wondering if there is a possibility of sensitive info being leaked into the logs when using `VERBOSE_LOGS`? If so maybe we need to add a warning with this property to state this fact.

Only `unset_secure_vars()` could have a minor possibility of a leak if customised by the customer. The added  debug logs will only tell when the function starts and nothing else.

PR #804, opened 2024-05-12
Optionally install OpenSearch helm chart with Bitbucket

bianchi2:
This PR adds an optional deployment of [OpenSearch cluster](https://github.com/opensearch-project/helm-charts/tree/main/charts/opensearch) together with the Bitbucket Helm chart:

* disabled by default
* if enabled starts a 1 node cluster, ssl disabled
* if enabled, Bitbucket will be automatically configured to use this cluster for search (setting PLUGIN_SEARCH* env vars)
* by default admin users is created, password is auto-generated
* it is possible to provision custom internal users file to create dedicated users and set admin password
* ssl is disabled by default, docs explain how to provide custom certificates and optionally add ca to Java trust store in Bitbucket containers

## Checklist
- [x] I have added unit tests


docs/docs/examples/bitbucket/.pages
```diff
@@ -1,5 +1,6 @@
 nav:
-    - Elasticsearch : BITBUCKET_ELASTICSEARCH.md
+    - OpenSearch    : BITBUCKET_OPENSEARCH.md
+    - Elasticsearch (Deprecated) : BITBUCKET_ELASTICSEARCH.md
```
nanux, line 3: We should add warning panel to the `BITBUCKET_ELASTICSEARCH.md` page and link the OpenSearch page.

src/main/charts/bitbucket/Chart.lock
```diff
@@ -2,5 +2,8 @@ dependencies:
 - name: common
   repository: https://atlassian.github.io/data-center-helm-charts
   version: 1.2.6
-digest: sha256:d5d71e640fd11f06ce6a7f380063611619b8ccfc678ed90093c096f3a63dd5db
-generated: "2024-01-30T06:19:50.968673+11:00"
+- name: opensearch
+  repository: https://opensearch-project.github.io/helm-charts
+  version: 2.19.0
```
nanux, line 7: Will we have CI build to update this version as well? Similar to our products?

src/main/charts/bitbucket/values.yaml
```diff
@@ -1553,3 +1553,59 @@ openshift:
   # configuration files as ConfigMaps.
   #
   runWithRestrictedSCC: false
+
+# OpenSearch config
```
nanux, line 1557: We can refer to our documentation for more details on how to set up OpenSearch properly.

Also, is it necessary to run helm `helm dependency update` after changing `opensearch.enabled` value?

src/main/charts/bitbucket/values.yaml
```diff
@@ -1553,3 +1553,59 @@ openshift:
...
+opensearch:
+
+  # -- Deploy OpenSearch Helm chart and automatically
+  # configure Bitbucket to use it as a search platform
+  #
+  enabled: false
+
+  # -- The base URL of the external OpenSearch instance to be used, for example:
+  # http://opensearch-master.<namespace>.svc.cluster.local:9200
+  # If this is defined and bitbucket.opensearch.enabled is set to false,
```
nanux, line 1568: This is slightly confusing. I can have:
`opensearch.enabled: false`
`opensearch.baseUrl: https://aws.opensearch.url` and it means that opensearch will be enabled on Bitbucket but it won't be installed. Would be more accurate to call it:
`opensearch.install: true` and separate it from `opensearch.enabled`?

src/main/charts/bitbucket/values.yaml
```diff
@@ -1553,3 +1553,59 @@ openshift:
...
+    # When undefined and bitbucket.opensearch.enabled is set to true, a random password
+    # for admin user will be generated and saved to 'opensearch-initial-password' secret.
+    secretName: # opensearch-internal-users
+
+    # -- The key in the Kubernetes Secret that contains the OpenSearch username.
+    #
+    usernameSecretKey: username
+
+    # -- The key in the Kubernetes Secret that contains the OpenSearch password.
+    # The password must be hashed
```
nanux, line 1591: Hashed? Base64?

src/main/charts/bitbucket/values.yaml
```diff
@@ -1553,3 +1553,59 @@ openshift:
...
+    #
+    passwordSecretKey: password
+
+  singleNode: true
+  resources:
+    requests:
+      cpu: 1
+      memory: 1Gi
+  persistence:
+    size: 10Gi
```
nanux, line 1601: It feels that these variables are applicable only when we are installing the opensearch. I think we need to clearly distinguish between both cases - when customer is just enabling opensearch that is external to our installation (whether on the cluster or somewhere in managed infra) and the situation when we are installing the helm charts.

src/main/charts/bitbucket/values.yaml
```diff
@@ -1553,3 +1553,59 @@ openshift:
...
   runWithRestrictedSCC: false
+
+# OpenSearch config
+#
+opensearch:
+
+  # -- Deploy OpenSearch Helm chart and automatically
+  # configure Bitbucket to use it as a search platform
+  #
+  enabled: false
```
nanux, line 1564: How does upgrade work? When I upgrade bitbucket helm charts and something has changed on the opensearch definition as well, what will happen?

src/test/java/test/BitbucketOpenSearchTest.java
```diff
@@ -0,0 +1,96 @@
...
+    private Helm helm;
+
+    @BeforeEach
+    void initHelm(TestInfo testInfo) {
+        helm = new Helm(testInfo);
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = Product.class, names = {"bitbucket"}, mode = EnumSource.Mode.INCLUDE)
+    void opensearch_statefulset_exists(Product product) throws Exception {
```
nanux, line 31: We should also have tests for `enabled: false` && `baseUrl: value` which reflects the situation when customers are bringing their own opensearch.

src/main/charts/bitbucket/templates/_helpers.tpl
```diff
@@ -524,6 +524,57 @@ volumeClaimTemplates:
 {{ end }}
 {{ end }}
 
+{{- define "bitbucket.openSearchEnvVars" -}}
+{{- if and .Values.bitbucket.clustering.enabled (or .Values.opensearch.enabled .Values.opensearch.baseUrl) }}
+- name: SEARCH_ENABLED
```
nanux, line 529: > Set 'false' to prevent OpenSearch (previously Elasticsearch) from starting in the container. This should be used if OpenSearch is running remotely, e.g. for if Bitbucket is running in a Data Center cluster. You may also use ELASTICSEARCH_ENABLED to set this property, however this is deprecated in favor of SEARCH_ENABLED

This confused me slightly :)

nanux: It should be probably false also if number of replicas is > 1, no?

src/main/charts/bitbucket/templates/_helpers.tpl
```diff
@@ -524,6 +524,57 @@ volumeClaimTemplates:
...
+  value: "false"
+- name: PLUGIN_SEARCH_CONFIG_BASEURL
+  value: {{ if .Values.opensearch.baseUrl }}{{ .Values.opensearch.baseUrl }}{{ else if .Values.opensearch.enabled }}http://opensearch-cluster-master:9200{{ else }}{{ .Values.opensearch.baseUrl }}{{ end }}
```
nanux, line 532: I think it would make more sense if `.Values.opensearch.enabled` would be `.Values.opensearch.install.enabled` or something expressing the fact that we have installed it ourselves.

docs/docs/examples/bitbucket/BITBUCKET_OPENSEARCH.md
```diff
@@ -0,0 +1,239 @@
...
+kubectl exec -ti bitbucket-0 -n atlassian -- /bin/bash -c  "chmod +x /opt/atlassian/bitbucket/opensearch/plugins/opensearch-security/tools/hash.sh && /opt/atlassian/bitbucket/opensearch/plugins/opensearch-security/tools/hash.sh -p mySecureBitbucketPassword123"
+
+$2y$12$x910YF09tcfhNs009vOzWOMy9fswhpsuV5/AiiPwbY4rp5BXKv2tv
+```
+
+### Create internal_users.yml file
+
+This is the minimal `internal_users.yml` file with an admin and a dedicated bitbucket user. See [internal_users.yml](https://opensearch.org/docs/latest/security/configuration/yaml/#internal_usersyml){.external} to get more details about users, roles and role mappings in OpenSearch.
+
+```yaml
```
nanux, line 49: It doesn't really say how this file should be mounted and used.

nanux: From the following steps I assume I can just create the file in the directory I am running the `helm` commands from but I think we should be more explicit.

docs/docs/examples/bitbucket/BITBUCKET_OPENSEARCH.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+## Enable SSL with Custom Certificates 
+
+By default, OpenSearch starts with the SSL http plugin disabled, meaning the cluster is accessible via HTTP at http://opensearch-cluster-master:9200. The OpenSearch service is not exposed through a LoadBalancer or Ingress unless the default configurations are explicitly overridden. Bitbucket communicates with the OpenSearch cluster using the service name within the internal Kubernetes network. This setup uses the Kubernetes DNS to resolve the service name to the appropriate cluster IP address.
+
+To enable SSL in OpenSearch and start the service on a secure port, you need to:
+* enable ssl http in Helm values
+* create Kubernetes secrets with ca, certificate and key, and pass them to OpenSearch
+* add ca.crt to Java trust store in Bitbucket containers if the custom certificate is not signed by a public authority
```
nanux, line 121:
```suggestion
* Enable SSL HTTPS in Helm values.
* Create Kubernetes secrets with `ca`, `certificate` and `key`, and pass them to OpenSearch.
* Add `ca.crt` to Java trust store in Bitbucket containers if the custom certificate is not signed by a public authority.
```

docs/docs/examples/bitbucket/BITBUCKET_OPENSEARCH.md
```diff
@@ -0,0 +1,239 @@
...
+
+Below is an example of how to generate a CA certificate, a server certificate, and a corresponding private key for securing communications with OpenSearch:
+
+```shell
+openssl req -new -newkey rsa:2048 -days 365 -nodes -x509 -subj "/C=US/ST=CA/L=LA/O=MyCompany Name/CN=opensearch-cluster-master" -keyout ca.key -out ca.crt
+openssl req -new -newkey rsa:2048 -nodes -keyout opensearch-cluster-master.key -subj "/C=US/ST=CA/L=LA/O=MyCompany Name/CN=opensearch-cluster-master" -out opensearch-cluster-master.csr
+openssl x509 -req -days 365 -in opensearch-cluster-master.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out opensearch-cluster-master.crt
```
nanux, line 128: Should we mention that the certificate will expire in a year?

docs/docs/examples/bitbucket/BITBUCKET_OPENSEARCH.md
```diff
@@ -0,0 +1,239 @@
...
+```shell
+kubectl create secret generic opensearch-ssl-ca -n atlassian --from-file=root-ca.pem=ca.crt
+kubectl create secret generic opensearch-ssl-key -n atlassian --from-file=esnode-key.pem=opensearch-cluster-master.key
+kubectl create secret generic opensearch-ssl-cert -n atlassian --from-file=esnode.pem=opensearch-cluster-master.crt
+```
+
+Enable ssl http plugin, override the default ca, certificate and key, as well as provide additional volume mounts:
+
+```yaml
+opensearch:
```
nanux, line 144: I think it is good to clarify that these changes need to be made to the values of the bitbucket helm chart.

src/main/charts/bitbucket/values.yaml
```diff
@@ -1553,3 +1553,59 @@ openshift:
   # configuration files as ConfigMaps.
   #
   runWithRestrictedSCC: false
+
+# OpenSearch config
+#
+opensearch:
```
nanux, line 1559: This might be far fetched but should we even call it `opensearch`? Should we call it `search` instead and the rest can be implementation detail?

nanux (approved):

PR #812, opened 2024-06-01
CLIP-1878: Update KinD scripts and config to be able to run DC functional tests

bianchi2:
There's a handful of smallish changes to existing KinD scripts and configuration files to make them more flexible. Changes include:

* bump KinD and K8S version, make it possible to pass overrides or use default versions
* deploy an insecure container registry and map its NodePort on host port 5001 (not used at the moment but we may have scenarios where we need to build a local image and push it to the container registry)
* update Postgres deployment function, use the latest Helm chart version, optionally init database if the path to init file is passed
* make deploy script runnable on local Mac, use tmp locations for config files (to run sed against), avoid cd to directories, modify sed command when on Mac
* introduce MICS_OVERRIDES env var in which you can pass whatever `--set`  you need
* group `--set` overrides to avoid duplication of `--set`
* optionally create hostPath PV and PVC (nfs server can't run on arm64 machines, so instead of nfs PV a hostPath PV is used)
* create backdoor NodePort services for the deployed DC app (with sts replica name as label selector and predefined NodePort which is mapped to a host port - 81 and 82)
* add additional env vars and sys props to common config to make deployed products ready for functional testing (disable welcome screen, enable insecure login etc)
* add comments to common config and deploy script to explain why we need extra values and what functions are doing

### DIFF
--- a/src/test/config/kind/hostpath-pv.yaml
+++ b/src/test/config/kind/hostpath-pv.yaml
@@ -27,6 +27,4 @@ spec:
   resources:
     requests:
       storage: 1Gi
-  selector:
-    matchLabels:
-      app: hostpath-shared-home
+


src/test/scripts/kind/configure_kind.sh
```diff
@@ -7,31 +7,43 @@ echo "[INFO]: current-context:" $(kubectl config current-context)
...
-helm install nfs-volume-provisioner nfs-subdir-external-provisioner/nfs-subdir-external-provisioner \
-     -f src/test/config/kind/nfs-values.yaml \
-     --set nfs.server=${nfs_server_ip} \
-     --set nfs.path=/srv/nfs \
-     -n atlassian \
-     --timeout=360s \
-     --wait
+# this is for local runs, because existing nfs server images does not run on arm64 platforms
+# instead, we create a hostPath RWX volume and override the default common settings
+if [ -z "${HOSTPATH_PV}" ]; then
```
yzha645, line 25: Where is this `HOSTPATH_PV` set? Do we set it when we want to run the test in our local arm64 machine?

src/test/config/kind/hostpath-pv.yaml
```diff
@@ -0,0 +1,32 @@
...
+  namespace: atlassian
+spec:
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 1Gi
+  selector:
+    matchLabels:
+      app: hostpath-shared-home
```
yzha645, line 32: Where is this label being set?

yzha645 (approved):

Issue #821, opened 2024-06-19
[Suggestion] - Defining tomcatconfig.proxyName should override ingress.host


Ansible-man:
### Suggestion

We are experiencing issues with both jira and confluence related to reverse proxies. We have our ingress.host set to confluence.doman.com and a proxy that forwards requests to confluence.domain.com from confluence-prod.domain.com which is also the base URL of confluence. This causes errors when logging into confluence because the proxyurl is set to the value of ingress.host despite us setting tomcatconfig.proxyName per https://confluence.atlassian.com/jirakb/base-url-warnings-on-manage-apps-page-in-jira-1381400679.html. We have also tried overriding the ENV vars with no luck. This has been the case with helm serverxml enabled and disabled. 

### Product

Jira, Confluence

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Ansible-man:
We have the following values set
confluence:
  tomcatConfig:
    generateByHelm: true
    proxyName: "confluence-prod.domain.com"

We are observing that the config map generated by helm still sets the proxy name to the value of ingress.host

Ansible-man:
We have tried that method as well will no luck. The ENV vars seem to get overwritten

inka:
@bianchi2 I tried your example and it does not work for me with helm chart version 1.20.1.

I try to deploy Confluence datacenter with dedicated Synchrony URL and fail because both are using 
"ingress.host" to define their env vars.

So a mechanism to override ATL_PROXY_NAME for confluence tomcat would help me out.

My relevant values snippet: 
```
    confluence:
      additionalEnvironmentVariables:
        # this override does not work because env var is defined already before in
        # https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/confluence/templates/statefulset.yaml#L200
        - name: ATL_PROXY_NAME
          value: ${var.fqdn}
      additionalJvmArgs:
        - '-synchrony.proxy.enabled=false'
        - '-Dsynchrony.service.url=https://synchrony-${var.fqdn}/synchrony'
    ingress:
      # this is used by synchrony in SYNCHRONY_SERVICE_URL env var and can not be overwritten
      # https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/confluence/templates/statefulset-synchrony.yaml#L100
      host: synchrony-${var.fqdn}
      https: true
```

inka:
If I run above config I get this in my container spec for confluence-0: 

```
[...]

PR #822, opened 2024-06-20
Make it possible to override proxy settings

bianchi2:
Addresses https://github.com/atlassian/data-center-helm-charts/issues/821
Fallback to ingress settings if no config is provided for proxies.

### DIFF
--- a/src/main/charts/bamboo-agent/Chart.yaml
+++ b/src/main/charts/bamboo-agent/Chart.yaml
@@ -3,7 +3,7 @@ name: bamboo-agent
 description: A chart for installing Bamboo Data Center remote agents on Kubernetes
 type: application
 version: '1.20.0'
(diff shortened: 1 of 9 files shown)


nanux (approved):

nanux:
Do we need to update the tests as well?

src/main/charts/bamboo/templates/configmap-server-config.yaml
```diff
@@ -34,8 +34,8 @@ data:
                    acceptCount="{{ .Values.bamboo.tomcatConfig.acceptCount | default "10" }}"
                    secure="{{ default (ternary "true" "false" .Values.ingress.https) .Values.bamboo.tomcatConfig.secure }}"
```
yzha645, line 35: Shall these be vice versa as well?

PR #832, opened 2024-07-15
Fix default for acceptCount

bianchi2:
It should be 100 not 10.


yzha645:
Shall we update it in corresponding Docker repo as well?

yzha645:
And there are incorrect defaults in `configmap-server-config.yaml`s as well.

yzha645 (approved):

PR #841, opened 2024-07-27
fix access logs mount naming

w1ndblow:
## Logs mounts naming

problem happen with config

```yaml
localHome:
    persistentVolumeClaim:
      create: true
```
Into the pod mounts other volume 
```bash
$ df
/dev/sdg   <> <> <>   1% /opt/atlassian/jira/logs ```


w1ndblow:
@bianchi2 when pvc is creating it binds to first local-home mount (/opt/atlassian/jira/logs) and logs get space whereas local-home directory /opt/atlassian/jira/ is not

PR #842, opened 2024-07-29
Fix mount naming

w1ndblow:


w1ndblow:
@bianchi2 why it impossible? there are `jira.additionalVolumeMounts` variable for that. How you think may be delete this volume? For now pvc binds by log volume, its not right

w1ndblow:
so i create issue https://github.com/atlassian/data-center-helm-charts/issues/844

PR #846, opened 2024-07-31
Add helm value for custom jmx jar location when using other jmx init container images

corang:


corang:
Do you want me to try to fix the output files for testing? @bianchi2

corang:
@bianchi2 If I can I would suggest adding `helm dep build` to the test you sent me, had to go into each chart and do it manually otherwise the tests fail

Michael-Kruggel:
Hey @bianchi2, I was looking to use this feature, but it looks like the subchart hasn't been released to reflect this change in the consuming charts. Is there an eta on this being released?

PR #852, opened 2024-08-08
Allow referencing multiple secrets in additionalCertificates

bianchi2:
There are cases when several self-signed certificates scattered across multiple K8S secrets need to be added to Java truststore. While there's a way to declare a custom `additionalInitContainer`, `additionalVolumes` and `additionalVolumeMounts`, this isn't trivial it requires knowledge of the DC Helm charts design.

This PR introduces an additional `secretList` value that makes it possible to reference multiple secrets with their respective keys in `additionalCertificates` stanza:

* existing secretName takes precedence if defined (**existing behavior is unchanged i.e. it's a non-breaking change**)
* `secretList` is a list of secret names and their respective keys
* each key will be used in subPath which makes it possible to mount certs from multiple secrets
* the resulting mounted file has `<secret-name>-<key>` format which makes it possible to have identical keys in different secrets

## Checklist
- [x] I have added unit tests


src/main/charts/bamboo/templates/statefulset.yaml
```diff
@@ -74,15 +74,26 @@ spec:
...
             - name: keystore
               mountPath: /var/ssl
+          {{- if.Values.bamboo.additionalCertificates.secretName }}
             - name: certs
               mountPath: /tmp/crt
+          {{- else }}
+          {{- range .Values.bamboo.additionalCertificates.secretList }}
+            {{- $secretName := .name }}
+            {{- range .keys }}
+            - name: {{ $secretName }}
```
yzha645, line 91: Correct me if I'm wrong, if there are 2 certs in this secret, it'll ended up having 2 volumeMounts having `- name: {{ $secretName }}` ?

src/main/charts/bamboo/values.yaml
```diff
@@ -971,7 +971,21 @@ bamboo:
...
   #
   additionalCertificates:
+    # -- Name of the Kubernetes secret with certificates in its data. All secret keys in the secret data
+    # will be treated as certificates to be added to Java truststore. If defined, this takes precedence over secretList.
+    #
     secretName:
+    # -- A list of secrets with their respective keys holding certificates to be added to the Java truststore.
+    # It is mandatory to specify which keys from secret data need to be mounted as files to the init container.
+    #
+    secretList:
```
yzha645, line 981:
```suggestion
    secretList: []
```
Probably need to add `[]` if commented out the value, otherwise the list will be treated as a String.

src/main/charts/bamboo/templates/statefulset.yaml
yzha645: which will not be favoured by K8s? If I remember correctly, each volumeMount name must be unique within a container.

yzha645: Ohh, that's the volume name, and will have different sub path. Ignore the above comment, sorry.

yzha645 (approved):

PR #870, opened 2024-08-25
Fix crowd fluentd config

bianchi2:
* local-home isn't mounted to fluentd sidecar
* extraVolumes was sitting under elasticsearch stanza and wasn't actually used in fluentd container
* replace default fluentd config to tail logs rather than to start as a server (yet, it's still recommended to write own fluentd config that fits the particular use case and infra)

## Checklist
- [x] I have added unit tests


nanux:
How was it working before?

nanux (approved):

yzha645 (approved):

Issue #897, opened 2024-10-21
[Suggestion] - Implement workaround for gracefull shutdown when JMX is enabled


farthinder:
### Suggestion

When exposeJmxMetrics is set to true, JIRA will always terminate abruptly on shutdown potentially corrupting DB and creating core dumps. 

In the comments of this [issue](https://jira.atlassian.com/browse/SCALE-141 ) there is a suggested workaround that could either be implemented in shutdown.command or by updating shutdown-wait.sh

I´ve only tested this for JIRA but my guess is this is true for the other products as well. 

### Product

Jira

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #899, opened 2024-10-25
Fix shutdown with jmx javaagent enabled

bianchi2:
Addresses https://jira.atlassian.com/browse/SCALE-141 and https://github.com/atlassian/data-center-helm-charts/issues/897

This PR adds jmx exporter JVM flags to `CATALINA_OPTS` instead of `JVM_SUPPORT_RECOMMENDED_ARGS` (all products except Bitbucket)

When jmx is enabled in Helm chart values, a javaagent is added to `JVM_SUPPORT_RECOMMENDED_ARGS` that is included `JAVA_OPTS`.

From [catalina.sh](https://github.com/apache/tomcat/blob/main/bin/catalina.sh#L65):
```
   #   JAVA_OPTS       (Optional) Java runtime options used when any command
   #                   is executed.
   #                   Include here and not in CATALINA_OPTS all options, that
   #                   should be used by Tomcat and also by the stop process,
   #                   the version command etc.
   #                   Most options should go into CATALINA_OPTS.
```
As a result, when Tomcat is stopped, javaagent starts and since it already listens on jmx exporter port, it fails. This results in non-graceful shutdown.

From [catalina.sh](https://github.com/apache/tomcat/blob/main/bin/catalina.sh#L47):

```
#   CATALINA_OPTS   (Optional) Java runtime options used when the "start",
#                   "run" or "debug" command is executed.
#                   Include here and not in JAVA_OPTS all options, that should
#                   only be used by Tomcat itself, not by the stop process,
#                   the version command etc.
```

## Checklist
- [x] I have added unit tests

### DIFF
--- a/src/test/resources/expected_helm_output/bitbucket/output.yaml
+++ b/src/test/resources/expected_helm_output/bitbucket/output.yaml
@@ -25,6 +25,7 @@ metadata:
 data:
   additional_jvm_args: >-
     -XX:ActiveProcessorCount=1
+    -javaagent:/var/atlassian/application-data/mesh/jmx_prometheus_javaagent.jar=9999:/opt/atlassian/jmx/jmx-config.yaml
   max_heap: 1g
   min_heap: 512m
 ---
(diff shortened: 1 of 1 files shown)


yzha645 (approved):

nanux (approved):

farthinder:
Thanks for this!

PR #901, opened 2024-10-31
Updated pod labels for all products

jimmybchopps:
## Pull request description

Updated labels on all pods under stateful sets to include the same labels on their associated pod workloads.  Previously, the pods only had the selector labels.  With this update, all pods will have the same labels as their parent resource to improve traceability.  Validated the common labels include the selector labels as well so this will only result in a few more labels being added that were previously missing from the associated pod resources.

## Checklist
- [ ] I have added unit tests
- [X ] I have applied the change to all applicable products

### DIFF
--- a/src/main/charts/bamboo/templates/statefulset.yaml
+++ b/src/main/charts/bamboo/templates/statefulset.yaml
@@ -36,7 +36,7 @@ spec:
         {{- end }}
         {{- include "bamboo.podAnnotations" . | nindent 8 }}
       labels:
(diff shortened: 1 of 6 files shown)


jimmybchopps:
@bianchi2 Anything else I need to do for this PR?  I noticed it had a checkbox for adding unit tests, but it doesn't seem like that applies for the changes made in this specific instance.

nanux:
I believe the expected output tests will need to be updated to reflect the fact the labels are now included.

PR #907, opened 2024-11-10
Add s3 backups to Helm values

bianchi2:
Now that Jira filestore-config.xml Jinja template supports S3 config for backups, let's add it to existing Helm values stanza.

Also, improving unit tests a little bit to loop through all 3: avatars, attachments and backups.

## Checklist
- [x] I have added unit tests


nanux (approved):
LGTM.
Thanks @bianchi2 🚀

PR #931, opened 2024-12-16
Make it possible to set nodePort in services

bianchi2:
Addresses https://jira.atlassian.com/browse/SCALE-144

Sometimes, it may be necessary to set an expected nodePort for the service.

## Checklist
- [x] I have added unit tests


src/main/charts/bamboo/values.yaml
```diff
@@ -568,6 +568,10 @@ bamboo:
     #
     type: ClusterIP
 
+    # -- NodePort for Bamboo service
```
nanux, line 571: Do we need to specify that this is applied only when the service is a type of `NodePort` or is this common pattern?

nanux:
We can also document it in https://atlassian.github.io/data-center-helm-charts/userguide/CONFIGURATION/#nodeport-service-type

nanux (approved):

PR #944, opened 2025-01-30
Round up fractional cpu values for ActiveProcessorsCount

bianchi2:
Kubernetes supports [fractional units in cpu requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu). When we calculate ActiveProcessorsCount, and cpu request is, say, 3.1 (which is equivalent to 3200m) we don't round it up, as a result, products fail to start because ActiveProcessorsCount expects int not float.

Changes in this PR round up fractional cpu requests:
* 0.1 becomes 1 (if it's <1 it'll be 1)
* 2.1 becomes 2
* 3.8 becomes 3

I decided not to round up to the closest non fractional value, so if it's 3.8 it's still not 4 cores, that's why it's safer to set 3.

Also, updating upload-action version because [v3 is no longer supported](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). [Test run](https://github.com/atlassian/data-center-helm-charts/actions/runs/13062188755) with new artifacts version.

## Checklist
- [x] I have added unit tests

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -99,5 +99,5 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: kind-artifacts
+          name: kind-artifacts-${{inputs.dc_app}}
           path: logs/


nanux (approved):
Does make sense. LGTM

PR #974, opened 2025-03-25
Update documentation related to Bitbucket shared home requirements

nanux:
## Pull request description

Historically we required a dedicated NFS server to store shared home as it contained the `git` data. With the introduction of BB Mesh, this is no longer the case. The documentation updates reflect relaxed requirements in case the Bitbucket Mesh is deployed.


yzha645:
Can we update the inline comment in Bitbucket [values.yaml](https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bitbucket/values.yaml#L287) file as well?

PR #1032, opened 2025-08-12
BBSDEV-33723 bump PG version to 16 for bitbucket

djansons-atlassian:
## Pull request description

Bumping postgres version to 16 to fix failing bitbucket integration [tests](https://server-syd-bamboo.internal.atlassian.com/browse/BSERV-K8S). Bitbucket 10 does not support PG14.

## Checklist
- [N/A] I have added unit tests
- [ ] I have applied the change to all applicable products - I don't think I should be bumping all of the other product's versions too. 


src/test/config/bitbucket/helm_parameters
```diff
@@ -4,7 +4,7 @@ CHART_TEST_VALUES_BASEDIR=${chart.testValues.basedir}
 PRODUCT_NAME=bitbucket
 DB_NAME=bitbucket
 POSTGRES_CHART_VERSION=11.6.2
-POSTGRES_APP_VERSION=14
+POSTGRES_APP_VERSION=16
```
djansons-atlassian, line 7: I would have thought just this bump would have been sufficient, but my tests were still using PG 14 until I also bumped the version in deploy_app.sh

src/test/config/bitbucket/helm_parameters
```diff
@@ -4,7 +4,7 @@ CHART_TEST_VALUES_BASEDIR=${chart.testValues.basedir}
 PRODUCT_NAME=bitbucket
 DB_NAME=bitbucket
 POSTGRES_CHART_VERSION=11.6.2
```
djansons-atlassian, line 6: Should I also be bumping this value to the latest bugfix? I wasn't sure if I should make it different compared to all of the other products, which use 11.6.2

PR #1037, opened 2025-08-24
fix: Update Bitnami image references due to deprecation notice

kush982:
## Pull request description

_TEMPORARY FIX: This commit provides interim solutions for Bitnami deprecation:

1. JMX Exporter Changes:
- Moved from bitnami/jmx-exporter to bitnamisecure/jmx-exporter:0.18.0
- Updated all applications (Jira, Confluence, Bitbucket, Bamboo, Crowd)
- Modified test output files to match new image references
- This is a temporary solution until direct Maven Central download is implemented

2. Future Work In Progress (branch: feat/bitnami_replacement):
- Implementing direct download of JMX exporter from Maven Central
  (https://repo1.maven.org/maven2/io/prometheus/jmx/jmx_prometheus_javaagent/)
- Replacing Bitnami PostgreSQL with CloudNativePG Helm chart in test cases
- Adding proper checksum verification for downloaded artifacts
- Improving test infrastructure_

- https://github.com/atlassian/data-center-helm-charts/actions/runs/17193113588/job/48771427384

## Checklist
- [ ] I have added unit tests


kush982:
> @kush982 bitnamisecure won't be free. I guess we want to use [bitnamylegacy](https://hub.docker.com/r/bitnamilegacy/jmx-exporter)? Also, perhaps it makes sense to keep jmx exporter version pinned rather than use latest?

`Q: How can I get a complete list of what images and charts are available from Bitnami, both free and commercial?
A: We are working on an enhanced public directory that inventories everything available from Bitnami. In the meantime, you can determine the subset of images that will remain free/community by looking at what is in this registry: [BSI Dockerhub](https://hub.docker.com/u/bitnamisecure/). All remaining Bitnami images/charts, as well as new ones being developed, are available via custom private registries provided as part of a BSI commercial subscription`

The latest tag will be free. However, we can also use the Bitnami legacy version 0.18, as we should be able to migrate this week with our own JMX setup. Additionally, I've added Bitnami Secure, considering security concerns if this gets released.

PR #1098, opened 2026-03-23
DEVPROD-3580: envoy proxy as NodePort for Kind tests

pbhardwaj6:
## Pull request description

Making Gateway Proxy use K8s NodePort, and Openshift Route instead of port forwarding.

## Checklist
- [ ] I have added unit tests

---
Rovo Dev code review: <strong>Rovo Dev couldn't review this pull request</strong>
Upgrade to Rovo Dev Standard to continue using code review.

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -53,10 +53,6 @@ jobs:
         run: |
           src/test/scripts/kind/configure_kind.sh
 
-      - name: Configure test hostname
-        run: |
-          echo "127.0.0.1 dc-app.test" | sudo tee -a /etc/hosts
-
       - name: Deploy postgres database
         run: |
           source src/test/scripts/kind/deploy_app.sh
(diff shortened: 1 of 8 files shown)


bianchi2 (approved):

.github/workflows/kind.yaml
```diff
@@ -53,6 +53,10 @@ jobs:
         run: |
           src/test/scripts/kind/configure_kind.sh
 
+      - name: Configure test hostname
+        run: |
+          echo "127.0.0.1 dc-app.test" | sudo tee -a /etc/hosts
```
amierzwicki: @pbhardwaj6 did you consider moving this to `configure_kind.sh`?

amierzwicki (approved):

.github/workflows/kind.yaml
pbhardwaj6: Yes, we can do that as well, also it would make sense to keep it with kind configuration and not in workflow pipeline.
Will do this change

PR #1116, opened 2026-04-24
feat(confluence): make synchrony readinessProbe timeoutSeconds config…

SuperWonjune:
## Pull request description
The synchrony StatefulSet's readinessProbe has no way to configure `timeoutSeconds`, so it falls back to Kubernetes' default of 1s. In our production environment the probe occasionally fails under load, and we need the ability to tune this value per-cluster.
This adds `synchrony.readinessProbe.timeoutSeconds` (default `1`, no behavior change) and wires it into the template — matching the pattern already used for `confluence.readinessProbe` / `livenessProbe`.

## Checklist
- [x] I have added unit tests


SuperWonjune:
I'll change my email for CLA and make a request again.

PR #1148, opened 2026-09-16
fix(jira,confluence): apply common labels to local-home and additional PVC templates

### LINKED ISSUE #1139: [Suggestion] - Add common labels to PersistentVolumeClaim resources
### Suggestion

The common label feature is important for proper resource management. Especially for backups/restore using Velero or OADP on OCP Kubernetes clusters. For a proper backup configuration not only PODs but also PVCs must be labeled in order to be backed up correctly. Unfortunately the common labels definable in the Atlassian Helm charts is only labeling the STS/PODs but not the PVCs. Hence the request to add the `common.labels.commonLabels` to PVCs as well. That includes the `shared-home`, `local-home` and `additionalVolumeClaimTemplates`.

```yaml
{{- if .Values.volumes.localHome.persistentVolumeClaim.create }}
- apiVersion: v1
  kind: PersistentVolumeClaim
  metadata:
    name: local-home
    labels:
      {{- include "common.labels.commonLabels" . | nindent 6 }}
  spec:
    accessModes: [ "ReadWriteOnce" ]
    {{- if .Values.volumes.localHome.persistentVolumeClaim.storageClassName }}
    storageClassName: {{ .Values.volumes.localHome.persistentVolumeClaim.storageClassName | quote }}
    {{- end }}
    {{- with .Values.volumes.localHome.persistentVolumeClaim.resources }}
    resources:
      {{- toYaml . | nindent 6 }}
    {{- end }}
{{- end }}
```

Please let me know your thoughts on this. I am willing to provide the PR.

### Product

Jira, Confluence

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

sabaighani:
`local-home` PersistentVolumeClaims and any entries under `additionalVolumeClaimTemplates` were missing the `commonLabels` block that every other resource (Pod, StatefulSet, and the `shared-home` PVC) already gets. This breaks backup tooling like Velero/OADP that selects resources by label — the PVC's actual data silently gets skipped from backups even though the Pod is labeled correctly.

Applied the same `labels:` + `include "common.labels.commonLabels"` pattern already used in `shared-home-pvc.yaml`, to both the Jira and Confluence charts.

**Verification:**
- Rendered both charts with `helm template` across several value combinations (default, `additionalVolumeClaimTemplates`, `additionalLabels`, `persistentVolumeClaimRetentionPolicy`) and validated the output with a YAML parser to confirm every PVC gets the correct labels.
- Ran the existing `VolumesTest` suite (`mvn test -Dtest=VolumesTest`) and confirmed via `git stash` A/B comparison that this change produces the exact same test results as `main` (67 run, 65 pre-existing failures unrelated to this change). No new test was added because the existing suite currently can't reach the label assertions for jira/confluence due to that unrelated, pre-existing failure (reported separately).

Closes #1139

## Checklist
- [ ] I have added unit tests


decompil3d:
@atlassian-cla-bot check

